PR #48, opened 2020-12-02
Feature: Add Sentry error tracking

hoh:
Sentry.io provides an easy way to track errors in processes without having to look at the serial output.
Errors can then be analyzed via a web interface and operators notified by email or other channels.

This adds the option to enable sending errors to Sentry, as well as a CLI argument to disable it.


moshemalawach (approved):

moshemalawach:
Seems ok to me.
We might want to make the requirement optionnal, and only import it if it's active?

PR #91, opened 2021-02-05
Hoh fix web3

hoh:


moshemalawach:
If we have the fix for the deprected code, the == is still needed?
Or is there something else?

hoh:
There is something else: the new version fails on a line that is supposed to work. This is why I also block the version number, 
```
File "/opt/pyaleph/src/aleph/web/controllers/metrics.py", line 125, in fetch_eth_height
  return await loop.run_in_executor(None, w3.eth.block_number)
File "/usr/lib/python3.8/concurrent/futures/thread.py", line 57, in run
  result = self.fn(*self.args, **self.kwargs)
TypeError: 'int' object is not callable
```

hoh:
I updated this branch, removed the freezing on fixing the version of the dependency.

hoh:
I raised the question on the web3 library:
https://github.com/ethereum/web3.py/issues/1875

> @hoh thank you for the report! It's historically been a property, and I think it accidentally became a public method, but I don't have any problem moving it back to being a public method as well +1

moshemalawach (approved):

PR #92, opened 2021-02-05
Refactor how processes are started

hoh:
This branch refactors how processes are started from the main function. All the code is running in different processes, the main one is just watching them.

image
_Output from htop, threads hidden_

It is currently deployed on my test node and does not seem to raise any unusual error, but more testing might be useful to have full confidence in these changes.


hoh:
Obsolete with the refactoring of p2p connections using an external daemon.

PR #94, opened 2021-02-22
updated chain name

antonioandara:


hoh:
Thanks for catching this !

PR #98, opened 2021-03-03
Run Pytest on push in Docker Actions

hoh:
Problem: Pytest did not run in CI
Solution: Run Pytest in CI on for both Docker images

Note that builds to fail since some tests are failing, independently from this PR.


hoh:
Obsolete, see #140

PR #119, opened 2021-10-14
WIP: Storage

hoh:


src/aleph/web/controllers/p2p.py
```diff
@@ -11,20 +16,122 @@
...
+        for message in messages
+    )
+    assert usage >= 0
+    return usage
+
+
+async def get_user_balance(address: str) -> float:
+    # TODO: Support ALEPH on chains other than ETH
+    # TODO: Check if using the freekey api key is acceptable
+    API_url = f"https://api.ethplorer.io/getAddressInfo/{address}?apiKey=freekey"
```
hoh, line 48: This API may be deprecated. We should call an Aleph VM that takes care of fetching the balance across chains.

hoh:
We should take into account the duplication of the same IPFS CID in multiple store messages as one.

odesenfans:
Superseded by #203 .

PR #134, opened 2022-01-24
[P2P] Generate private key using cryptography

odesenfans:
Rewrote the private key generation method to use the cryptography
package instead of libp2p. This is part of an ongoing effort to
remove py-libp2p as a dependency.


odesenfans:
Closing, it's not the best thing to do after all. Still going to remove the dependency to py-libp2p but some files were imported in p2pclient to maintain some compatibility.

PR #136, opened 2022-01-25
[P2P] Use the official P2P daemon in the P2P service

odesenfans:
Updated the codebase to remove the dependency to py-libp2p and
use the official JavaScript P2P daemon (jsp2pd) instead.
This includes the following changes:

* Replace all calls to libp2p by RPC calls to jsp2pd. This is
  achieved using `p2pclient`, aka a client library for the P2P
  daemon.
* Remove all dependencies to libp2pd by equivalent functionalities
  provided by p2pclient.
* Change the private/public key generation system for compatibility
  with jsp2pd.
* Provide a Dockerfile for jsp2pd and use it in the docker-compose
  flow.

Breaking changes:
* `pyaleph --gen-key` was renamed to `--gen-keys`. It now creates
  a directory containing the private key in PEM format, the public
  key in PEM format and the private key in a serialized protobuf
  + DER format required by jsp2pd.
* Similarly, the `--key-path` option was renamed to `--key-dir`.
* pyaleph now looks for the serialized key file instead of
  the PEM file. It now exits with exit code 1 (instead of 0)
  if the file does not exist.
* Providing the private key by specifying it as a string in
  the config file (config.p2p.key) is not possible anymore.

# Todo
- [x] Update the "official" docker-compose file in deployment/docker-compose
- [x] Provide a migration script to generate the serialized key file from a PEM key
- [ ] Push the new Docker images and use them in the docker-compose files

### DIFF
--- a/deployment/docker-build/pyaleph.dockerfile
+++ b/deployment/docker-build/pyaleph.dockerfile
@@ -75,13 +75,16 @@ COPY .git /opt/pyaleph/.git
 
 # Setup directories for `python setup.py develop`
 USER root
-RUN mkdir /opt/pyaleph/src/pyaleph.egg-info
+RUN mkdir -p /opt/pyaleph/src/pyaleph.egg-info
 RUN mkdir /opt/pyaleph/.eggs
 RUN chown -R source:source /opt/pyaleph/src /opt/pyaleph/.eggs /opt/pyaleph/.git
 
 # Install PyAleph source
 USER source
 WORKDIR /opt/pyaleph
+# TODO: replace by a proper install of p2pclient once the changes merged and released on the mainline repo
+RUN pip install -U --use-deprecated=legacy-resolver git+https://github.com/odesenfans/py-libp2p-daemon-bindings.git@c36b0262bf0b7581c0f9662c3f2fb4368e6b3c28
+
 RUN pip install -U --use-deprecated=legacy-resolver git+https://github.com/aleph-im/nuls2-python.git
 RUN pip install -U --use-deprecated=legacy-resolver cosmospy
 RUN pip install -U --use-deprecated=legacy-resolver substrate-interface 'eth-keys==0.3.3'
@@ -99,5 +102,3 @@ CMD ["pyaleph"]
 
 # PyAleph API
 EXPOSE 8000
-# PyAleph p2p network
-EXPOSE 4024 4025
--- a/deployment/docker-compose/docker-compose.yml
+++ b/deployment/docker-compose/docker-compose.yml
@@ -7,30 +7,47 @@ volumes:
 services:
   pyaleph:
     restart: always
-    image: alephim/pyaleph-node:beta
-    command: pyaleph --config /opt/pyaleph/config.yml
+    image: pyaleph:test
+    command: pyaleph --config /opt/pyaleph/config.yml --key-dir /opt/pyaleph/keys -vv
     ports:
-      - 127.0.0.1:8000:8000/tcp
-      - 4024:4024/tcp
-      - 4025:4025/tcp
+      - "127.0.0.1:8000:8000/tcp"
+      - "4024:4024/tcp"
     volumes:
       - ./config.yml:/opt/pyaleph/config.yml
-      - ./node-secret.key:/opt/pyaleph/node-secret.key
+      - ./keys:/opt/pyaleph/keys
     depends_on:
       - mongodb
       - ipfs
+      - p2pd
     networks:
       - pyaleph
     logging:
       options:
         max-size: 50m
 
+  p2pd:
+    restart: always
+    image: alephim/jsp2pd:0.10.2-1.0.0
+    networks:
+      - pyaleph
+    environment:
+      PRIVATE_KEY_FILE: "/etc/jsp2pd/keys/serialized-node-secret.key"
+      LISTEN_MADDR: "/ip4/0.0.0.0/tcp/4020"
+      HOST_MADDRS: "/ip4/0.0.0.0/tcp/4025"
+      PUBSUB: "true"
+      PUBSUB_ROUTER: "floodsub"
+    ports:
+      - "4020:4020"
+      - "4025:4025"
+    volumes:
+      - ./keys:/etc/jsp2pd/keys
+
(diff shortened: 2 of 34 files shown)


odesenfans:
Todo list:
[ ] Update the "official" docker-compose file in deployment/docker-compose
[ ] Provide a migration script to generate the serialized key file from a PEM key

src/aleph/services/p2p/protocol.py
```diff
@@ -29,100 +28,84 @@
...
+            LOGGER.exception("Error while reading data")
+
+        await stream.send_all(json.dumps(result).encode("utf-8"))
+
+    async def make_request(self, request_structure: Dict[str, Any]) -> Optional[Dict[str, Any]]:
+        peers = [peer for peer in self.peers]
+        # Randomize the list of peers to contact to distribute the load evenly
+        random.shuffle(peers)
+
+        for peer in peers:
```
hoh, line 92: This was in a `while True`, why not anymore ?

src/aleph/services/p2p/protocol.py
```diff
@@ -29,100 +28,84 @@
...
+            try:
+                value = json.loads(response.decode("UTF-8"))
+            except json.JSONDecodeError:
+                logging.warning("Could not decode response from %s", peer)
+                continue
+
+            if value.get("content") is not None:
+                return value
+
+        logging.info("Could not retrieve content from any peer")
```
hoh, line 107: Should this be a warning ?

deployment/docker-build/pyaleph.dockerfile
```diff
@@ -75,13 +75,16 @@ COPY .git /opt/pyaleph/.git
...
-RUN mkdir /opt/pyaleph/src/pyaleph.egg-info
+RUN mkdir -p /opt/pyaleph/src/pyaleph.egg-info
 RUN mkdir /opt/pyaleph/.eggs
 RUN chown -R source:source /opt/pyaleph/src /opt/pyaleph/.eggs /opt/pyaleph/.git
 
 # Install PyAleph source
 USER source
 WORKDIR /opt/pyaleph
+# TODO: replace by a proper install of p2pclient once the changes merged and released on the mainline repo
+RUN pip install -U --use-deprecated=legacy-resolver git+https://github.com/odesenfans/py-libp2p-daemon-bindings.git@c36b0262bf0b7581c0f9662c3f2fb4368e6b3c28
```
hoh, line 86: Repository to migrate in `aleph-im`.

odesenfans: There should be a release of the official p2pclient package soon, I am in contact with the maintainer.

deployment/docker-compose/docker-compose.yml
```diff
@@ -7,30 +7,48 @@ volumes:
...
+      - pyaleph
+    environment:
+      PRIVATE_KEY_FILE: "/etc/jsp2pd/keys/serialized-node-secret.key"
+      LISTEN_MADDR: "/ip4/0.0.0.0/tcp/4020"
+      HOST_MADDRS: "/ip4/0.0.0.0/tcp/4022"
+      PUBSUB: "true"
+      PUBSUB_ROUTER: "floodsub"
+    ports:
+      - "4020:4020"
+      - "4022:4022"
```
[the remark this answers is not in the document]
odesenfans: Okay I did not take the discovery process into account, it's true that the initial addresses are all in 4025. At runtime the port doesn't matter much, since the node advertises its own address.

I changed the port to make sure I wasn't messing with other config options, it's easy to reverse.

setup.cfg
```diff
@@ -31,37 +31,37 @@ setup_requires =
...
+      coincurve<16.0.0
       configmanager
-      uvloop
-      web3
-      eth_account
+      configparser
+      cosmospy
+      dataclasses_json>=0.5.2
+      eth_account<0.6.0
+      eth-rlp==0.2.1
```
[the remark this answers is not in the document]
odesenfans: I had to peg this one to a known version, there seemed to be a few issues with the pip dependency resolver otherwise. I'll try to remove it now that the code is cleaned up, otherwise we can certainly remove it safely in #138 .

src/aleph/commands.py
```diff
@@ -143,14 +151,11 @@ def main(args):
...
-            config.p2p.key.value = key_file.read()
-
-    if not config.p2p.key.value:
-        LOGGER.critical("Node key cannot be empty")
-        return
+    # We only check that the serialized key exists.
+    serialized_key_file_path = os.path.join(args.key_dir, "serialized-node-secret.key")
+    if not os.path.isfile(serialized_key_file_path):
+        LOGGER.critical(f"Serialized node key ({serialized_key_file_path}) not found.")
+        sys.exit(1)
```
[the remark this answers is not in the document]
odesenfans: Fair point, but then we need an exception handler at the top level to exit with the correct exit code.

src/aleph/services/p2p/__init__.py
```diff
@@ -1,28 +1,33 @@
...
+    return P2PClient(control_maddr=control_maddr, listen_maddr=listen_maddr)
 
 
-async def init_p2p(config, listen=True, port_id=0) -> List[Coroutine]:
-    pkey = config.p2p.key.value
-    port = config.p2p.port.value + port_id
-    singleton.host, singleton.pubsub, singleton.streamer, tasks = await initialize_host(
-        key=pkey,
+async def init_p2p(config, listen: bool = True, port_id: int = 0) -> List[Coroutine]:
+    singleton.client = init_p2p_client(config)
```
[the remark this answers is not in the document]
odesenfans: No, but that's something I left for another PR.

src/aleph/services/p2p/protocol.py
odesenfans: I don't think so, I don't think we check that the content must exist beforehand?

src/aleph/services/p2p/protocol.py
odesenfans: It was not a real while True, but more of a loop on all the streams opened with all the peers to see if someone has the content we are looking for. I just simplified it.

PR #137, opened 2022-01-25
[Storage] Decommission RocksDB

odesenfans:
[Storage] Decommission RocksDB

Decommissioned the support for RocksDB. The only supported storage
engine is now MongoDB.

Breaking changes:
* The config.storage.engine config value can now only take
  "mongodb" as value.


odesenfans:
Depends on #136.

PR #138, opened 2022-01-26
[Setup] Fixed versions for dependencies

odesenfans:
Specified exact versions for all dependencies of PyAleph.
This avoids differences between installs of the same version
of PyAleph on different nodes and will avoid issues linked
to new releases of some packages that may bring breaking
changes.

Simplified the PyAleph Dockerfile as most of the tricks used
to work around the dependency issues of libp2p are no longer
necessary.

Deleted the unused requirements.txt file at the root of the project.


odesenfans:
Depends on #136 .

odesenfans:
To discuss:
- [ ] Should cosmos be in the base install or remain in [cosmos]?
- [ ] Should nuls2 be in the base install or remain in [nuls2]?

PR #139, opened 2022-01-26
[Setup] Install dependencies early in Dockerfile

odesenfans:
Modified the PyAleph Dockerfile to separate the install of
the dependencies from PyAleph itself. This way, we can install
the requirements once and cache them in a layer of the image,
avoiding to fetch them each time.

A new script, `extract_requirements.py`, is used to read
the install requirements from setup.cfg into a temporary
requirements.txt file.

Removed the install of pytest from the Docker image as it is
not relevant to a Docker image.


odesenfans:
Depends on #138 .

hoh:
Looks like a decent enough approach to the problem.

PR #140, opened 2022-02-01
[CI] New CI/CD pipeline and fixes for unit tests

odesenfans:
Added a new pipeline to run unit tests and build the Docker images.
The Docker part is similar to the previous pipeline.

Fixed the unit tests that were easy to fix, and marked the others
as skipped. One important fix is the resolution of a circular
dependency between the chains module and the network module.

Deleted the "demo" pipeline as it served no clear purpose.


.github/workflows/pyaleph-ci.yml
```diff
@@ -0,0 +1,49 @@
...
+jobs:
+  tests:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        # Fetch the whole history for all tags and branches (required for aleph.__version__)
+        fetch-depth: 0
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v2
```
hoh, line 25: Is this better than using the distribution's packages via `apt` ?

tests/chains/test_register.py
```diff
@@ -58,7 +56,7 @@ async def test_register_outgoing_worker(monkeypatch):
 
 @pytest.mark.asyncio
 async def test_outgoing():
-    assert len(OUTGOING_WORKERS.keys()) == 3  # 3 verifiers are included by default
+    assert len(OUTGOING_WORKERS) == 2  # 2 verifiers are included by default
```
hoh, line 59: Do you know why this went from 3 to 2 ?

hoh (commented):
Looks good. I have 2 minor comments that need to be addressed.

.github/workflows/pyaleph-ci.yml
odesenfans: Yes, because you can specify the exact version of Python.

tests/chains/test_register.py
odesenfans: The support for BNB was dropped, it was one of the 3.

.github/workflows/pyaleph-ci.yml
hoh: And in the future test with multiple versions of Python.

PR #141, opened 2022-02-02
[CI] Add a CI job to run type checking with mypy

odesenfans:
Fixed type checking issues on the whole codebase.

### DIFF
--- a/.github/workflows/pyaleph-ci.yml
+++ b/.github/workflows/pyaleph-ci.yml
@@ -8,7 +8,6 @@ on:
     branches:
       - dev
       - master
-      - od-mypy
   pull_request:
     branches:
       - dev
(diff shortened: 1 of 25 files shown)


.github/workflows/pyaleph-ci.yml
```diff
@@ -8,6 +8,7 @@ on:
     branches:
       - dev
       - master
+      - od-mypy
```
hoh, line 11: Really ?

mypy.ini
```diff
@@ -0,0 +1,65 @@
+# Global options:
```
hoh, line 1: Were all of these necessary for the codebase to pass ?

odesenfans: No, these are a few default options I like. We can tweak it as we go.

.github/workflows/pyaleph-ci.yml
odesenfans: Fixed.

mypy.ini
hoh: OK.

PR #142, opened 2022-02-02
Fix: Monitoring did not show available memory and did not use unit MB

hoh:
The rest are automatic schema changes with a newer version of Grafana


odesenfans (approved):
Looks alright.

PR #154, opened 2022-02-08
[Updater] New migration system for config updates

odesenfans:
Added a new tool to update the config and key files in case
of breaking change. This will help users move to newer versions
more easily. The tool is inspired from alembic and runs custom
Python migration scripts one by one.


deployment/migrations/scripts/0001-update-keys-for-p2pd.py
```diff
@@ -0,0 +1,63 @@
...
+    with open(key_file) as f:
+        private_key_str = f.read()
+
+    private_key = RSAPrivateKey(RSA.import_key(private_key_str))
+    key_pair = KeyPair(private_key=private_key, public_key=private_key.get_public_key())
+    save_keys(key_pair, output_dir)
+
+
+def upgrade(**kwargs):
+    key_dir = kwargs["key_dir"]
```
hoh, line 32: Using `pathlib.Path` can make joining paths easier: https://docs.python.org/3/library/pathlib.html , especially for joining using `/`, checking if a file exists, ...

```suggestion
    from pathlib import Path
    key_dir = Path(kwargs["key_dir"])
```

deployment/migrations/config_updater.py
```diff
@@ -0,0 +1,103 @@
...
+    migration_scripts_dir = os.path.join(os.path.dirname(__file__), "scripts")
+    migration_scripts = sorted(
+        f for f in os.listdir(migration_scripts_dir) if f.endswith(".py")
+    )
+
+    command = args.command
+
+    for migration_script in migration_scripts:
+        migration_script_path = os.path.join(migration_scripts_dir, migration_script)
+        spec = importlib.util.spec_from_file_location(
```
hoh, line 83: This approach seems pretty complex. Can you explain why such a complex framework instead of either importing and running every migration (should be able to self-check and run more than once) or running executable scripts independently of the language ?

deployment/migrations/scripts/0001-update-keys-for-p2pd.py
odesenfans: I prefer the old school way of manipulating paths, but sure if you prefer. Not a big deal.

hoh: I don't always have the reflex neither, but I think it would be useful here.

deployment/migrations/config_updater.py
odesenfans: I didn't invent anything, this is really similar to the way alembic (the SQLAlchemy migration tool) works. I like it so I figured I'd do something similar (upgrade/downgrade functions in Python scripts, automatic import from a scripts directory).

To answer your questions directly:
* I don't see your point regarding self-check and running more than once? This is up to the migration script. For example, the current script checks that the file to create already exists.
* Language independence: that's a fair point, but worst case just spawn a subprocess in the migration script and that's all.

As for the benefits:
* Automatic discovery of the scripts, impossible to forget one somewhere
* Pretty natural to write, just need to implement upgrade and downgrade and you can forget about all the boilerplate like command-line parsing etc.

hoh (approved):

Issue #156, opened 2022-02-09
Support search by key in messages.json API endpoint


moshemalawach:
The messages.json api endpoint should support a "keys" param that searches in "content.key".

The change should be done in src/aleph/web/controllers/messages.py both in KNOWN_QUERY_FIELDS and get_filters

odesenfans:
Is this what you need? https://github.com/aleph-im/pyaleph/pull/157

I pointed it to the dev branch but I can point to `master` if you prefer.

odesenfans:
Resolved by #157.

PR #157, opened 2022-02-09
[API] Search messages by content keys

odesenfans:
Added a new `contentKeys` filter to the messages GET endpoints
to filter messages by content key. Several keys can be specified.

Fixed an issue related to the filtering of tags and hashes
introduced a few commits ago.


odesenfans:
Resolves #156.

hoh (approved):

PR #180, opened 2022-02-20
[DB] Remove partial attribute on retries + time index

odesenfans:
Fixed an issue where this index would not be used by the query
listing pending messages because the query looks for all messages
and not only those who have already been retried. This would result
in a memory error with MongoDB given the number of pending messages.


odesenfans:
Note that running this PR on a node using v0.2.0-rc1 requires to drop the previous index manually, as they have the same name.

hoh (approved):

PR #182, opened 2022-02-21
Feature: Run migrations using docker-compose

hoh:


docs/guides/upgrade.rst
```diff
@@ -41,17 +41,17 @@ Upgrade your node
...
+    docker-compose [-f <docker-compose-file>] down
+
+    docker-compose [-f <docker-compose-file>] \
+            run pyaleph /opt/pyaleph/deployment/migrations/config_updater.py \
             --key-dir /opt/pyaleph/keys \
             --key-file /opt/pyaleph/node-secret.key \
             --config /opt/pyaleph/config.yml \
             upgrade
 
+    docker-compose [-f <docker-compose-file>] down
```
odesenfans, line 53: down -> up -d

docs/guides/upgrade.rst
```diff
@@ -41,17 +41,17 @@ Upgrade your node
...
-    docker run --rm -ti \
-        -v $(pwd)/keys:/opt/pyaleph/keys \
-        -v $(pwd)/node-secret.key:/opt/pyaleph/node-secret.key:ro \
-        -v $(pwd)/config.yml:/opt/pyaleph/config.yml:ro \
-        alephim/pyaleph-node:beta \
-        python3 /opt/pyaleph/migrations/config_updater.py \
+    docker-compose [-f <docker-compose-file>] down
+
+    docker-compose [-f <docker-compose-file>] \
+            run pyaleph /opt/pyaleph/deployment/migrations/config_updater.py \
```
odesenfans, line 47: Incorrect path, the correct path is `/opt/pyaleph/migrations/config_updater.py`.

odesenfans (changes requested):

docs/guides/upgrade.rst
hoh: The file was missing from the container.
I updated the commit to add it.

docs/guides/upgrade.rst
hoh: The docker-compose `run` command will start the other services.
Do do the migration independently from restarting the service, I here suggest to first tear everything down before restarting it.

odesenfans: Okay, if you prefer.

docs/guides/upgrade.rst
odesenfans: It's not missing, it's just not copied in the same place.

deployment/docker-build/pyaleph.dockerfile
```diff
@@ -70,6 +71,7 @@ RUN rm /opt/build/extract_requirements.py /opt/build/requirements.txt
 COPY deployment/migrations /opt/pyaleph/migrations
```
odesenfans, line 71: @hoh see here, this is where the config updater is copied.

PR #183, opened 2022-02-21
Fix: All PermanentPin were only created on new nodes

hoh:
This solves an issue where existing nodes would not create PermanentPin documents for chaindata that has already been processed.

To achieve this, it resets the chaindata height so that all chaindata is processed once more by the node.

### DIFF
--- a/deployment/migrations/scripts/0002-refresh-chain-pins.py
+++ b/deployment/migrations/scripts/0002-refresh-chain-pins.py
@@ -23,13 +23,17 @@ async def async_upgrade(config_file: Optional[str], **kwargs):
 
     init_db(config=config, ensure_indexes=False)
 
-    if (await PermanentPin.count(filter={})) < 5000:
+    # We measure over 5000 permanent pins on new nodes that did process all chaindata.
+    # We therefore use this value to estimate if a node did process all chaindata already or not.
+    expected_permanent_pins = 5000
+
+    if (await PermanentPin.count(filter={})) < expected_permanent_pins:
         logger.info("PermanentPin documents missing, fetching chaindata again")
-        start_height = get_defaults()["ethereum"]["start_height"]
+        start_height = config.ethereum.start_height.value
         await Chain.set_last_height("ETH", start_height)
     else:
         logger.info(
-            "PermanentPin documents already present, not need to re-fetch chaindata"
+            "PermanentPin documents already present, no need to re-fetch chaindata"
         )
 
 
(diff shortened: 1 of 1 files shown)


deployment/migrations/scripts/0002-refresh-chain-pins.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+async def async_upgrade(config_file: Optional[str], **kwargs):
+    config = Config(schema=get_defaults())
+    if config_file is not None:
+        config.yaml.load(config_file)
+
+    init_db(config=config, ensure_indexes=False)
+
+    if (await PermanentPin.count(filter={})) < 5000:
```
odesenfans, line 26: Maybe use a constant here to be more explicit? I think we should document why we use 5000 here.

deployment/migrations/scripts/0002-refresh-chain-pins.py
```diff
@@ -0,0 +1,42 @@
...
+
+logger = logging.getLogger()
+
+
+async def async_upgrade(config_file: Optional[str], **kwargs):
+    config = Config(schema=get_defaults())
+    if config_file is not None:
+        config.yaml.load(config_file)
+
+    init_db(config=config, ensure_indexes=False)
```
odesenfans, line 24: The DB init part should be mutualized for all migration scripts moving forward, and a `db` object passed as argument to the scripts.

deployment/migrations/scripts/0002-refresh-chain-pins.py
```diff
@@ -0,0 +1,42 @@
...
+
+    if (await PermanentPin.count(filter={})) < 5000:
+        logger.info("PermanentPin documents missing, fetching chaindata again")
+        start_height = get_defaults()["ethereum"]["start_height"]
```
odesenfans, line 28: Why `get_defaults()` and not `config`?

deployment/migrations/scripts/0002-refresh-chain-pins.py
```diff
@@ -0,0 +1,42 @@
...
+        await Chain.set_last_height("ETH", start_height)
+    else:
+        logger.info(
+            "PermanentPin documents already present, not need to re-fetch chaindata"
```
odesenfans, line 32: not -> no

odesenfans (commented):
A few remarks.

deployment/migrations/scripts/0002-refresh-chain-pins.py
hoh: Good point, that was before importing the config

deployment/migrations/scripts/0002-refresh-chain-pins.py
hoh: I prefer to follow the [YAGNI](https://en.wikipedia.org/wiki/You_aren't_gonna_need_it) principle. This would make sense if other migration scripts rely on the database, but that is not the case at the moment.

odesenfans: Fair enough.

odesenfans (approved):

PR #184, opened 2022-02-21
Fix: On some servers, ip returned as chinese chars instead if IPv4

hoh:
The bug only happened when using aiohttp and was not reproduced using requests.

We should investigate the issue more since this does not identify the source of the problem and the solution is not elegant.


odesenfans (approved):
Extremely werird but ok.

odesenfans:
Outdated because of #185.

PR #185, opened 2022-02-22
Fix: Protect against wrong IPv4 format from 3rd party service.

hoh:
On some servers, get_IP() returned chinese chars instead if a proper IPv4. This protects against invalid IP formats.

### DIFF
--- a/src/aleph/services/utils.py
+++ b/src/aleph/services/utils.py
@@ -1,18 +1,48 @@
+import logging
+import re
 import socket
 
 import aiohttp
 
+logger = logging.getLogger(__name__)
 
-async def get_IP():
-    ip = None
+IP4_SERVICE_URL = "https://v4.ident.me/"
+IP4_SOCKET_ENDPOINT = "8.8.8.8"
+
+
+def is_valid_ip4(ip: str) -> bool:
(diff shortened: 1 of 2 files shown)


src/aleph/services/utils.py
```diff
@@ -1,18 +1,44 @@
...
+    finally:
         s.close()
 
-    return ip
+
+async def get_IP() -> str:
+    """Get the public IPv4 of this system."""
+    try:
+        return await get_ip4_from_service()
+    except:
```
odesenfans, line 43: Maybe log the error as a warning?

src/aleph/services/utils.py
```diff
@@ -1,18 +1,44 @@
...
 import aiohttp
 
+IP4_SERVICE_URL = "https://v4.ident.me/"
+IP4_SOCKET_ENDPOINT = "8.8.8.8"
 
-async def get_IP():
-    ip = None
+
+def is_valid_ip4(ip: str) -> bool:
+    return bool(re.match(r"\d+\.\d+\.|d+\.\d+", ip))
```
odesenfans, line 11: Typo, | -> \. Maybe add a small test for this function?

odesenfans (changes requested):

odesenfans (approved):

PR #195, opened 2022-02-26
[Storage] Unit tests and type hints

odesenfans:
Added unit tests for the storage component. These tests validate
the logic to retrieve data from MongoDB, the P2P network and/or
IPFS.

Added type hints.

This work precedes a larger refactoring of the storage module
to get rid of status codes entirely.


src/aleph/model/hashes.py
```diff
@@ -31,19 +35,19 @@
...
-async def get_value(key: str):
+async def get_value(key: str) -> Optional[bytes]:
     from aleph.model import fs
     if fs is None:
         raise ValueError("MongoDB fs not initialized")
 
     try:
         gridout = await fs.open_download_stream_by_name(key)
-        return await gridout.read()
+        return await gridout.reaopen_download_stream_by_named()
```
hoh, line 45: I cannot find the definition of `reaopen_download_stream_by_named` in GridOut.
Where is it defined ?

odesenfans: Typo, the "a" should not be there.

PR #196, opened 2022-02-26
[Storage] Rewrite status codes as exceptions

odesenfans:
Simplified the storage code and made it more pythonic by replacing
status codes like None and -1 by exceptions. Two new exceptions
describe if the content the user is looking for is invalid (-1)
or just currently unavailable because of a synchronisation issue
or a technical issue (None).

This simplifies the whole logic and avoids the propagation of weird
codes all over the place.

Added new dataclasses to keep track of the retrieved content,
its hash and the source from which the data was fetched.
The `RawContent` class keeps track of binary content while
the `JsonContent` class represents raw content that was interpreted
as JSON.

### DIFF
--- a/src/aleph/chains/common.py
+++ b/src/aleph/chains/common.py
@@ -214,7 +214,7 @@ async def incoming(
             return IncomingStatus.FAILED_PERMANENTLY
 
         except (ContentCurrentlyUnavailable, Exception) as e:
-            if not isinstance(ContentCurrentlyUnavailable, e):
+            if not isinstance(e, ContentCurrentlyUnavailable):
                 LOGGER.exception("Can't get content of object %r" % hash)
             await mark_message_for_retry(
                 message=message,
--- a/src/aleph/chains/common.py
+++ b/src/aleph/chains/common.py
@@ -18,6 +18,11 @@ from aleph.permissions import check_sender_authorization
 from aleph.storage import get_json, pin_hash, add_json, get_message_content
 from aleph.types import UnknownHashError
 from aleph.web import app
+from aleph.exceptions import (
+    AlephStorageException,
+    InvalidContent,
+    ContentCurrentlyUnavailable,
+)
 
 LOGGER = logging.getLogger("chains.common")
 
@@ -94,15 +99,15 @@ async def mark_message_for_retry(
 
 
 async def incoming(
-        message,
-        chain_name: Optional[str] = None,
-        tx_hash: Optional[str] = None,
-        height: Optional[int] = None,
-        seen_ids: Optional[Dict[Tuple, int]] = None,
-        check_message: bool = False,
-        retrying: bool = False,
-        bulk_operation: bool = False,
-        existing_id: Optional[ObjectId] = None
+    message,
+    chain_name: Optional[str] = None,
+    tx_hash: Optional[str] = None,
+    height: Optional[int] = None,
+    seen_ids: Optional[Dict[Tuple, int]] = None,
+    check_message: bool = False,
+    retrying: bool = False,
+    bulk_operation: bool = False,
+    existing_id: Optional[ObjectId] = None,
 ) -> Union[IncomingStatus, UpdateOne]:
     """New incoming message from underlying chain.
 
@@ -202,13 +207,14 @@ async def incoming(
         #     new_values = {'confirmed': False}  # this should be our default.
 
         try:
-            content, size = await get_message_content(message)
-        except Exception:
-            LOGGER.exception("Can't get content of object %r" % hash)
-            content, size = None, None
+            content = await get_message_content(message)
 
-        if content is None:
-            LOGGER.info("Can't get content of object %r, retrying later." % hash)
+        except InvalidContent:
+            LOGGER.warning("Can't get content of object %r, won't retry." % hash)
+            return IncomingStatus.FAILED_PERMANENTLY
+
+        except (ContentCurrentlyUnavailable, Exception):
+            LOGGER.exception("Can't get content of object %r" % hash)
             await mark_message_for_retry(
                 message=message,
                 chain_name=chain_name,
@@ -220,15 +226,12 @@ async def incoming(
             )
             return IncomingStatus.RETRYING_LATER
 
-        if content == -1:
-            LOGGER.warning("Can't get content of object %r, won't retry." % hash)
-            return IncomingStatus.FAILED_PERMANENTLY
+        json_content = content.value
+        if json_content.get("address", None) is None:
+            json_content["address"] = message["sender"]
 
-        if content.get("address", None) is None:
-            content["address"] = message["sender"]
-
-        if content.get("time", None) is None:
-            content["time"] = message["time"]
+        if json_content.get("time", None) is None:
+            json_content["time"] = message["time"]
 
         # warning: those handlers can modify message and content in place
         # and return a status. None has to be retried, -1 is discarded, True is
@@ -236,19 +239,17 @@ async def incoming(
         # TODO: change this, it's messy.
         try:
             if message["type"] == MessageType.store:
-                handling_result = await handle_new_storage(message, content)
+                handling_result = await handle_new_storage(message, json_content)
             elif message["type"] == MessageType.forget:
                 # Handling it here means that there we ensure that the message
                 # has been forgotten before it is saved on the node.
                 # We may want the opposite instead: ensure that the message has
                 # been saved before it is forgotten.
-                handling_result = await handle_forget_message(message, content)
+                handling_result = await handle_forget_message(message, json_content)
             else:
                 handling_result = True
         except UnknownHashError:
-            LOGGER.warning(
-                f"Invalid IPFS hash for message {hash}, won't retry."
-            )
+            LOGGER.warning(f"Invalid IPFS hash for message {hash}, won't retry.")
             return IncomingStatus.FAILED_PERMANENTLY
         except Exception:
             LOGGER.exception("Error using the message type handler")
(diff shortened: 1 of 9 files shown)


hoh:
Cool !
Je vois pas mal de lignes de changements, n'hésite pas à bien séparer tout ce qui est type hints et cleanup afin de sortir ça du gros de la PR, voire le merger avant.

odesenfans:
Yes, there's #195 already for type hints and unit tests, this one depends on it. The main change has impacts all over the place, but I'll check if I can simplify it.

src/aleph/storage.py
```diff
@@ -22,113 +24,207 @@
 LOGGER = logging.getLogger("STORAGE")
 
 
-async def json_async_loads(s: str):
+class ContentSource(str, Enum):
```
hoh, line 27: Can you add a docstring to explain the use of this class ?

src/aleph/storage.py
```diff
@@ -22,113 +24,207 @@
...
+class RawContent(StoredContent):
+    value: bytes
+
+    def __len__(self):
+        return len(self.value)
+
+
+@dataclass
+class JsonContent(StoredContent):
+    value: Any
```
hoh, line 50: It could be worth defining a type for JSON compatible types instead of using Any everywhere: `Union[Dict, List, str, int, float, bool, None]`

src/aleph/storage.py
```diff
@@ -22,113 +24,207 @@
...
+    raw_content: bytes
```
hoh, line 51: This name makes me think about `class RawContent` just above.
```suggestion
    raw_value: bytes
```

src/aleph/storage.py
```diff
@@ -22,113 +24,207 @@
...
+            raise InvalidContent(error_msg)  # never retry, bogus data
         try:
             item_content = await json_async_loads(message["item_content"])
-        except (json.JSONDecodeError, KeyError):
-            LOGGER.exception("Can't decode JSON")
-            return -1, 0  # never retry, bogus data
-        return item_content, len(message["item_content"])
+        except (json.JSONDecodeError, KeyError) as e:
+            error_msg = f"Can't decode JSON: {e}"
+            LOGGER.exception(error_msg)
```
hoh, line 75: Isn't this already logged by the caller ?

src/aleph/storage.py
```diff
@@ -22,113 +24,207 @@
...
+    enabled_clients = app["config"].p2p.clients.value
+
+    if "protocol" in enabled_clients:
+        streamer = get_streamer()
+        content = await streamer.request_hash(content_hash)
+
+    if content is None and "http" in enabled_clients:
+        content = await p2p_http_request_hash(content_hash, timeout=timeout)
+
+    if content is not None:
```
hoh, line 103: When would content be `None`, and how should that be handled ?

src/aleph/storage.py
```diff
@@ -22,113 +24,207 @@
...
+    Computes the IPFS hash of the content.
+    :param content: Content to hash.
+    :param expected_hash: Expected hash of the content.
+    :return: The computed hash of the content. Can return None if the operation fails for some reason.
+    """
+
+    # TODO: get a better way to compare hashes (without depending on the IPFS daemon)
+    try:
+        cid_version = 0
+        if len(expected_hash) >= 58:
```
hoh, line 122: > If a CID is 46 characters starting with "Qm", it's a CIDv0
https://docs.ipfs.io/concepts/content-addressing/#identifier-formats

src/aleph/storage.py
```diff
@@ -22,113 +24,207 @@
...
+    return content
+
+
+async def compute_content_hash_ipfs(
+    content: bytes, expected_hash: str
+) -> Optional[str]:
+    """
+    Computes the IPFS hash of the content.
+    :param content: Content to hash.
+    :param expected_hash: Expected hash of the content.
```
hoh, line 115: It seems strange to have a function `compute_hash` expecting a mandatory `expected_hash`. Should it be optional ? 

Or should the functions be named `compare_content_hash_ipfs` or `validate_content_hash_ipfs` ?
I feel like this splitting between `verify_content_hash` and `compute_content_hash_ipfs` is odd, and it could have more sense to have:

```python

 async def verify_content_hash_ipfs(content: bytes, expected_hash: str):
    ...

 async def verify_content_hash_sha256(content: bytes, expected_hash: str):
    ...

 async def verify_content_hash(content: bytes, engine: ItemType, expected_hash: str):
    ...
```

src/aleph/storage.py
```diff
@@ -22,113 +24,207 @@
...
+        computed_hash = await add_ipfs_bytes(content, cid_version=cid_version)
+        return computed_hash
+
+    except asyncio.TimeoutError:
+        LOGGER.warning(f"Timeout while computing IPFS hash for '{expected_hash}'.")
+        return None
+
+
+async def compute_content_hash_sha256(content: bytes, expected_hash: str) -> str:
+    return get_sha256(content)
```
hoh, line 134: Should this call run in a thread (`asyncio.run_in_executor`) ?

src/aleph/chains/common.py
```diff
@@ -365,37 +366,41 @@ async def get_chaindata_messages(chaindata, context, seen_ids: Optional[List]=No
...
-        else:
+            error_msg = f"Can't get content of offchain object {chaindata['context']!r}"
+            LOGGER.exception(error_msg)
+            raise ContentCurrentlyUnavailable(error_msg)
+
+        try:
+            messages = await get_chaindata_messages(content.value, context)
+        except AlephStorageException:
             LOGGER.debug("Got no message")
+            raise
```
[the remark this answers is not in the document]
odesenfans: `raise` is fine, we do not have more information to add here besides the exception itself (content invalid/unavailable).

src/aleph/exceptions.py
```diff
@@ -2,6 +2,14 @@ class AlephException(Exception):
     ...
 
 
+class AlephStorageException(AlephException):
```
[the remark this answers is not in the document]
odesenfans: It might be useful later, and it's not much of an overhead so I'd keep it.

src/aleph/handlers/storage.py
```diff
@@ -35,7 +36,13 @@ async def handle_new_storage(message, content):
         return -1  # not allowed, ignore.
 
     is_folder = False
-    item_hash = content["item_hash"]
+    try:
+        item_hash = content["item_hash"]
+    except KeyError as e:
+        error_msg = f"Could not load item_hash / message: {message} - content {content}"
+        LOGGER.error(error_msg)
+        raise Exception(error_msg) from e
```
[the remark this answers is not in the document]
odesenfans: Ah yes this code is not supposed to be here, it's just some logging I added to debug an issue.

src/aleph/storage.py
odesenfans: Originally I used it to store the raw content, but then I realized it was more convenient to just store bytes since not all JSON contents originate from raw contents. Fixing.

src/aleph/storage.py
odesenfans: `content` is None if none of the previous methods returned valid data. This can happen if the file is simply not on the P2P network, e.g. if the file is on IPFS or the user forgot to upload it.

If the file is not on IPFS either, the caller throws a `ContentUnavailable` exception.

src/aleph/storage.py
odesenfans: No, not necessarily and not with that level of detail. I'm not sure it should be logged at exception level though, it's not an issue with our code but with the content of the message.

src/aleph/storage.py
odesenfans: To be honest I don't really see a point of running computing-intensive tasks in the executor, at least if we use the `ThreadPoolExecutor`. It just adds overhead and no parallelization whatsoever.

If you want we can investigate creating a task pool for CPU-bound tasks running on a `ProcessPoolExecutor`, but I'm really doubtful about the efficiency for most messages.

src/aleph/storage.py
odesenfans: I tried this just now, it makes mypy angry about a lot of lines so I think I'll make the change in a different PR.

src/aleph/storage.py
odesenfans: I split it this way to reduce duplication. The error handling is the same for sha256 and IPFS hashes, we just use a different function to compute them. I'm not 100% happy about the signature either but IMO it's worth it.

src/aleph/storage.py
hoh: I am reading this code again, and don't get why not returning the value instead of storing it in a temporary variable.

```python
    return await p2p_http_request_hash(content_hash, timeout=timeout)
```

odesenfans: Because this function verifies the content hash at the end, so we need a `if content is not None: verify_content_hash(content, ...)` block.

src/aleph/chains/common.py
```diff
@@ -365,37 +366,42 @@ async def get_chaindata_messages(chaindata, context, seen_ids: Optional[List]=No
...
+                )
+                await asyncio.wait_for(pin_hash(chaindata["content"]), timeout=4.0)
+            except asyncio.TimeoutError:
+                LOGGER.warning(f"Can't pin hash {chaindata['content']}")
         return messages
     else:
-        LOGGER.info("Got unknown protocol/version object in tx %r" % context)
-        return -1
+        error_msg = f"Got unknown protocol/version object in tx {context!r}"
+        LOGGER.info("%s", error_msg)
```
hoh, line 403:
```suggestion
        LOGGER.info(""Got unknown protocol/version object in tx '%s'", f"{error_msg!r}")
```
```

src/aleph/storage.py
hoh: OK

src/aleph/chains/common.py
odesenfans: `f"{error_msg!r}"` -> `repr(error_msg)`? Or  is there a reason to use an f-string I'm not aware of?

hoh: That's better indeed. I naively copied from your string.

PR #199, opened 2022-02-28
[Storage] Use Aleph message model for validation

odesenfans:
Mirrored the validation system used for forget messages in store
messages. Used the `StoreMessage` class to validate the message
and content dictionaries.

We now discard the message if it fails validation instead of
retrying it through the default exception handler. This fixes
an issue where invalid store messages would be retried indefinitely.

Added unit tests for the storage handler.

### DIFF
--- a/src/aleph/handlers/storage.py
+++ b/src/aleph/handlers/storage.py
@@ -12,6 +12,9 @@ import logging
 
 import aioipfs
 import asyncio
+from typing import Dict
+from aleph_message.models import StoreMessage
+from pydantic import ValidationError
 
 from aioipfs import InvalidCIDError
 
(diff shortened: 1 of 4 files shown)


odesenfans:
Depends on #196.

src/aleph/handlers/storage.py
```diff
@@ -24,24 +27,27 @@
...
     if not app["config"].storage.store_files.value:
         return True  # Ignore
 
-    item_type = content.get("item_type")
+    # TODO: ideally the content should be transformed earlier, but this requires more clean up
+    #       (ex: no more in place modification of content, simplification of the flow)
+    try:
+        store_message = StoreMessage(**message, content=content)
+    except ValidationError as e:
+        print(e)
```
hoh, line 39: Should we raise an exception here and follow up on invalid messages reported on Sentry during testing ?

odesenfans: Mmh I don't think Sentry is the right tool for this, because this is not something that we can fix. IMO right approach is to log invalid messages in a dedicated table and have some monitoring tool (like the one we discussed yesterday) pick up the info from an API.

Issue #198, opened 2022-02-28
Warning about msgpack not installed confuses node operators


hoh:
When launching pyaleph, especially for operations such as key generation, node operators get confused by the following warnings:

```
ujson module not found, using json
msgpack not installed, MsgPackSerializer unavailable
```

These appear to come from a dependency of PyAleph and not PyAleph itself.

We have three possibilities to hide these:
- Fix it upstream
- Change the log level for the dependencies displaying the warnings
- Installing `ujson` and `msgpack` in PyAleph.

Note that `ujson` would probably increase the performance of the JSON packing and unpacking.
msgpack is only used by [aleph-vm](https://github.com/aleph-im/aleph-vm), not by the CCN.

odesenfans:
I'll have a look.

odesenfans:
Ok so both warnings come from `aiocache`. The reasonable way to fix these IMO is to install the missing packages.

hoh:
OK, let's do that !

odesenfans:
Resolved by #200.

PR #200, opened 2022-02-28
[Dependencies] Install ujson and msgpack for aiocache

odesenfans:
Fixed warnings coming from aiocache complaining about not being
able to import ujson and msgpack.


odesenfans:
Resolves #198.

PR #202, opened 2022-03-01
[Forget] Fix issue with inline messages

odesenfans:
Fixed an invalid condition that caused forgetting inline messages
to fail because they do not have an `item_type` field.

Added basic unit tests for the functionality.


odesenfans:
Depends on #199.

PR #205, opened 2022-03-09
[Fix] Buffer P2P incomplete read errors

odesenfans:
Implemented a workaround to avoid spamming exception logs when
the communication with the P2P daemon enters an invalid state
and spams IncompleteRead exceptions. The exceptions are now
buffered until reaching a threshold (30000 exceptions).

### DIFF
--- a/src/aleph/services/p2p/protocol.py
+++ b/src/aleph/services/p2p/protocol.py
@@ -6,6 +6,7 @@ import random
 from typing import Any, Dict, Optional, Set
 
 from anyio.abc import SocketStream
+from anyio.exceptions import IncompleteRead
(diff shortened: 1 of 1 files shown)


hoh:
Would it not make sense to log the first time the exception happens ?

Something like
```python
        except IncompleteRead:
            if not (incomplete_read_counter % incomplete_read_threshold):
                LOGGER.exception(
                    "Incomplete read (%d times), reconnecting. Try to restart the application.",
                    incomplete_read_counter,
                )
            incomplete_read_counter += 1
```

odesenfans:
Ok for me, I took your implementation.

PR #210, opened 2022-03-11
[Typing] Make shared stats non-optional

odesenfans:
The message processing loop now requires the `shared_stats` dict
to be non-null.


hoh:
Not sure what problem this fixes, but nothing against it.

PR #212, opened 2022-03-21
[Fix] API servers as managed shared variable

odesenfans:
Fixed an issue where the API servers found by the refresh job
were never updated in the process in charge of retrying messages.
This was caused by the use of the `api_servers` singleton, which
was not allocated in shared memory.

The variable is now allocated from shared memory using the shared
memory manager, similarly to what is done for shared stats.


odesenfans:
Depends on #210.

Issue #211, opened 2022-03-11
Very high usage of container PyAleph since v0.2.0 from file `/tmp/messages_task_loop.log`


hoh:
Reported by Tg users `aiz` and `Arimav`, reproduced by @hoh .

`/var/lib/docker/overlay2` increases sharply to more than a hundred GB.

The easiest fix in production is to restart the pyaleph container.
```
docker-compose down
docker-compose up
```

> Executing `truncate -s 0 /tmp/messages_task_loop.log` inside the container raises the error `truncate: cannot open '/tmp/messages_task_loop.log' for writing: Permission denied`

The issue comes from verbose logging of a PyAleph process.
It may be the case that this logging setup ignores the default logging level (WARNING) and still writes DEBUG level logs.

odesenfans:
After investigation, setting the log level on the CLI works as expected. The log level is correctly propagated to the subprocesses. The only configuration in which this is not true is if the user specifies:

```
logging:
    level: 10
```
in his config.yml file, apparently this overrides the CLI value for subprocesses. It should not, I will propose a fix.

Besides that, I'll change the default log level to INFO in the Docker Compose files.

PR #216, opened 2022-03-24
[Logging] Configuration improvements

odesenfans:
* Made the CLI logging level override the value from the config
  file, if specified
* Subprocesses now use the same logging setup function to configure
  the format of the logs
* Changed the default log level to INFO in Docker Compose files.


odesenfans:
This should help with the issue described in #211.

PR #218, opened 2022-03-25
[P2P] Reduce amount of logs in case of errors

odesenfans:
Improved the threshold mechanism for incomplete read errors linked
to the P2P daemon that sometimes occur over long periods of time.

* Added a sleep of 2 seconds before the next reconnection attempt
* Adapted the threshold so that a log is sent approximately every
  10 minutes
* Added a retry to some monitoring jobs that were not restarted
  after the first failure.

### DIFF
--- a/src/aleph/services/peers/monitor.py
+++ b/src/aleph/services/peers/monitor.py
@@ -1,8 +1,10 @@
+import asyncio
 import json
 import logging
 from typing import Any, Dict
 from urllib.parse import unquote
 
+from anyio.exceptions import IncompleteRead
 from p2pclient import Client as P2PClient
 from p2pclient.pb.p2pd_pb2 import PSMessage
 from p2pclient.utils import read_pbmsg_safe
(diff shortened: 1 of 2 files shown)


src/aleph/services/peers/monitor.py
```diff
@@ -51,21 +53,38 @@ async def handle_incoming_host(pubsub_msg: Dict[str, Any], source: Protocol = Pr
...
-        async for mvalue in sub_ipfs(IPFS_ALIVE_TOPIC):
-            await handle_incoming_host(mvalue, source=Protocol.IPFS)
-    except Exception:
-        LOGGER.exception("Exception in pubsub peers monitoring, resubscribing")
+    while True:
+        try:
+            async for mvalue in sub_ipfs(IPFS_ALIVE_TOPIC):
+                await handle_incoming_host(mvalue, source=Protocol.IPFS)
+        except Exception:
+            LOGGER.exception("Exception in pubsub peers monitoring, resubscribing")
```
hoh, line 90: No sleep here ?

hoh (commented):
LGTM, except the question about adding a sleep() in the third `while True`.

src/aleph/services/peers/monitor.py
odesenfans: I saw no reason to add it here, we do not have this spurious logging issue with IPFS pubsubs so far. I just noticed that the job would die and never be restarted on error.

PR #231, opened 2022-04-07
 [Config] Remove dependency to API in non-API code

odesenfans:
Removed all dependencies to the aiohttp app object in non-API code.
This object was used all over the place to retrieve the config
object. Instead, we now use a global variable to store the config
object itself.

### DIFF
--- a/src/aleph/config.py
+++ b/src/aleph/config.py
@@ -1,5 +1,7 @@
 import logging
 
+from configmanager import Config
+
 
 def get_defaults():
     return {
@@ -85,3 +87,10 @@ def get_defaults():
             "traces_sample_rate": None,
         },
     }
+
+
+app_config = Config(schema=get_defaults())
+
+
+def get_config() -> Config:
+    return app_config
--- a/src/aleph/storage.py
+++ b/src/aleph/storage.py
@@ -19,8 +19,8 @@ from aleph.services.p2p.http import request_hash as p2p_http_request_hash
 from aleph.services.p2p.singleton import get_streamer
 from aleph.types import ItemType
 from aleph.utils import run_in_executor, get_sha256
-from aleph.web import app
 from aleph.services.ipfs.common import get_cid_version
+from aleph.config import get_config
 
 LOGGER = logging.getLogger("STORAGE")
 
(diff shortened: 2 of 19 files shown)


odesenfans:
Depends on #229.

src/aleph/config.py
```diff
@@ -83,3 +85,10 @@ def get_defaults():
             "traces_sample_rate": None,
         },
     }
+
+
+app_config = Config(schema=get_defaults())
```
hoh, line 90: I see multiple uses of `aleph.config.app_config`. Why not using `aleph.config.get_config()` everywhere and marking this variable as private with a `_` prefix ?

hoh: Is this variable useful versus using `@lru_cache` on `def get_config()` ?

src/aleph/storage.py
```diff
@@ -145,7 +146,8 @@ async def verify_content_hash(
     Checks that the hash of a content we fetched from the network matches the expected hash.
     :return: True if the hashes match, False otherwise.
     """
-    ipfs_enabled = app["config"].ipfs.enabled.value
+    config = get_config()
+    ipfs_enabled = config.ipfs.enabled.value
```
hoh, line 150: Why not `get_config().ipfs.enabled.value` ?

hoh (commented):
Great fix.
A few comments I'd like to discuss.

src/aleph/config.py
odesenfans: It conveys the intent better IMO. Using LRU cache when there's only one return value seems a bit counterintuitive.

src/aleph/storage.py
odesenfans: While coding, my train of thought was 1) get the config and 2) use it where necessary (there are several uses of the config object in most cases). Your suggestion could be useful once we have a dynamic configuration system, reloadable at runtime. Until we have that, I'd keep the local variable.

src/aleph/config.py
hoh: What about the first comment ?

odesenfans: I think the usage in job_utils.py is legit as we need to initialize the global variable. The other one in `get_ipfs_api` looks like an oversight from my part.

PR #232, opened 2022-04-07
[Tests] Database support in tests

odesenfans:
Added a fixture to initialize the database. This fixture initializes
the collections used by a CCN in a separate database and resets
it on every run.

This feature will allow to write more extensive tests by preloading
fixture data in the DB.

Implemented a first test using the fixture for the pending TX job.
On the CI side, we now spawn a service container to run MongoDB.


hoh (commented):
Looks good ! 
A small note on filepath that could be simplified.

hoh (approved):

PR #248, opened 2022-05-04
[Logging] Specify max log file size

odesenfans:
Set the default maximum log file size to 1GB.

### DIFF
--- a/src/aleph/config.py
+++ b/src/aleph/config.py
@@ -7,7 +7,7 @@ def get_defaults():
     return {
         "logging": {
             "level": logging.WARNING,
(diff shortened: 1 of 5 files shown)


src/aleph/config.py
```diff
@@ -7,6 +7,7 @@ def get_defaults():
     return {
         "logging": {
             "level": logging.WARNING,
+            "max_log_file_size": 1 * 1024 * 1024 * 1024,
```
hoh, line 10: Would it not be more clear like this ?
```suggestion
            "max_log_file_size": 1_000_000_000,  # 1 GB
```

odesenfans: Yes, I applied your suggestion.

PR #249, opened 2022-05-05
[Forget] Multi-users test

odesenfans:
Added a test that tries to forget a file stored by multiple users.


src/aleph/chains/common.py
```diff
@@ -258,7 +258,8 @@ async def incoming(
         except UnknownHashError:
             LOGGER.warning(f"Invalid IPFS hash for message {item_hash}, won't retry.")
             return IncomingStatus.FAILED_PERMANENTLY, []
-        except Exception:
+        except Exception as e:
+            print(e)
```
hoh, line 262: Why ?

hoh (approved):

src/aleph/chains/common.py
odesenfans: Just a quick and dirty change I forgot to revert. Fixing it.

PR #250, opened 2022-05-05
[Storage] Garbage collector

odesenfans:
Added a new process in charge of deleting files from local storage.
Files can now be marked for deletion by being listed in
the scheduled_deletions collection. The garbage collector process
will periodically look up this collection and delete all
the files for which the `delete_by` datetime field is in the past.

Files are now automatically marked for deletion when the user posts
them using the /storage/add_json and /storage/add_file endpoints.
The deletion is cancelled if a user creates a message using this
content before a given period of time (one hour by default).

Added a migration script that goes through all the files currently
stored on a CCN and schedules all the files not related to an Aleph
message for deletion.


odesenfans:
To discuss:
* [ ] Ideally, we should have integration tests
* [ ] Do we want to use the same mechanism for the garbage collection linked to FORGET messages? (If yes, can be a separate PR)
* [ ] Need to analyze the results of the migration script on a system with a fully synchronized node.

odesenfans:
Replaced by #269, as running a GC automatically on the main file collection might be a bit dangerous for user data.

PR #251, opened 2022-05-05
[TXs] Increase IPFS pin timeout to 2 minutes

odesenfans:
Fixed a warning that occurs in the TX processing loop. The logs
warn about not being able to pin the chaindata files:

Can't pin hash Qm...

The files end up being pinned anyway, the operation just takes
longer than 4 seconds. As we now know, some IPFS fetches can take
up to 100s of seconds. Increased the timeout to 2 minutes to
silence the warning.

### DIFF
--- a/src/aleph/chains/common.py
+++ b/src/aleph/chains/common.py
@@ -393,7 +393,6 @@ async def get_chaindata_messages(
 
         LOGGER.info("Got bulk data with %d items" % len(messages))
         if config.ipfs.enabled.value:
-            # wait for 4 seconds to try to pin that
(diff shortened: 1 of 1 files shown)


hoh:
Are 2 minutes enough for very large files ?

hoh (approved):
LGTM, just wondering if 120s is enough when 100s is reached in testing.

odesenfans:
100s is the number that was cited at P2P Paris, I'm taking it as an upper bound if everything goes wrong. The file will be pinned in any case.

PR #258, opened 2022-05-09
[Permissions] Process unauthorized messages gracefully

odesenfans:
Fixed an issue where the check of permissions would fail if
the content.address field of the message was set to a different
address than the sender and the owner of this address did not
authorize the sender to post message on their behalf.

Added basic unit tests for the permission system.


src/aleph/model/messages.py
```diff
@@ -139,17 +144,20 @@ async def get_computed_address_aggregates(address_list=None, key_list=None, limi
         {"$addFields": {"address": "$_id", "contents": {"$arrayToObject": "$items"}}},
         {"$project": {"_id": 0, "address": 1, "contents": 1}},
     ]
+
+    # Note: type-ignore statements are required because mypy does not understand that
+    # the nested dictionaries we declared just above are dictionaries.
```
hoh, line 149: Could this be fixed with type annotation on the variable ?

odesenfans: Technically yes, with [TypedDict](https://docs.python.org/3/library/typing.html#typing.TypedDict). But it's gonna be a PITA to maintain so I'd rather keep the type ignores. Also, there's an off-chance that this will someday be fixed in mypy. That day, the type ignore statements will trigger errors so we will know :stuck_out_tongue:

PR #260, opened 2022-05-09
[Tests] Make API endpoints testable

odesenfans:
It is now possible to test API endpoints with the test client of aiohttp.
Added an example test that checks that the /version endpoint returns
the expected version.

Refactored the declaration of endpoints in the initialization of
the app. We now have a `register_routes` function centralized in
a single module. All references to the global app object inside the
implementation of endpoints were replaced by accesses to `request.app`.
This allows to import the app object, configured properly, without
importing other files.


src/aleph/web/controllers/__init__.py
```diff
@@ -1,13 +1,62 @@
...
     stats,
-    channels,
-    listener,
-    info,
+    storage,
+    version,
 )
+
+
+def register_routes(app: web.Application):
```
hoh, line 19: +1 to centralize all routes. Should we also move them to a new file `src/aleph/web/routes.py` to make it easier to find ?

tests/api/test_version.py
```diff
@@ -0,0 +1,15 @@
...
+
+@pytest.mark.asyncio
+async def test_get_version(aiohttp_client):
+    client = await aiohttp_client(app)
+
+    response = await client.get("/api/v0/version")
+    assert response.status == 200, await response.text()
+
+    data = await response.json()
+    assert data["version"] == get_git_version()
```
hoh, line 15: `""/api/v0/version"` may return either `get_distribution(...).version` or `get_git_version()`, so this could be inconsistent depending on the testing environment.

src/aleph/web/controllers/version.py
```diff
@@ -0,0 +1,9 @@
+from aiohttp import web
+from aleph import __version__
```
hoh, line 2: We should decide if the main way to get the version is through `aleph.__version__` or `aleph.get_git_version()`. If we go for the first option, I would rename `get_git_version` -> `_get_git_version` to mark it as private.

src/aleph/web/controllers/__init__.py
odesenfans: Ok for me.

src/aleph/web/controllers/version.py
odesenfans: Works for me, __version__ is more Pythonic IMO.

PR #262, opened 2022-05-10
[Storage] Remove filename parameter

odesenfans:
The name of the file uploaded through IPFS and storage endpoints
is useless because we only rely on hashes.


odesenfans:
I asked myself the same question, I'll change it another PR.

PR #268, opened 2022-05-15
[CLI] New CLI tool for node operators

odesenfans:
We now provide a CLI tool that integrates all the operations
commonly performed by node operators. Currently, this CLI
allows to:

* generate private keys for the node, replacing a functionality
  that was implemented in the CCN main app directly.
* run migrations, replacing the config updater script.

PR #269, opened 2022-05-15
[Storage] Garbage collector CLI

odesenfans:
Added a new command to the CCN operator CLI to run a garbage
collector on local storage. The new `gc run` command lists
all the files that are not linked to any message or permanent
pin and deletes them.

Using the --verbose option, the command will print more details
on the files it will preserve and delete. The --dry-run option
allows to run the GC without actually deleting any file.


odesenfans:
Depends on #268. Replaces #250.

PR #270, opened 2022-05-15
[Messages] Check the presence of all fields

odesenfans:
Broadened the check of fields in new messages. We now check for
the presence of all the authorized fields, as parts of the code
expect them later anyway.

This fixes an issue detected on Sentry where the missing "time"
field on some messages triggered an exception.


odesenfans:
Superseded by #272, but we can imagine a situation where this could be merged and released earlier.

hoh (dismissed):

src/aleph/network.py
```diff
@@ -78,7 +78,7 @@ async def check_message(
 
     if "item_hash" not in message:
         raise InvalidMessageError("Missing field 'item_hash' in message")
```
hoh, line 80: There is duplication with this code.

hoh (changes requested):

hoh (approved):

PR #276, opened 2022-05-17
[Forget] Fetch less data

odesenfans:
Reduced the data usage of the forget handler in two ways.

1. Stop relying on `get_message_content` to retrieve the content
of a message. As the message is guaranteed to already be on
the node, we can directly get the parts of interest from the DB.

2. Only retrieve the fields relevant to the handling of the forget
message from the DB. Rather than retrieving the whole content,
which can be quite large, we now only retrieve the address, item
type and item hash from the DB. This should improve the performance
of the handler. Introduced a new class, `TargetMessageInfo`, to
represent the data we fetch from the DB.

### DIFF
--- a/src/aleph/handlers/forget.py
+++ b/src/aleph/handlers/forget.py
@@ -1,16 +1,47 @@
 import logging
-from typing import Dict
+from dataclasses import dataclass
+from typing import Dict, Optional, List
 
 from aioipfs.api import RepoAPI
 from aioipfs.exceptions import NotPinnedError
-from aleph_message.models import ForgetMessage, ItemType, MessageType, StoreContent
+from aleph_message.models import ForgetMessage, MessageType
+from aleph_message.models import ItemType
 
 from aleph.model.filepin import PermanentPin
 from aleph.model.hashes import delete_value
 from aleph.model.messages import Message
 from aleph.services.ipfs.common import get_ipfs_api
-from aleph.storage import get_message_content
 from aleph.utils import item_type_from_hash
+
+
+@dataclass
+class TargetMessageInfo:
+    item_hash: str
+    sender: str
+    type: MessageType
+    forgotten_by: List[str]
+    content_address: Optional[str]
+    content_item_hash: Optional[str]
+    content_item_type: Optional[ItemType]
+
+    @classmethod
+    def from_db_object(cls, message_dict: Dict) -> "TargetMessageInfo":
+        content = message_dict.get("content", {})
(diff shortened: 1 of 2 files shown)


src/aleph/handlers/forget.py
```diff
@@ -1,17 +1,48 @@
...
+    content_address: Optional[str]
+    content_item_hash: Optional[str]
+    content_item_type: Optional[ItemType]
+
+    @classmethod
+    def from_db_object(cls, message_dict: Dict) -> "TargetMessageInfo":
+        content = message_dict.get("content", {})
+        content_item_type = content.get("item_type")
+
+        if content_item_type is not None:
```
hoh, line 32: What if it is None ?

src/aleph/handlers/forget.py
```diff
@@ -119,15 +166,46 @@ async def forget_if_allowed(target_message: Dict, forget_message: ForgetMessage)
...
-    if target_message.get("type") == MessageType.store:
-        store_content = StoreContent(**target_message["content"])
-        await garbage_collect(store_content.item_hash, store_content.item_type)
+    if target_info.type == MessageType.store:
+        if (
+            target_info.content_item_type is None
+            or target_info.content_item_hash is None
+        ):
+            raise ValueError(
+                f"Could not garbage collect content linked to STORE message {target_hash}."
```
hoh, line 183: Can you make this message more explicit ? Something like:
```suggestion
                f"Information missing, could not garbage collect content linked to STORE message {target_hash}."
```

hoh, line 183: Why put the target_hash in the string and not as another argument ?
```suggestion
                f"Could not garbage collect content linked to STORE message", target_hash
```

hoh (commented):
I am not a big fan of redefining multiple Message types recent PRs, but that can always be improved later and this PR improves over the current situation.

A few minor comments but nothing blocking or consequent.

src/aleph/handlers/forget.py
```diff
@@ -1,17 +1,48 @@
...
+    item_hash: str
+    sender: str
+    type: MessageType
+    forgotten_by: List[str]
+    content_address: Optional[str]
+    content_item_hash: Optional[str]
+    content_item_type: Optional[ItemType]
+
+    @classmethod
+    def from_db_object(cls, message_dict: Dict) -> "TargetMessageInfo":
```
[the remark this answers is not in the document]
odesenfans: Sold, I always forget about this one.

PR #277, opened 2022-05-17
[Messages] Use pending message model everywhere

odesenfans:
Propagated the use of the `BasePendingMessage` subclasses in all
functions related to message processing.

Split the message validation function, check_message, in two parts:
* `parse_message` checks that the message is semantically valid,
  i.e. that all the required fields are present, are of the correct
  type and have sensible values.
* `verify_signature` checks the signature of the message using
  the public key of the sender.

We now call parse_message as early as possible in the process.
This allows to simplify hypotheses about the presence/absence
of specific fields and allows to simplify the codebase.

One of the current issues is that the content of the message
can only be loaded conditionally (if the message has inline content).
Otherwise, we must load the content from the network. As this
operation goes beyond simple validation, we perform it later
in the flow, leading to a situation where the content of a message
is not guaranteed to be available at a given point depending
on the `item_type` field of the message.

Modified tests to use the raw message classes. Removed a few tests
that tested corner cases linked to incomplete message dictionaries.

### DIFF
--- a/src/aleph/chains/common.py
+++ b/src/aleph/chains/common.py
@@ -5,7 +5,6 @@ from dataclasses import asdict
 from enum import IntEnum
 from typing import Dict, Optional, Tuple, List
 
-from aleph_message.models import MessageType
 from bson import ObjectId
 from pymongo import UpdateOne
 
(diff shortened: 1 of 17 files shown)


odesenfans:
To discuss with @hoh (not urgent): this PR as it stands is a good improvement already because it forces the usage of Pydantic models in the processing of messages. However, it doesn't implement the last mile, i.e. using Pydantic models to write to the DB. We still maintain dictionaries by hand towards the end of the job.

I plan to add new models to represent the documents in the `messages` collection, but that represents some additional work and I want to implement this in another PR. Ok for you?

PR #284, opened 2022-05-23
[API] Add content hash query parameter for messages

odesenfans:
Users can now specify one or more content hashes to filter
messages in the message API with the new `contentHashes`
query parameter. This new parameter filters messages
on their `content.item_hash` field.


odesenfans:
Depends on #282.

hoh (approved):

PR #305, opened 2022-07-07
[API] Validate messages on API POST

odesenfans:
Messages published by users using the P2P/IPFS pubsub endpoint
are now validated against the `PendingMessage` Pydantic model.
This will let users detect if their messages are constructed
properly without propagating bad messages to the rest of
the network.


src/aleph/web/controllers/p2p.py
```diff
@@ -11,21 +16,44 @@
...
+    topic = request_data.get("topic")
+    data = request_data.get("data")
+
+    # Currently, we only check validate messages
+    message_topic = config.aleph.queue_topic.value
+    if topic == message_topic:
+        try:
+            _ = parse_message(data)
+        except InvalidMessageError as e:
+            raise web.HTTPUnprocessableEntity(text=str(e))
```
hoh, line 37: The error message should, when the issue was raised by Pydantic, use the content from ValidationError.json.

odesenfans: Okay sure, fine by me.

hoh (approved):

PR #312, opened 2022-08-16
fix: (minor) Validation typo

BjrInt:
Display "Avalanche" instead of "Solana" when validations fails for an Avalanche message.


MHHukiewitz (approved):
LGTM

PR #314, opened 2022-08-18
Hoh fix float dates

### LINKED ISSUE #313: API crashes when requesting float values for startDate and endDate
## Problem

Calling the failing URL:
http://api1.aleph.im:4024/api/v0/messages.json?pagination=200&page=1&msgType=PROGRAM&startDate=1660828440.0&endDate=1660828500.0

Returns the error message: `invalid literal for int() with base 10: '1660828440.0'`

This URL works (no message but no error message):
http://api1.aleph.im:4024/api/v0/messages.json?pagination=200&page=1&msgType=PROGRAM&startDate=1660828440&endDate=1660828500

## Diagnostic

Calling `int()` on a float is not supported by Python. Ex: `int(0.5)`.

Relevant code:
https://github.com/aleph-im/pyaleph/blob/26c1e30eff26909d0f4c6c07f0f4d799073588c4/src/aleph/web/controllers/utils.py#L71-L72

## Proposal

Replace `int()` with `float()`.

Floating datetimes may be required for some usecases, so this should be fixed server side and not client side.

hoh:
Fixes #313 ; does not change the underlying logic even if tempted to ^^

### DIFF
--- a/src/aleph/web/controllers/utils.py
+++ b/src/aleph/web/controllers/utils.py
@@ -71,11 +71,16 @@ def prepare_date_filters(request, filter_key):
     start_date = float(request.query.get("startDate", 0))
     end_date = float(request.query.get("endDate", 0))
 
-    if start_date > 0:
+    if start_date < 0:
+        raise ValueError("startDate field may not be negative")
+    if end_date < 0:
+        raise ValueError("endDate field may not be negative")
+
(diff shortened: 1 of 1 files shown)


odesenfans (approved):
Ok for me. The logic is ugly indeed, but let's worry about that another day.

PR #318, opened 2022-08-26
Fix: `get_verification_buffer` is a sync function

odesenfans:
Removed the useless async qualifier.

### DIFF
--- a/src/aleph/chains/avalanche.py
+++ b/src/aleph/chains/avalanche.py
@@ -70,7 +70,7 @@ async def verify_signature(message: BasePendingMessage):
         return False
 
     try:
-        verification = await get_verification_buffer(message)
+        verification = get_verification_buffer(message)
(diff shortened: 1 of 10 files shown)


moshemalawach:
having the function async allowed the scheduler to allow other stuff between calls, unless I'm mistaken?

odesenfans:
To the best of my understanding, marking a sync function as async does not do anything. The only time when the task scheduler gets involved is if the async function yields. A blocking function (=one without await) never yields (in other terms, "it's always ready"), therefore there is no difference whether the function is marked as sync or async.

hoh:
I confirm @odesenfans 's description of the behavior. In order to allow another task to run, developers must `await` for something truly asynchronous and not just synchronous code (some code uses `await asyncio.sleep(0)`, or use `loop.run_in_executor(...)`.

PR #331, opened 2022-10-03
Fix: delete chain data files that could not be broadcast

odesenfans:
Problem: if the chain data push operation fails, the associated file is not deleted from the node. As failures can occur repeatedly (ex: insufficient balance), such files end up taking an unreasonable amount of storage space.

Solution: delete the chain data file from local storage if the broadcast operation fails.

We do not delete the file from IPFS at the moment, as for IPFS only a few blocks are added on each failed push.

### DIFF
--- a/src/aleph/chains/ethereum.py
+++ b/src/aleph/chains/ethereum.py
@@ -29,6 +29,7 @@ from aleph.register_chain import (
 )
 from aleph.utils import run_in_executor
 from .tx_context import TxContext
+from ..model.hashes import delete_value
 from ..schemas.pending_messages import BasePendingMessage
 
 LOGGER = logging.getLogger("chains.ethereum")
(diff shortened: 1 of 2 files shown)


src/aleph/chains/ethereum.py
```diff
@@ -304,18 +305,24 @@ async def ethereum_packer(config):
...
+                    config,
+                    contract,
+                    web3,
+                    account,
+                    int(gas_price * 1.1),
+                    nonce,
+                    content,
+                )
+                LOGGER.info("Broadcasted %r on %s" % (response, CHAIN_NAME))
+            except Exception:
```
hoh, line 321: Can you avoid catching _any_ exception silently ? 
Whitelist the exception classes and/or log them so we can catch them in Sentry.

odesenfans: After some tests, an insufficient balance will result in a ValueError. I think the safest approach is to catch Exception, drop the file and then re-raise the exception.

hoh: OK with the re-raise

PR #333, opened 2022-10-04
 Feature: CLI command to repair local storage

odesenfans:
Problem: the local storage of a CCN may get incoherent or corrupted
because of issues like downtime or maintenance gone wrong.

Solution: a new CLI command, 'repair'. This command checks the DB
to determine the files that should be stored on the node and
fetches them from the network.


odesenfans:
Depends on #269.

PR #335, opened 2022-10-06
Feature: return errors for invalid message API queries

odesenfans:
Problem: the messages.json endpoint was too permissive, allowing users to specify invalid chains, message types, make incoherent pagination / time filtering, etc.

Solution: detect these cases and return a 422 error.

Replaced the custom validation code of the endpoint by a Pydantic model. Added tests for start/end date parameters and pagination.

Breaking changes:
* The "endDate" field is now considered as exclusive. Only messages with a time field strictly lower than the value will be returned, instead of a value lower or equal before.
* The endpoint now returns a 422 error instead of a 400 when incorrect parameters are passed.

Moreover, a 422 error code will now be returned in the following situations, where the previous implementation would simply return a 200:
* if an invalid/unknown chain appears in the "chains".
* if an invalid/unknown message type is specified in "msgType".
* if an invalid item hash (=not a hexadecimal sha256, CIDv0 or CIDv1) is specified in the "hashes" or "contentHashes" field.
* if the "endDate" field is lower than the "startDate" field.
* if "endDate" or "startDate" are negative.
* if pagination parameters ("page" and "pagination") are negative.

### DIFF
--- a/src/aleph/web/controllers/messages.py
+++ b/src/aleph/web/controllers/messages.py
@@ -1,129 +1,200 @@
-from typing import Any, Dict, List, Optional, Set
+import asyncio
+import logging
+from enum import IntEnum
+from typing import Any, Dict, List, Optional, Mapping
+
+from aiohttp import web
+from aleph_message.models import MessageType, ItemHash, Chain
+from bson.objectid import ObjectId
+from pydantic import BaseModel, Field, validator, ValidationError, root_validator
+from pymongo.cursor import CursorType
 
 from aleph.model.messages import CappedMessage, Message
-from aiohttp import web
-from aiohttp.web_exceptions import HTTPBadRequest
-import asyncio
-from pymongo.cursor import CursorType
-from bson.objectid import ObjectId
-from aleph.web.controllers.utils import Pagination, cond_output, prepare_date_filters
-import logging
+from aleph.web.controllers.utils import (
+    LIST_FIELD_SEPARATOR,
+    Pagination,
+    cond_output,
+    make_date_filters,
+)
 
-LOGGER = logging.getLogger("MESSAGES")
-
-KNOWN_QUERY_FIELDS = {
-    "sort_order",
-    "msgType",
-    "addresses",
-    "refs",
-    "contentHashes",
-    "contentKeys",
-    "contentTypes",
-    "chains",
-    "channels",
-    "tags",
-    "hashes",
-    "history",
-    "pagination",
-    "page",  # page is handled in Pagination.get_pagination_params
-    "startDate",
-    "endDate",
-}
+LOGGER = logging.getLogger(__name__)
 
 
-async def get_filters(request: web.Request):
-    def get_query_list_field(field: str, separator=",") -> Optional[List[str]]:
-        field_str = request.query.get(field, None)
-        return field_str.split(separator) if field_str is not None else None
+DEFAULT_MESSAGES_PER_PAGE = 20
+DEFAULT_PAGE = 1
+DEFAULT_WS_HISTORY = 10
 
-    unknown_query_fields: Set[str] = set(request.query.keys()).difference(
-        KNOWN_QUERY_FIELDS
+
+class SortOrder(IntEnum):
+    ASCENDING = 1
+    DESCENDING = -1
+
+
+class MessageQueryParams(BaseModel):
+    sort_order: SortOrder = Field(
+        default=SortOrder.DESCENDING,
+        description="Order in which messages should be listed: "
+        "-1 means most recent messages first, 1 means older messages first.",
     )
-    if unknown_query_fields:
-        raise ValueError(f"Unknown query fields: {unknown_query_fields}")
+    message_type: Optional[MessageType] = Field(
+        default=None, alias="msgType", description="Message type."
+    )
+    addresses: Optional[List[str]] = Field(
+        default=None, description="Accepted values for the 'sender' field."
+    )
+    refs: Optional[List[str]] = Field(
+        default=None, description="Accepted values for the 'content.ref' field."
+    )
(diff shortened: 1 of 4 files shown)


src/aleph/web/controllers/utils.py
```diff
@@ -19,7 +19,9 @@ def get_pagination_params(request):
         with_pagination = pagination_param != 0
 
         if pagination_page < 1:
-            raise web.HTTPBadRequest(text=f"Query field 'page' must be ≥ 1, not {pagination_page}")
+            raise web.HTTPBadRequest(
+                text=f"Query field 'page' must be ≥ 1, not {pagination_page}"
```
hoh, line 23: Should we add a comment that this counter starts at 1 not 0 ?

src/aleph/web/controllers/utils.py
```diff
@@ -66,6 +68,21 @@ def iter_pages(self, left_edge=2, left_current=2, right_current=5, right_edge=2)
                 last = num
 
 
+def make_date_filters(start: float, end: float, filter_key: str):
```
hoh, line 71: Would it make sense to create a `NewType` named `TimeStamp` to specify that these floats represent timestamps ?

```suggestion
Timestamp = NewType('Timestamp', float)

def make_date_filters(start: Timestamp, end: Timestamp, filter_key: str):
```

src/aleph/web/controllers/utils.py
```diff
@@ -66,6 +68,21 @@ def iter_pages(self, left_edge=2, left_current=2, right_current=5, right_edge=2)
...
+    filters = []
+    if start:
+        filters.append({filter_key: {"$gte": start}})
+    if end:
+        filters.append({filter_key: {"$lt": end}})
+
+    if len(filters) > 1:
+        return {"$and": filters}
+    if filters:
```
hoh, line 80:
```suggestion
    if filters:
        assert len(filters) == 1
```

tests/api/test_messages.py
```diff
@@ -1,11 +1,19 @@
...
+import aiohttp
 import pytest
 
 from .utils import get_messages_by_keys
 
 MESSAGES_URI = "/api/v0/messages.json"
+MESSAGES_PAGE_URI = "/api/v0/messages/page/{page}.json"
+
+
+def get_messages_by_predicate(
```
hoh, line 13: This function seems to add more overhead than it solves 🤔. 
I also feel like I commenting on this exact code before 🤔🤔.

What about using https://docs.python.org/3/library/functions.html#filter instead ? It seems appears to do the exact same thing.

hoh (commented):
The approach of using `query_params = MessageQueryParams.parse_obj(request.query)` makes validation much more clear. 

I added a few minor comments, but looks good overall.

src/aleph/web/controllers/messages.py
```diff
@@ -1,129 +1,200 @@
...
+        description="Accepted values for the 'content.keys' field.",
+    )
+    content_types: Optional[List[ItemHash]] = Field(
+        default=None,
+        alias="contentTypes",
+        description="Accepted values for the 'content.type' field.",
+    )
+    chains: Optional[List[Chain]] = Field(
+        default=None, description="Accepted values for the 'chain' field."
+    )
```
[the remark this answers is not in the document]
odesenfans: What do you mean exactly?

src/aleph/web/controllers/messages.py
```diff
@@ -1,129 +1,200 @@
...
+        if self.content_types is not None:
+            filters.append({"content.type": {"$in": self.content_types}})
+        if self.refs is not None:
+            filters.append({"content.ref": {"$in": self.refs}})
+        if self.tags is not None:
+            filters.append({"content.content.tags": {"$elemMatch": {"$in": self.tags}}})
+        if self.chains is not None:
+            filters.append({"chain": {"$in": self.chains}})
+        if self.channels is not None:
+            filters.append({"channel": {"$in": self.channels}})
```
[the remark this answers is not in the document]
odesenfans: Any idea how? I've always had to resort to writing lists of filters like this, so if you have a better approach in mind I'm curious.

src/aleph/web/controllers/utils.py
odesenfans: There's a comment in the new Pydantic model. The end goal here is also to replace this class by a Pydantic model (ex: `PaginatedQuery`) that can be subclassed for any endpoint that requires pagination.

src/aleph/web/controllers/utils.py
odesenfans: I have it in a corner of my head, but that's for after the big refactoring.

src/aleph/web/controllers/utils.py
odesenfans: Don't need to, the `if` above catches lists with multiple elements.

tests/api/test_messages.py
odesenfans: Yeah you're not wrong. I'll remove it.

src/aleph/web/controllers/messages.py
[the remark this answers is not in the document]
hoh: Create this "code" automatically from the aleph_message models.

src/aleph/web/controllers/messages.py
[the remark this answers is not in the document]
hoh: Magic, something like:
```python3
for attribute in ProgramMessage:
    if hasattr(self, attribute + "s"):
        ...
```

src/aleph/web/controllers/utils.py
hoh: OK

hoh (approved):

PR #337, opened 2022-10-06
Fix: refuse time and pagination parameters for message WS

odesenfans:
Problem: the message websocket could receive "startDate"/"endDate"
and pagination parameters.

Solution: use a slightly different Pydantic model for the websocket
to reject these parameters.


odesenfans:
Depends on #335.

PR #338, opened 2022-10-06
Fix: general 500 error on posts endpoint

odesenfans:
Problem: the "/api/v0/posts.json" endpoint returns a 500 error because of an invalid parameter being used in a function (an attempt to remove the MongoDB object ID from the endpoint).

Solution: remove the object ID directly in the aggregate query.

Added a simple test to verify that the endpoint is at least working a little bit.

### DIFF
--- a/src/aleph/web/controllers/posts.py
+++ b/src/aleph/web/controllers/posts.py
@@ -85,10 +85,7 @@ async def view_posts_list(request):
     context = {"posts": posts}
 
     if pagination_per_page is not None:
-        total_msgs = await Message.collection.count_documents(
-            filter=find_filters,
-            projection={"_id": 0},
-        )
+        total_msgs = await Message.collection.count_documents(filter=find_filters)
 
         pagination = Pagination(
             pagination_page,
--- a/tests/api/test_posts.py
+++ b/tests/api/test_posts.py
@@ -1,0 +1,59 @@
+from typing import Dict, Iterable
+
+import aiohttp
+import pytest
+from aleph_message.models import MessageType
+
+from .utils import get_messages_by_keys
+
+POSTS_URI = "/api/v0/posts.json"
+
+
+def assert_posts_equal(posts: Iterable[Dict], expected_messages: Iterable[Dict]):
+    posts_by_hash = {post["item_hash"]: post for post in posts}
+
+    for expected_message in expected_messages:
+        post = posts_by_hash[expected_message["item_hash"]]
+        assert "_id" not in post
+
+        assert post["chain"] == expected_message["chain"]
+        assert post["channel"] == expected_message["channel"]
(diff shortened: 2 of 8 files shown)


tests/api/utils/__init__.py
```diff
@@ -0,0 +1,13 @@
+from typing import Dict, Iterable, List, Callable
+
+
+def get_messages_by_predicate(
+    messages: Iterable[Dict], predicate: Callable[[Dict], bool]
+) -> List[Dict]:
+    return [msg for msg in messages if predicate(msg)]
```
hoh, line 7: Is this function really useful ?
It introduces the uncommon term `predicate` and only contains a trivial one line of code.

tests/api/utils/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def get_messages_by_keys(messages: Iterable[Dict], **keys) -> List[Dict]:
```
hoh, line 10: What is this function doing ? Can you add a docstring to explain why we need it ?

src/aleph/web/controllers/posts.py
```diff
@@ -85,10 +85,7 @@ async def view_posts_list(request):
     context = {"posts": posts}
 
     if pagination_per_page is not None:
-        total_msgs = await Message.collection.count_documents(
-            filter=find_filters,
-            projection={"_id": 0},
```
hoh, line 90: Is the `_id` field still removed from the API ? If so, where is it filtered out ?

tests/api/test_posts.py
```diff
@@ -0,0 +1,59 @@
...
+        assert post["content"] == expected_message["content"]["content"]
+
+
+async def get_posts(api_client, **params) -> aiohttp.ClientResponse:
+    return await api_client.get(POSTS_URI, params=params)
+
+
+async def get_posts_expect_success(api_client, **params):
+    response = await get_posts(api_client, **params)
+    assert response.status == 200, await response.text()
```
hoh, line 44: Why not use something like [response.raise_for_status()](https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientResponse.raise_for_status) instead ?

tests/api/utils/__init__.py
odesenfans: I use it a ton in another PR, just thought I'd introduce it here so that it's already in. I can rename 'predicate' to 'cond', that works as well.

src/aleph/web/controllers/posts.py
odesenfans: In the aggregate query directly.

tests/api/test_posts.py
odesenfans: Because 'raise_for_status()' does not print the response body, which I want to happen in case of error.

PR #339, opened 2022-10-06
Fix: Validation errors did not return valid JSON on POST

hoh:
When posting an invalid message, PyDantic exposes the reason of the validation failure as JSON. This result was serialized in a string with the wrong headers (text/plain) instead of being returned as a proper JSON body.


odesenfans:
Integrated with work on Postgres.

PR #341, opened 2022-10-06
Internal: improve validation of the aggregates endpoint

odesenfans:
Use a Pydantic model instead of manual validation.


odesenfans:
Depends on #340.

PR #354, opened 2022-10-28
Internal: reduce Docker image size

odesenfans:


odesenfans:
Integrated with work on Postgres.

PR #355, opened 2022-11-18
Feature: Tezos message sync

odesenfans:
Problem: Users need a solution to pin content on Aleph automatically when minting NFTs on Tezos.

Solution: Add a sync worker to fetch messages from the Aleph sync smart contract on Tezos and transform them into valid Aleph messages.


hoh (approved):
Looks good to me. Testing...

PR #359, opened 2022-12-13
Add a limit parameter on the stats endpoint

BjrInt:


src/aleph/web/controllers/stats.py
```diff
@@ -81,6 +82,13 @@ async def addresses_stats_view(request):
 
     addresses = request.query.getall("addresses[]", [])
     check_time = None
+    limit = request.query.get("limit", None)
+
+    if limit is not None:
+        try:
+            limit = int(limit)
+        except ValueError:
+            limit = None
```
BjrInt, line 91: Seems very verbose, is there a better way to do this?

src/aleph/web/controllers/stats.py
```diff
@@ -81,6 +82,13 @@ async def addresses_stats_view(request):
 
     addresses = request.query.getall("addresses[]", [])
     check_time = None
+    limit = request.query.get("limit", 0)
+
+    if limit is not 0:
+        try:
+            limit = int(limit)
+        except ValueError:
+            limit = 0
```
BjrInt, line 91: Seems very verbose. Is there a better way to set a default value after type casting?

hoh: Have a look at this approach: https://github.com/aleph-im/pyaleph/blob/deaa4c3f0d7d9463078def59f2bfb69ac7023f3c/src/aleph/web/controllers/programs.py#L12-L16

PR #370, opened 2023-01-24
Feature: PostgreSQL

odesenfans:
Includes the following features:

* Replace MongoDB with PostgreSQL
* New message processing pipeline
* New endpoint POST /api/v0/messages/
* Python 3.11
* orjson as JSON parser
* Tezos smart contract support
* Switch to OOP for most services


BjrInt:
Here are some of the differences with the MongoDB API
- The channels endpoint (`api/v0/channels/list.json`) now returns a `null` entry
- There is no way to determine if a stored content on IPFS is a directory or a file. This information was previously available inside the store messages (but could be part of the new `files` endpoint)

odesenfans:
Merged manually to dev.

PR #374, opened 2023-02-28
Fix: Default peers of p2p-service did not work

hoh:
The p2p service failed to connect to the rest of the network using the configured peers. The default configuration of pyaleph is in Python code and cannot be read from the p2p-service.

Solution: override the default peers in the default configuration file as well.

Related: Fix the default values in the configuration of p2p-service in https://github.com/aleph-im/p2p-service/pull/42


odesenfans (approved):
Didn't think about this issue. LGTM.

PR #375, opened 2023-03-01
Feature: performance monitoring with Sentry

odesenfans:
Problem: we would like to measure the performance of API responses.

Solution: use Sentry performance monitoring.


odesenfans:
Integrated with the Postgres work.

PR #388, opened 2023-04-06
Fix: Dockerfile missed VOLUME field

hoh:
See https://docs.docker.com/engine/reference/builder/#volume

> The VOLUME instruction creates a mount point with the specified name and marks it as holding externally mounted volumes from native host or other containers


odesenfans:
As discussed, this field seems pretty useless. It does not reuse the same volume across runs (which is normal), and it is not useful when running with Docker Compose.

PR #395, opened 2023-04-14
hoh simplify config

hoh:


odesenfans:
Cool! I had this in the back of my mind for a while. Just fix the unit tests config, tests run outside of Docker so you'll need to use 127.0.0.1 as host address for Postgres and Redis.

PR #396, opened 2023-04-14
hoh update kubo

hoh:


odesenfans:
I had a look at the changelog since 0.15, in theory there's been no breaking change but I see they added a new P2P resource control mechanism that might have side effects. Gotta test this on a node.

odesenfans:
Superseded by #515.

PR #409, opened 2023-04-25
Chore: update to web3 6.2.0

odesenfans:


odesenfans:
Apparently mypy crashes on a recent change from web3 6.0.0b9 to 6.0.0. Upgrading to mypy 1.2.0 fixes the issue, but also reveals some other warnings that must be fixed before. TL;DR: upgrade mypy to 1.2.0 -> fix new mypy warnings -> merge this PR.

PR #421, opened 2023-04-28
Fix: support large number of open websockets + delete resources for unused websockets

odesenfans:


odesenfans:
Split in two: #422 and #423.

PR #423, opened 2023-04-28
Fix: delete RabbitMQ queues for unused websockets

odesenfans:
Problem: RabbitMQ queues can stay open indefinitely if the websocket filters only match few or no message. This can lead to increased resource consumption.

Solution: add a large timeout (5 minutes) and check periodically if the websocket connection is still open. If the user is still listening, keep the queue open. Otherwise, delete the queue and terminate the request.


odesenfans:
This is not enough yet, the timeout does work but `ws.closed` does not detect if the websocket was closed. We need to read from the websocket to determine if the connection is still open.

PR #444, opened 2023-05-30
Allow multiple message types

MHHukiewitz:
Problem: `msgType` is the only parameter when fetching messages, which does not take a list of items as a parameter.

Solution: add `msgTypes` as a parameter to the API and deprecate `msgType`.


src/aleph/db/accessors/messages.py
```diff
@@ -87,7 +89,15 @@ def make_matching_messages_query(
         select_stmt = select_stmt.where(MessageDb.sender.in_(addresses))
     if chains:
         select_stmt = select_stmt.where(MessageDb.chain.in_(chains))
+    if message_types:
+        if len(message_types) == 1:
+            select_stmt = select_stmt.where(MessageDb.type == message_types[0])
+        else:
+            select_stmt = select_stmt.where(MessageDb.type.in_(message_types))
```
odesenfans, line 96: Just use `in_`, we didn't measure if there was a performance difference so let's keep it simple. I'd be surprised if there's one.

src/aleph/db/accessors/messages.py
```diff
@@ -87,7 +89,15 @@ def make_matching_messages_query(
...
     if message_type:
+        warnings.warn(
+            "Warning: `msgType`/`message_type` query parameter is deprecated in favor of `msgTypes`\`message_types` and will be removed in future versions."
+        )
```
odesenfans, line 100: This warning is useless as it will only appear in our logs. We need OpenAPI docs, some day.

src/aleph/web/controllers/messages.py
```diff
@@ -64,7 +65,10 @@ class BaseMessageQueryParams(BaseModel):
         "-1 means most recent messages first, 1 means older messages first.",
     )
     message_type: Optional[MessageType] = Field(
-        default=None, alias="msgType", description="Message type."
+        default=None, alias="msgType", description="[DEPRECATED] Message type."
+    )
+    message_types: Optional[List[MessageType]] = Field(
+        default=None, alias="msgTypes", description="Accepted types of messages."
```
odesenfans, line 71: "Accepted message types", for consistency?

src/aleph/web/controllers/messages.py
```diff
@@ -64,7 +65,10 @@ class BaseMessageQueryParams(BaseModel):
         "-1 means most recent messages first, 1 means older messages first.",
     )
     message_type: Optional[MessageType] = Field(
-        default=None, alias="msgType", description="Message type."
+        default=None, alias="msgType", description="[DEPRECATED] Message type."
```
odesenfans, line 68: Maybe more in the form of : "Message type. Deprecated: use msgTypes instead".

odesenfans (changes requested):
A few minor things to change.

odesenfans (approved):

MHHukiewitz:
Rebased and ready @odesenfans

PR #447, opened 2023-06-01
Fix: no DB calls in message websocket

odesenfans:
Problem: the websocket uses the DB to apply filters to processed messages to determine whether they must be sent on the websocket. This leads to an exhaustion of DB connections when many websockets are open with the same filters.

Solution: send the full message on the RabbitMQ topic and apply filters in Python instead of using a DB query.

### DIFF
--- a/src/aleph/schemas/api/messages.py
+++ b/src/aleph/schemas/api/messages.py
@@ -27,6 +27,7 @@ from pydantic import BaseModel, Field
 from pydantic.generics import GenericModel
 
 import aleph.toolkit.json as aleph_json
+from aleph.db.models import MessageDb
 from aleph.types.message_status import MessageStatus, ErrorCode
 
 MType = TypeVar("MType", bound=MessageType)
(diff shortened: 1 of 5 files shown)


src/aleph/schemas/api/messages.py
```diff
@@ -97,19 +108,30 @@ class StoreMessage(
...
-    InstanceMessage,
-    PostMessage,
-    ProgramMessage,
-    StoreMessage,
-]
+def format_message(message: MessageDb) -> AlephMessage:
+    message_type = message.type
+
+    message_cls = MESSAGE_CLS_DICT[message_type]
+    return message_cls.from_orm(message)  # type: ignore[return-value]
```
hoh, line 128: What is the issue here ?

src/aleph/web/controllers/messages.py
```diff
@@ -237,50 +240,90 @@ async def _send_history_to_ws(
...
-        # TODO: put the filter key/values in the RabbitMQ message?
-        with session_factory() as session:
-            matching_messages = get_matching_messages(
-                session=session,
-                hashes=[item_hash],
-                include_confirmations=True,
-                **message_filters,
-            )
+        payload_bytes = mq_message.body
+        payload_dict = aleph_json.loads(payload_bytes)
```
hoh, line 321: Not related to this PR, but `aleph_json` looks like the project uses its own kind of JSON format.

src/aleph/schemas/api/messages.py
odesenfans: mypy does not infer the type of `message_cls` correctly. Says it's an instance of `BaseMessage`, which is correct, but does not match the `AlephMessage` union type. Basically unhappy because it can't figure out that `AlephMessage` is a union of all the possible child classes of `BaseMessage`.

src/aleph/web/controllers/messages.py
odesenfans: I know, but I wanted to make calls to the JSON module wrapper explicit. If you have a better idea, don't hesitate.

PR #462, opened 2023-08-03
Feature: Control of balance for instances

1yam:
Added functionality to check the current balance and compute costs for instances.


deployment/migrations/versions/0017_f9fa39b6bdef_vm_instances.py
```diff
@@ -129,7 +129,7 @@ def recreate_cost_views():
                                            free_disk.included_disk_space,
                                            0) AS additional_disk_space) additional_disk,
                  LATERAL ( SELECT CASE
-                                      WHEN vms.persistent THEN 2000
+                                      WHEN COALESCE(vms.persistent, true) THEN 2000
                                       ELSE 200
                                       END AS base_compute_unit_price) bcp,
                  LATERAL ( SELECT 1 + vms.environment_internet::integer AS compute_unit_price_multiplier) m,
```
odesenfans, line 135: The changes to this migration script will have to be put in a separate migration script that recreates the VM cost view (views must be recreated entirely, you cannot just add a column). Create a new empty migration script with `alembic revision -m "fix VM cost view"` and then take the pieces that you need from this migration script to drop the vm cost view and recreate it.

src/aleph/handlers/content/content_handler.py
```diff
@@ -47,6 +47,19 @@ async def process(self, session: DbSession, messages: List[MessageDb]) -> None:
         """
         pass
 
+    async def check_balance(self, session: DbSession, message: MessageDb) -> None:
+        """
+        Check user's balance and cost.
+
+        This function should verify whether the user has sufficient balance to perform
+        the requested action.
+        It's raise InsufficientBalanceException
```
odesenfans, line 56:
```suggestion
        Checks whether the user has enough Aleph tokens to process the message.

        Raises InsufficientBalanceException if the balance of the user is too low.
```

src/aleph/handlers/content/vm.py
```diff
@@ -298,6 +354,24 @@ class VmMessageHandler(ContentHandler):
 
     """
 
+    async def check_balance(self, session: DbSession, message: MessageDb):
```
odesenfans, line 357: Missing return value type (-> None).

src/aleph/handlers/content/content_handler.py
odesenfans: Also, I thought about this, but it might actually be better to define the function as `compute_cost(session: DbSession, message: MessageDb) -> Decimal` and then mutualize all the logic to check the current costs in the `MessageHandler` class. What do you think?

src/aleph/handlers/content/vm.py
```diff
@@ -289,6 +301,50 @@ class HasParent(Protocol):
             )
 
 
+def get_extra_storage(content: InstanceContent, session: DbSession) -> int:
+    volumes: int = 0
+    for i in content.volumes:
```
odesenfans, line 306: Rename `i` to `volume`.

src/aleph/handlers/content/vm.py
```diff
@@ -289,6 +301,50 @@ class HasParent(Protocol):
             )
 
 
+def get_extra_storage(content: InstanceContent, session: DbSession) -> int:
+    volumes: int = 0
```
odesenfans, line 305: Rename to `total_volume_size`. It would be better if this function returned the size in bytes, and then let the caller do whatever it wants (convert to MB, MiB, Watts, I don't care).

src/aleph/handlers/content/vm.py
```diff
@@ -289,6 +301,50 @@ class HasParent(Protocol):
             )
 
 
+def get_extra_storage(content: InstanceContent, session: DbSession) -> int:
```
odesenfans, line 304: Rename to `get_volume_size`? Irrelevant here, but this function could also be called on `ProgramContent` objects. Reflect that in the type hints.

src/aleph/handlers/content/vm.py
```diff
@@ -289,6 +301,50 @@ class HasParent(Protocol):
...
+                    session.query(StoredFileDb)
+                    .filter(StoredFileDb.hash == file_pin.file_hash)
+                    .first()
+                )
+                if file_record:
+                    volumes += file_record.size / (1024 * 1024)
+            else:
+                volumes += i.size_mib
+        except:
+            pass
```
odesenfans, line 326: This should be encapsulated in a function. Several remarks:

1. Instead of using `hasattr(i, ref)`, use `isinstance(volume, ImmutableVolume`).
2. I checked the code in the repo, you can actually just look for a matching file tag by using the `get_file_tag` function. You can then directly access the file data using the `FileTag.file` property/relationship.
3. Never use `.first()` when attempting to get one result only. SQLAlchemy has `one()` and `one_or_none()`.
4. The `query()` API of SQLAlchemy is deprecated, use the `select` approach instead.
5. We store all queries in the `accessors` module, here you won't need this query but now you know for next time.
6. I don't understand why you `try .. except` this. In general it's bad practice to `except` all exceptions, we usually narrow it down. Here I guess it's because you used `first()` and tried to catch the case where there's no result. Using `one_or_none()` resolves this issue. Catching the exception is fine as well, but catch a specific exception type, not just "everything".

src/aleph/handlers/content/vm.py
```diff
@@ -289,6 +301,50 @@ class HasParent(Protocol):
...
+    volumes += content.rootfs.size_mib
+    return volumes
+
+
+def get_additional_storage_price(content, session: DbSession) -> float:
```
odesenfans, line 331: Always use decimals when manipulating prices.

src/aleph/handlers/message_handler.py
```diff
@@ -316,6 +315,7 @@ async def process(
             session=session, pending_message=pending_message
         )
         content_handler = self.get_content_handler(message.type)
+        await content_handler.check_balance(session=session, message=message)
```
odesenfans, line 318: This should definitely be called after `check_dependencies` (otherwise we might look for volumes that don't exist) and probably after `check_permissions` (we should reduce the amount of computation done before checking permissions).

src/aleph/types/message_status.py
```diff
@@ -295,3 +297,22 @@ def __init__(self, target_hash: str):
 
     def details(self) -> Optional[Dict[str, Any]]:
         return {"errors": [{"message": self.target_hash}]}
+
+
+class InsufficientBalanceException(InvalidMessageException):
+    """
+    You don't have enough in balance
```
odesenfans, line 304:
```suggestion
    The user does not have enough tokens to perform the action.
```

tests/message_processing/test_process_instances.py
```diff
@@ -2,6 +2,7 @@
 import itertools
```
odesenfans, line 2: I 'd like to see more tests: 
* unit tests for the code that computes the cost of a single InstanceContent object.
* a test here that tries to process an instance without having enough tokens.

src/aleph/db/accessors/vms.py
```diff
@@ -113,3 +113,14 @@ def refresh_vm_version(session: DbSession, vm_hash: str) -> None:
     )
     session.execute(delete(VmVersionDb).where(VmVersionDb.vm_hash == vm_hash))
     session.execute(upsert_stmt)
+
+
+def get_total_cost_for_address(session: DbSession, address: str) -> int:
```
odesenfans, line 118: The return value should be a Decimal (costs can be fractional).

odesenfans (changes requested):
Good job, a few things to change here and there to keep you busy but the logic seems sound. Needs more unit tests though!

src/aleph/types/message_status.py
odesenfans: Same, this comment is not addressed?

tests/message_processing/test_process_instances.py
odesenfans: Where are the additional tests?

odesenfans (changes requested):
A lot of my comments from the first review still need to be addressed.

odesenfans (approved):

PR #463, opened 2023-08-21
Feature: authenticated file upload

1yam:
Problem: file upload is performed in two steps, one to push the file and one to push the associated STORE message.

Solution: improve `/api/v0/storage/add_file` to allow the user to send a STORE message along with his file, making upload an atomic operation.

### DIFF
--- a/src/aleph/api_entrypoint.py
+++ b/src/aleph/api_entrypoint.py
@@ -74,7 +74,7 @@ async def configure_aiohttp_app(
         app[APP_STATE_NODE_CACHE] = node_cache
         app[APP_STATE_STORAGE_SERVICE] = storage_service
         app[APP_STATE_SESSION_FACTORY] = session_factory
-        # app[APP_STATE_CHAIN_SERVICE] = chain_service
+        app[APP_STATE_CHAIN_SERVICE] = chain_service
 
     return app
 
--- a/src/aleph/api_entrypoint.py
+++ b/src/aleph/api_entrypoint.py
@@ -6,6 +6,7 @@ from aiohttp import web
 from configmanager import Config
 
 import aleph.config
+from aleph.chains.chain_service import ChainService
 from aleph.db.connection import make_engine, make_session_factory
 from aleph.services.cache.node_cache import NodeCache
 from aleph.services.ipfs import IpfsService
@@ -21,7 +22,10 @@ from aleph.web.controllers.app_state_getters import (
     APP_STATE_NODE_CACHE,
     APP_STATE_P2P_CLIENT,
     APP_STATE_SESSION_FACTORY,
-    APP_STATE_STORAGE_SERVICE, APP_STATE_MQ_CHANNEL, APP_STATE_MQ_WS_CHANNEL,
+    APP_STATE_STORAGE_SERVICE,
+    APP_STATE_MQ_CHANNEL,
+    APP_STATE_MQ_WS_CHANNEL,
+    APP_STATE_CHAIN_SERVICE,
 )
 
 
@@ -49,6 +53,9 @@ async def configure_aiohttp_app(
             ipfs_service=ipfs_service,
             node_cache=node_cache,
         )
+        chain_service = ChainService(
+            storage_service=storage_service, session_factory=session_factory
+        )
 
         app = create_aiohttp_app()
 
@@ -67,6 +74,7 @@ async def configure_aiohttp_app(
         app[APP_STATE_NODE_CACHE] = node_cache
         app[APP_STATE_STORAGE_SERVICE] = storage_service
         app[APP_STATE_SESSION_FACTORY] = session_factory
+        # app[APP_STATE_CHAIN_SERVICE] = chain_service
 
     return app
 
--- a/src/aleph/web/controllers/storage.py
+++ b/src/aleph/web/controllers/storage.py
@@ -170,7 +170,7 @@ async def storage_add_file_with_message(
 
     session.add(pending_message_db)
     session.commit()
-    if storage_metadata.sync and mq_queue:
+    if mq_queue:
         mq_message = await mq_read_one_message(mq_queue, 30)
 
         if mq_message is None:
--- a/src/aleph/web/controllers/storage.py
+++ b/src/aleph/web/controllers/storage.py
@@ -112,7 +112,7 @@ async def _verify_user_balance(
         session=session, address=pending_message_db.sender
     )
     if current_balance < (Decimal(required_balance) + current_cost_for_user):
-        raise web.HTTPPaymentRequired
+        raise web.HTTPPaymentRequired()
 
 
 async def _verify_user_file(message: PendingStoreMessage, size: int, file_io) -> None:
--- a/src/aleph/web/controllers/storage.py
+++ b/src/aleph/web/controllers/storage.py
@@ -1,49 +1,36 @@
-import asyncio
 import base64
 import datetime as dt
 import functools
 import logging
 from hashlib import sha256
-from io import StringIO
-from typing import Union, Tuple, Dict, Optional
+from typing import Union, Tuple
 
 import aio_pika
 from eth_account import Account
 from eth_account.messages import encode_defunct
 
 from aleph.chains.common import get_verification_buffer
-from aleph.jobs.process_pending_messages import PendingMessageProcessor
 
 from aiohttp import web
 from aiohttp.web_request import FileField
 from aleph_message.models import ItemType
 from multidict import MultiDictProxy
-
-from aleph.chains.chain_service import ChainService, LOGGER
-from aleph.chains.nuls import NulsConnector
 from aleph.db.accessors.balances import get_total_balance
 from aleph.db.accessors.files import count_file_pins, get_file
-from aleph.db.accessors.messages import get_message_status, message_exists
-from aleph.db.connection import make_session_factory
 from aleph.db.models import PendingMessageDb
 from aleph.exceptions import AlephStorageException, UnknownHashError
-from aleph.services.p2p import init_p2p_client
-from aleph.services.storage.engine import StorageEngine
-from aleph.services.storage.fileystem_engine import FileSystemStorageEngine
-from aleph.storage import StorageService
 from aleph.toolkit import json
-from aleph.toolkit.timestamp import timestamp_to_datetime
-from aleph.types.db_session import DbSessionFactory, DbSession
+from aleph.types.db_session import DbSession
 from aleph.utils import run_in_executor, item_type_from_hash
 from aleph.web.controllers.app_state_getters import (
     get_session_factory_from_request,
-    get_storage_service_from_request, get_mq_channel_from_request, get_config_from_request, get_mq_conn_from_request,
+    get_storage_service_from_request,
+    get_config_from_request,
 )
 from aleph.web.controllers.utils import multidict_proxy_to_io
 from aleph.schemas.pending_messages import BasePendingMessage
 
 logger = logging.getLogger(__name__)
-from aleph.schemas.pending_messages import parse_message
 
 MAX_FILE_SIZE = 100 * 1024 * 1024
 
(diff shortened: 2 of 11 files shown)


src/aleph/web/controllers/storage.py
```diff
@@ -56,20 +80,133 @@ async def add_storage_json_controller(request: web.Request):
...
+    except Exception as e:
+        verified = False
+    return verified
+
+
+async def get_message_content(
+    post_data: MultiDictProxy[Union[str, bytes, FileField]]
+) -> Tuple[dict, int]:
+    message_bytearray = post_data.get("message", b"")
+    value = post_data.get("size") or 0
```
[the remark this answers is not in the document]
odesenfans: Also, we probably need a maximum size somewhere. Put 1 GiB at the moment and return a 413 (Payload too large) if the file is larger.

src/aleph/web/controllers/storage.py
```diff
@@ -56,20 +80,133 @@ async def add_storage_json_controller(request: web.Request):
...
+        message_dict["time"] = float(message_dict["time"])
+    else:
+        message_dict = {}
+
+    return message_dict, int(str(value))
+
+
+async def verify_and_handle_request(
+    pending_message_db, file_io, message, size, session_factory
+):
```
odesenfans, line 128: You should split this function in (at least) 3 functions.

1. verify that the message has a valid signature
2. verify that the user has enough tokens to pay for the feature
3. read the file and verify that it matches the item hash.

This will make your code clearer and separate concerns.

src/aleph/web/controllers/storage.py
```diff
@@ -56,20 +80,133 @@ async def add_storage_json_controller(request: web.Request):
...
-    # No need to pin it here anymore.
-    # TODO: find a way to specify linked ipfs hashes in posts/aggr.
     post = await request.post()
     file_io = multidict_proxy_to_io(post)
+    message, size = await get_message_content(post)
+    pending_message_db = PendingMessageDb.from_message_dict(
+        message_dict=message, reception_time=dt.datetime.now(), fetched=True
+    )
+    mq_channel = await get_mq_channel_from_request(request, logger=logger)
+    mq_queue = await mq_make_aleph_message_topic_queue(
```
odesenfans, line 167: The MQ stuff only needs to be done if a) there is a message attached to the request and b) the user specified `sync=True` in the request. The `sync` parameter is nowhere to be found in your PR, add it.

src/aleph/web/controllers/storage.py
```diff
@@ -56,20 +80,133 @@ async def add_storage_json_controller(request: web.Request):
...
+        status_str, _item_hash = mq_message.routing_key.split(".")
+        processing_status = MessageProcessingStatus(status_str)
+        status_code = _processing_status_to_http_status(processing_status)
+        return web.json_response(status=status_code, text=file_hash)
+
+
+async def storage_add_file(request: web.Request):
+    post = await request.post()
+    if post.get("message", b"") is not None and post.get("size") is not None:
+        return await storage_add_file_with_message(request)
```
odesenfans, line 200: This split is not ideal because you duplicate the part where you read the file and add it in the DB. Also, there is a major bug in your new function because you never actually write the file to disk nor add the corresponding entry in the `files` table. I'd rather have one function where you call other functions, example (fleshed out, just to give an example):

```
if message and size:
  pending_message = await verify_message(message)
  await verify_balance(pending_message.content.address, size)
  expected_file_hash = pending_message.content.item_hash  
else:
  expected_file_hash = None

file_io = multidict_proxy_to_io(post)
file_content, item_hash = await read_and_check_file(file_io, expected_file_hash)

# Note that this requires overloading the `StorageService.add_file` method to support passing bytes instead of an IO object 
await storage_service.add_file(session=session, file=file_content, engine=ItemType.storage)
```

odesenfans (changes requested):
I still need to review the tests. There are a lot of things to change in the implementation as it is.

src/aleph/web/controllers/storage.py
```diff
@@ -56,20 +80,133 @@ async def add_storage_json_controller(request: web.Request):
...
+            ),
+        )
+        if address == message.sender:
+            verified = True
+        else:
+            return False
+
+    except Exception as e:
+        verified = False
+    return verified
```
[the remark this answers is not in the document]
1yam: At the moment it is half implemented (must be added to ccn_api_client to be able to use it)

src/aleph/web/controllers/storage.py
```diff
@@ -56,20 +80,133 @@ async def add_storage_json_controller(request: web.Request):
...
+    return message_dict, int(str(value))
+
+
+async def verify_and_handle_request(
+    pending_message_db, file_io, message, size, session_factory
+):
+    content = file_io.read(size)
+    item_content = json.loads(message["item_content"])
+    actual_item_hash = sha256(content).hexdigest()
+    c_item_hash = item_content["item_hash"]
```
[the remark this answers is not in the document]
1yam: c == client on my head.
It's the hash give by the "client"

src/aleph/api_entrypoint.py
```diff
@@ -67,6 +74,7 @@ async def configure_aiohttp_app(
         app[APP_STATE_NODE_CACHE] = node_cache
         app[APP_STATE_STORAGE_SERVICE] = storage_service
         app[APP_STATE_SESSION_FACTORY] = session_factory
+        # app[APP_STATE_CHAIN_SERVICE] = chain_service
```
odesenfans, line 77: Why is this line commented?

src/aleph/web/controllers/storage.py
```diff
@@ -56,20 +92,133 @@ async def add_storage_json_controller(request: web.Request):
...
+) -> None:
+    current_balance = get_total_balance(
+        session=session, address=pending_message_db.sender
+    ) or Decimal(0)
+    required_balance = (size / MiB) / 3
+    current_cost_for_user = get_total_cost_for_address(
+        session=session, address=pending_message_db.sender
+    )
+    if current_balance < (Decimal(required_balance) + current_cost_for_user):
+        raise web.HTTPPaymentRequired
```
odesenfans, line 115: Pretty sure you need to build the object.

```suggestion
        raise web.HTTPPaymentRequired()
```

src/aleph/web/controllers/storage.py
```diff
@@ -56,20 +92,133 @@ async def add_storage_json_controller(request: web.Request):
...
+
+
+async def _verify_user_file(message: PendingStoreMessage, size: int, file_io) -> None:
+    file_io.seek(0)
```
odesenfans, line 119: Why do you need to seek? Are you reading the file twice?

src/aleph/web/controllers/storage.py
```diff
@@ -56,20 +92,133 @@ async def add_storage_json_controller(request: web.Request):
...
+        mq_channel = await get_mq_channel_from_request(request, logger=logger)
+        mq_queue = await mq_make_aleph_message_topic_queue(
+            channel=mq_channel,
+            config=config,
+            routing_key=f"*.{pending_message_db.item_hash}",
+        )
+
+    session.add(pending_message_db)
+    session.commit()
+    if storage_metadata.sync and mq_queue:
```
odesenfans, line 174:
```suggestion
    if mq_queue:
```

is enough.

src/aleph/web/controllers/storage.py
```diff
@@ -56,20 +92,133 @@ async def add_storage_json_controller(request: web.Request):
...
+        mq_message = await mq_read_one_message(mq_queue, 30)
+
+        if mq_message is None:
+            raise web.HTTPAccepted()
+        if mq_message.routing_key is not None:
```
odesenfans, line 179: What happens if `routing_key` is None?

Note that since you subscribe using a routing key, this case should never happen.

src/aleph/web/controllers/storage.py
```diff
@@ -56,20 +92,133 @@ async def add_storage_json_controller(request: web.Request):
...
+            return web.json_response(status=status_code, text=_item_hash)
+
+
 async def storage_add_file(request: web.Request):
     storage_service = get_storage_service_from_request(request)
     session_factory = get_session_factory_from_request(request)
-
-    # No need to pin it here anymore.
-    # TODO: find a way to specify linked ipfs hashes in posts/aggr.
+    # TODO : Add chainservice to ccn_api_client to be able to call get_chainservice_from_request
```
odesenfans, line 189: Do it.

odesenfans (changes requested):

src/aleph/web/controllers/storage.py
1yam: [            file_content = fileobject.read()](https://github.com/1yam/pyaleph/blob/4db385efd6ee8e1d2677091cb8c13a36e283272d/src/aleph/storage.py#L274)

The functions add_file on Storage Service class wasn't doing seek when after read (he was doing it for ipf,  I replace the seek here and not on _verify_user_file

odesenfans: We really shouldn't read the file twice, that's an expensive operation.

src/aleph/web/controllers/storage.py
```diff
@@ -56,20 +92,129 @@ async def add_storage_json_controller(request: web.Request):
...
+    post = await request.post()
+    metadata = post.get("metadata", b"")
+    storage_metadata = None
+    try:
+        if isinstance(metadata, FileField):
+            metadata_content = metadata.file.read()
+            storage_metadata = StorageMetadata.parse_raw(metadata_content)
+        else:
+            storage_metadata = StorageMetadata.parse_raw(metadata)
+    except Exception as e:
```
odesenfans, line 201: Avoid wide `except Exception` clauses. What are you trying to intercept?

src/aleph/web/controllers/storage.py
```diff
@@ -56,20 +92,129 @@ async def add_storage_json_controller(request: web.Request):
...
     post = await request.post()
     file_io = multidict_proxy_to_io(post)
+    post = await request.post()
+    metadata = post.get("metadata", b"")
+    storage_metadata = None
+    try:
+        if isinstance(metadata, FileField):
+            metadata_content = metadata.file.read()
+            storage_metadata = StorageMetadata.parse_raw(metadata_content)
+        else:
```
odesenfans, line 199: I'm pretty sure this `else` breaks the endpoint when the metadata is not specified. You default to `b""` so you will get a Pydantic exception in this branch.

src/aleph/web/controllers/storage.py
```diff
@@ -56,20 +92,129 @@ async def add_storage_json_controller(request: web.Request):
...
+            raise web.HTTPUnprocessableEntity()
+    if metadata is None:
+        if len(file_io.read()) > (25 * MiB):
+            raise web.HTTPUnauthorized()
+    file_io.seek(0)
+
     with session_factory() as session:
         file_hash = await storage_service.add_file(
             session=session, fileobject=file_io, engine=ItemType.storage
         )
```
odesenfans, line 212: Keeping this block for all cases means that you add the file directly in all cases before checking the signature, balance etc.

src/aleph/web/controllers/utils.py
```diff
@@ -160,3 +164,42 @@ async def mq_make_aleph_message_topic_queue(
...
+        return None
+    finally:
+        await mq_queue.cancel(consumer_tag)
+
+
+def validate_message_dict(message_dict: Mapping[str, Any]) -> BasePendingMessage:
+    try:
+        return parse_message(message_dict)
+    except InvalidMessageException as e:
+        raise web.HTTPUnprocessableEntity(body=str(e))
```
odesenfans, line 205: Nitpick: configure your IDE to add an empty line on save.

odesenfans (changes requested):
The logic of the endpoint is still not what I'd expect. You read the file multiple times, add it to the storage before checks, etc.

src/aleph/web/controllers/storage.py
```diff
@@ -56,20 +92,129 @@ async def add_storage_json_controller(request: web.Request):
...
+    if mq_queue:
+        mq_message = await mq_read_one_message(mq_queue, 30)
+
+        if mq_message is None or mq_message.routing_key is None:
+            raise web.HTTPAccepted()
+
+        status_str, _item_hash = mq_message.routing_key.split(".")
+        processing_status = MessageProcessingStatus(status_str)
+        status_code = processing_status_to_http_status(processing_status)
+        return web.json_response(status=status_code, text=_item_hash)
```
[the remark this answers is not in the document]
1yam: https://github.com/1yam/pyaleph/blob/e625014f76ced48de20fc3279b5e5ac23ce30634/src/aleph/web/controllers/storage.py#L218

i am

src/aleph/web/controllers/storage.py
```diff
@@ -56,20 +92,129 @@ async def add_storage_json_controller(request: web.Request):
...
+        raise web.HTTPForbidden()
+
+
+async def _verify_user_balance(
+    pending_message_db: PendingMessageDb, session: DbSession, size: int
+) -> None:
+    current_balance = get_total_balance(
+        session=session, address=pending_message_db.sender
+    ) or Decimal(0)
+    required_balance = (size / MiB) / 3
```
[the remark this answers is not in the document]
1yam: https://github.com/1yam/pyaleph/blob/eb91e2190ef4cd7c941a173b5a68705a0b42e31f/deployment/migrations/versions/0018_7bcb8e5fe186_fix_vm_cost_view.py#L96

1048576 = (1024 * 1024) so i thinks the view is correct ?

odesenfans (approved):

PR #471, opened 2023-09-01
Feature: upgrade balance endpoint

1yam:
Problem:
  Currently, the Balance API Endpoint only returns the total balance for an address. And user (also frontend) as no easy way to get number of locked token.

On : 
`/api/v0/addresses/{address}/balance`

Solution:
  Enhance the Balance API Endpoint to include the "locked_amount" field in the response, indicating the amount of tokens in use for a specific address.

Example:

```json
{
    "address": "0x9319Ad3B7A8E0eE24f2E639c40D8eD124C5520Ba",
    "balance": 22192,
    "locked_amount": 2726
}

```


odesenfans (changes requested):
Looks good, just add tests.

tests/conftest.py
```diff
@@ -126,3 +150,179 @@ async def ccn_api_client(
...
+        address="0x9319Ad3B7A8E0eE24f2E639c40D8eD124C5520Ba",
+        chain=Chain.ETH,
+        balance=Decimal(22_192),
+        eth_height=0,
+    )
+
+    with session_factory() as session:
+        session.add(balance)
+        session.commit()
+        insert_volume_refs(session, fixture_instance_message)
```
odesenfans, line 326: I don't understand why you insert the volumes in this fixture, as it has nothing to do with the balance. I expect 3 fixtures:

* `instance_message`
* `instance_message_with_volumes_in_db` (depends on the first one)
* `user_balance` (depends on nothing).

tests/api/test_balance.py
```diff
@@ -0,0 +1,13 @@
...
+    AlephBalanceDb,
+)
+MESSAGES_URI = "/api/v0/addresses/0x9319Ad3B7A8E0eE24f2E639c40D8eD124C5520Ba/balance"
+
+@pytest.mark.asyncio
+async def test_get_balance(ccn_api_client, user_balance: AlephBalanceDb):
+    response = await ccn_api_client.get(MESSAGES_URI)
+    assert response.status == 200, await response.text()
+    data = await response.json()
+    assert data["balance"] == 22192
```
odesenfans, line 12: Use the value from the fixture.
```suggestion
    assert data["balance"] == user_balance.balance
```

tests/api/test_balance.py
```diff
@@ -0,0 +1,13 @@
...
+    assert data["locked_amount"] == 6
```
odesenfans, line 13: How come this value is so low?

odesenfans (changes requested):

tests/api/test_balance.py
1yam: now it's implement too, just miss to add

odesenfans (approved):

PR #522, opened 2023-11-13
Solve balance_check issue for instances

nesitor:


src/aleph/services/cost.py
```diff
@@ -85,7 +85,7 @@ def _get_nb_compute_units(content: ExecutableContent) -> int:
 
 def _get_compute_unit_multiplier(content: ExecutableContent) -> int:
     compute_unit_multiplier = 1
-    if content.environment.internet:
+    if content.environment.internet and isinstance(content, ProgramContent):
```
hoh, line 88: The pricing of a persistent VM should be identical to the instances.  This makes them different.

hoh: We decided to update the pricing to be identical for both, and the pricing of both instances and on-demand VMs must include internet access without supplement.

nesitor: The price between persistent VMs and instances should be the same, but not for ephimeral on-demand VMs.

PR #532, opened 2023-12-11
Add Kubo IPFS daemon configuration

hoh:
- Fix: Conflict between multiple database migrations
- Work-in-progress: Add customization for Kubo and block ISP KPN


hoh:
This still needs documentation regarding installation and upgrade manuals.

Psycojoker (approved):
I'm not very familiar with Kubo but that looks like a good addition.

PR #535, opened 2023-12-19
Fix: Old method upload

1yam:
Problem:
We require the old upload method to remain functional. In some cases, it may crash due to the HTTP client not setting the multipart header for the file_field.

Solution:
If no message is sent, we don't need to know the file size. We can simply read chunks until we reach the maximum size for unauthenticated uploads or end of file.

### DIFF
--- a/src/aleph/web/controllers/storage.py
+++ b/src/aleph/web/controllers/storage.py
@@ -114,6 +114,7 @@ class MultipartUploadedFile:
     def __init__(self, file_field: FileField, metadata: bool):
         self.metadata = metadata
         self.file_field = file_field
+
         self.MAX_SIZE = (
             MAX_UPLOAD_FILE_SIZE
             if self.metadata
@@ -124,10 +125,10 @@ class MultipartUploadedFile:
             if self.metadata:
                 content_length_str = file_field.headers["Content-Length"]
                 self.size = int(content_length_str)
-                self._content = self.file_field.file.read()
+                self._content = self.read_file_with_max_size(self.MAX_SIZE)
             else:
                 self._content = self.read_file_with_max_size(
-                    MAX_UNAUTHENTICATED_UPLOAD_FILE_SIZE
+                    self.MAX_SIZE
                 )
                 if self._content:
                     self.size = len(self._content)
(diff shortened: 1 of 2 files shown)


src/aleph/web/controllers/storage.py
```diff
@@ -118,18 +108,53 @@ def content(self) -> Union[str, bytes]:
...
 
-    def __init__(self, file_field: FileField):
+    def __init__(self, file_field: FileField, metadata: bool):
+        self.metadata = metadata
         self.file_field = file_field
+        self.MAX_SIZE = (
+            MAX_UPLOAD_FILE_SIZE
+            if self.metadata
+            else MAX_UNAUTHENTICATED_UPLOAD_FILE_SIZE
+        )
```
MHHukiewitz, line 121: This attribute doesn't seem to be used

src/aleph/web/controllers/storage.py
```diff
@@ -118,18 +108,53 @@ def content(self) -> Union[str, bytes]:
...
 
         try:
-            content_length_str = file_field.headers["Content-Length"]
-            self.size = int(content_length_str)
+            if self.metadata:
+                content_length_str = file_field.headers["Content-Length"]
+                self.size = int(content_length_str)
+                self._content = self.file_field.file.read()
```
MHHukiewitz, line 127: This should use `self.read_file_with_max_size()` and possibly throw an exception, if we were not able to read everything

src/aleph/web/controllers/storage.py
```diff
@@ -118,18 +108,53 @@ def content(self) -> Union[str, bytes]:
...
+
+        while True:
+            chunk = self.file_field.file.read(buffer_size)
+
+            if not chunk:
+                break
+
+            total_read += len(chunk)
+            if total_read > max_size:
+                raise web.HTTPForbidden()
```
MHHukiewitz, line 153: Throw `413 Payload Too Large` instead. Throw `400 Bad Request` if Content-Length does not match actual body bytes.

MHHukiewitz (changes requested):
Needs a bit of work. Also, some tests checking the failure states (too big file, Content-Length mismatch) should be included.

MHHukiewitz (approved):
LGTM

MHHukiewitz (changes requested):
After doing some research, I discovered that `await request.post()` reads the WHOLE request body into memory, clearly defeating the purpose of much of the code here.

nesitor:
Closed in favour https://github.com/aleph-im/pyaleph/pull/559

PR #538, opened 2023-12-29
Exclude `stream` payment system from balance check

nesitor:
Problem: Checking the balance on Pay-as-you-go payment method don't allow the message to be processed.

Solution: Exclude stream payment method on balance check system.

### DIFF
--- a/src/aleph/handlers/content/vm.py
+++ b/src/aleph/handlers/content/vm.py
@@ -323,6 +323,9 @@ class VmMessageHandler(ContentHandler):
             if not content.on.persistent:
                 return
 
+        if content.payment.is_stream:
+            return
+
         required_tokens = compute_cost(session=session, content=content)
 
         current_balance = (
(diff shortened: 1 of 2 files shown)


src/aleph/handlers/content/vm.py
```diff
@@ -323,6 +323,9 @@ async def check_balance(self, session: DbSession, message: MessageDb) -> None:
             if not content.on.persistent:
                 return
 
+        if content.payment and content.payment.is_stream:
+            return
```
aliel, line 327: how do you ensure that a payment stream exists?

aliel (approved):

src/aleph/handlers/content/vm.py
nesitor: It is only checked on the CRNs and should check it also on the Scheduler, on the CCNs we just accept the message.

MHHukiewitz (approved):
LGTM

MHHukiewitz:
Same question though, should this really be already merged into `master`? @hoh

PR #528, opened 2023-12-01
Feature: New Metrics endpoint for CRN & CNN

1yam:
### _**Problem:**_ 

> The current approach for getting those information is to fetch the whole list of nodes using ranges of post messages, which means pulling a whole lot of needless information (35mb for a week of CRN metrics).

### _**Solutions:**_

> Implementing two API endpoints to query CRN & CNN metrics in the Pyaleph database.

1. `/api/v0/core/{node_id}/metrics`
```
{
    "metrics": {
        "item_hash": [<string>, <string>, ...],
        "measured_at": [<UNIX timestamp>, <UNIX timestamp>, ...],
        "base_latency": [<float>, <float>, ...],
        "base_latency_ipv4": [<float>, <float>, ...],
        "metrics_latency": [<float>, <float>, ...],
        "aggregate_latency": [<float>, <float>, ...],
        "file_download_latency": [<float>, <float>, ...],
        "pending_messages": [<int>, <int>, ...],
        "eth_height_remaining": [<int>, <int>, ...],
    }
}

```

3. `/api/v0/compute/{node_id}/metrics`
```
{
    metrics: {
        item_hash: [<string>, <string>, ...],
        measured_at: [<UNIX timestamp>, <UNIX timestamp>, ...],
        base_latency: [<float>, <float>, ... ],
        base_latency_ipv4: [<float>, <float>, ...],
        full_check_latency: [<float>, <float>, ...],
        diagnostic_vm_latency: [<float>, <float>, ...],
    }
}
``` 

### _**Params:**_

- `start_date`: float
- `end_date`: float
- `sort`: ASC or DESC

PR #547, opened 2024-01-24
Fix: All node metrics were returned by default

hoh:
This causes a lot of data to be returned in the case of old nodes and is usually unnecessary.

Solution: Default to limiting the query to the last 2 weeks of data, or to 2 weeks before the specified date of last data.

This also renames `start_date -> start_timestamp` and `end_date -> end_timestamp` to prevent confusion since these fields are expected to be timestamps and note datetime or ISO 8601.

Related to https://github.com/aleph-im/pyaleph/pull/528

### DIFF
--- a/src/aleph/db/accessors/metrics.py
+++ b/src/aleph/db/accessors/metrics.py
@@ -1,3 +1,4 @@
+import time
 from typing import Optional
 
 from sqlalchemy import select, text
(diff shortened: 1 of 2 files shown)


MHHukiewitz (approved):
LGTM

nesitor:
Please, can you ammend the last commits to just avoid to have 3 different commits with the same name?

src/aleph/web/controllers/main.py
```diff
@@ -82,8 +82,8 @@ async def metrics_json(request: web.Request) -> web.Response:
 
 
 class Metrics(BaseModel):
-    start_date: Optional[float] = None
-    end_date: Optional[float] = None
+    start_timestamp: Optional[float] = None
+    end_timestamp: Optional[float] = None
```
nesitor, line 86: Will this change break the frontend account and console pages?

hoh: This API is not used anywhere that I am aware of. We plan to upgrade the metrics graph on CRN homepages to use it in the future.

PR #555, opened 2024-02-19
Fix sortBy/sortOrder on Message/Post queries

philogicae:
# Found issues:
- Linked to [issue 95 on SDK](https://github.com/aleph-im/aleph-sdk-python/issues/95) -> Solved here: [PR](https://github.com/aleph-im/aleph-sdk-python/pull/104)
- No sorting when queries request it, even after changing SDK codebase: 

# Proposed solution:
- Added missing aliases on BaseMessageQueryParams & PostQueryParams

@hoh

### DIFF
--- a/tests/api/test_list_messages.py
+++ b/tests/api/test_list_messages.py
@@ -272,8 +272,8 @@ async def fetch_messages_filter_time(
 ) -> aiohttp.ClientResponse:
 
     params: Dict[str, Union[float, int, str]] = {
-        "sort_by": sort_by,
-        "sort_order": sort_order,
+        "sortBy": sort_by,
+        "sortOrder": sort_order,
     }
     if start:
         params["startDate"] = start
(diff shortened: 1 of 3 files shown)


MHHukiewitz (approved):
LGTM

tests/api/test_list_messages.py
```diff
@@ -373,7 +373,7 @@ async def fetch_messages_with_pagination(
 ):
     return await api_client.get(
         MESSAGES_URI,
-        params={"page": page, "pagination": pagination, "sort_order": sort_order},
+        params={"page": page, "pagination": pagination, "sortOrder": sort_order},
```
hoh, line 376: Would this break backward compatibility with current software using `sort_order` ? How can we ensure it does not ?

hoh (commented):
Thanks, looks good overall :+1:

tests/api/test_list_messages.py
philogicae: Yes, better double check before to merge this PR. The missing aliases were an irregularity, since camelCase is used for key names at this endpoint. I also checked the JS SDK, sorting is not part yet of param queries, so all good there.

philogicae: I just double checked, for PyAleph, Python SDK and JS SDK, it's ok. Then I don't know if/which other repos are using those endpoints.

MHHukiewitz (approved):

PR #473, opened 2023-09-11
Feature: modification and creation date in the aggregate messages

1yam:
Problem :
To avoid fetching all the messages to find the creation/modification time of a specific aggregate entry it will be useful to have this info directly in the aggregate message

Solution:
Update '/api/v0/aggregates' Endpoint

param : 
 - with_info: bool default : False 

From :

```
{
  "address": "0xa1B3bb7d2332383D96b7796B908fB7f7F3c2Be10",
  "data": {
    "key1": ...,
    "key2": ...,
  }
}
```

To:
```
{
  "address": "0xa1B3bb7d2332383D96b7796B908fB7f7F3c2Be10",
  "data": {
    "key1": ...,
    "key2": ...,
  }
  "info": {
    "key1": {
      "created": ...,
      "last_updated": ...,
      "original_item_hash": ...,
      "last_update_item_hash: ...,
    },
    "key2": ...
  }
}
```

Fixes #470.

PR #556, opened 2024-02-19
Fix aggregate filter by keys

aliel:
Fix regression introduced by (#473)


hoh (approved):
Thank you ! Simple fix and the test that goes with it, perfect !

PR #557, opened 2024-02-19
Add param to only return the value of an aggregate by its key

aliel:


src/aleph/web/controllers/aggregates.py
```diff
@@ -62,17 +60,27 @@ async def address_aggregate(request: web.Request) -> web.Response:
...
     if not aggregates:
         raise web.HTTPNotFound(text="No aggregate found for this address")
 
+    if query_params.only_value and query_params.keys and len(query_params.keys) == 1:
+        output = {}
+        target_key = query_params.keys[0]
+        for result in aggregates:
+            output[result[0]] = result[1]
+
+        return web.json_response(output[target_key])
```
hoh, line 77: Can you add a test to cover this case ?

src/aleph/web/controllers/aggregates.py
```diff
@@ -22,6 +19,7 @@ class AggregatesQueryParams(BaseModel):
     keys: Optional[List[str]] = None
     limit: int = DEFAULT_LIMIT
     with_info: bool = False
+    only_value: bool = False
```
hoh, line 22: Nitpick: would `value_only` make more sense than `only_value` ? Seems more like English literal :thinking:

src/aleph/web/controllers/aggregates.py
```diff
@@ -62,17 +60,27 @@ async def address_aggregate(request: web.Request) -> web.Response:
...
-                session=session, owner=address, with_info=query_params.with_info
+                session=session, owner=address,
+                with_info=query_params.with_info, keys=query_params.keys
             )
         )
 
     if not aggregates:
         raise web.HTTPNotFound(text="No aggregate found for this address")
 
+    if query_params.value_only and query_params.keys and len(query_params.keys) == 1:
```
hoh, line 71: What happens else ? For example if there are multiple keys ?

src/aleph/web/controllers/aggregates.py
```diff
@@ -62,17 +60,27 @@ async def address_aggregate(request: web.Request) -> web.Response:
...
+        output = {}
+        target_key = query_params.keys[0]
+        for result in aggregates:
+            output[result[0]] = result[1]
```
hoh, line 75: Why not return it here ?

src/aleph/web/controllers/aggregates.py
MHHukiewitz: It will continue as normal

src/aleph/web/controllers/aggregates.py
MHHukiewitz: There might be multiple but we are only interested in the target_key

MHHukiewitz (approved):
LGTM

Issue #564, opened 2024-04-15
Upload Endpoint file reading


1yam:
On : [Storage Endpoint](https://github.com/aleph-im/pyaleph/blob/555c166eddfd8370411308c6909b56e2d7908925/src/aleph/web/controllers/storage.py#L241)

**Problem Description:**
The current implementation of the file upload endpoint reads the entire file before checking if it exceeds the maximum allowed size for unauthenticated uploads (`MAX_UNAUTHENTICATED_UPLOAD_FILE_SIZE = 25 * MiB`) or  (`MAX_UPLOAD_FILE_SIZE = 1000 * MiB`) for authenticated uploads.

**Impact:**
This leads to unnecessary memory usage and potential performance degradation, especially with large files or high volumes of traffic, which could risk server stability.

**Suggested Improvement:**
The file size should be checked against the maximum allowed size before the file is fully read into memory to prevent these issues.

PR #565, opened 2024-04-25
Fix: Upload Endpoints

1yam:
[Issue](https://github.com/aleph-im/pyaleph/issues/564)

Issue Description:
   The current file upload endpoint reads the entire file before checking if it exceeds the maximum allowed size. The size limits are 25 MB for unauthenticated uploads (MAX_UNAUTHENTICATED_UPLOAD_FILE_SIZE) and 1000 MB for authenticated uploads (MAX_UPLOAD_FILE_SIZE).

Proposed Solutions:
- Read the file in chunks and check if the size exceeds the maximum allowed size.
- Use a temporary file to handle the data, which helps avoid using too much memory.

### DIFF
--- a/src/aleph/web/controllers/storage.py
+++ b/src/aleph/web/controllers/storage.py
@@ -148,10 +148,10 @@ class UploadedFile:
 
 
 class MultipartUploadedFile(UploadedFile):
-    def __init__(self, file_field: BodyPartReader, max_size: int, file_hash: str = None):
+    def __init__(self, file_field: BodyPartReader, max_size: int):
         self.file_field = file_field
         self.max_size = max_size
-        self.file_hash = file_hash
+        self._hash = None
         try:
             self._temp_file = tempfile.NamedTemporaryFile(delete=False)
             self._file_content = bytearray()
@@ -189,7 +189,7 @@ class MultipartUploadedFile(UploadedFile):
             self._temp_file.write(chunk)
             self._file_content.extend(chunk)
             hash_sha256.update(chunk)
-        self.file_hash = hash_sha256.hexdigest()
+        self._hash = hash_sha256.hexdigest()
         self._temp_file.seek(0)
 
     @property
@@ -206,7 +206,7 @@ class MultipartUploadedFile(UploadedFile):
 
     @property
     def get_hash(self) -> str:
-        return self.file_hash
+        return self._hash
 
 
 class RawUploadedFile(UploadedFile):
--- a/src/aleph/web/controllers/storage.py
+++ b/src/aleph/web/controllers/storage.py
@@ -1,5 +1,7 @@
 import base64
 import logging
+import os
+import tempfile
 from decimal import Decimal
 from typing import Optional, Protocol, Union
 
@@ -92,7 +94,7 @@ async def add_storage_json_controller(request: web.Request):
 
 
 async def _verify_message_signature(
-    pending_message: BasePendingMessage, signature_verifier: SignatureVerifier
+        pending_message: BasePendingMessage, signature_verifier: SignatureVerifier
 ) -> None:
     try:
         await signature_verifier.verify_signature(pending_message)
@@ -114,56 +116,81 @@ class StorageMetadata(pydantic.BaseModel):
     sync: bool
 
 
-class UploadedFile(Protocol):
-    @property
-    def size(self) -> int: ...
+class UploadedFile:
+    def __enter__(self):
+        return self
 
-    @property
-    def content(self) -> Union[str, bytes]: ...
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.cleanup()
 
-
-class MultipartUploadedFile:
-    _content: Optional[bytes]
-    size: int
-
-    def __init__(self, file_field: FileField, max_size: int):
-        self.file_field = file_field
-        self.max_size = max_size
-        self.size = 0
-        self._content = None
-
-    @property
-    def content(self) -> bytes:
-        """Lazily loads the content, ensuring it is only read once and within the max size limit."""
-        if self._content is None:
-            self._content = b""
-            total_read = 0
-            chunk_size = 8192
-
-            while total_read < self.max_size:
-                chunk = self.file_field.file.read(chunk_size)
-                if not chunk:
-                    break
-                self.size += len(chunk)
-                total_read += len(chunk)
-                if total_read > self.max_size:
-                    raise web.HTTPRequestEntityTooLarge(
-                        reason="File size exceeds the maximum limit.",
-                        max_size=self.max_size,
-                        actual_size=total_read,
-                    )
-                self._content += chunk
-
-        return self._content
-
-
-class RawUploadedFile:
-    def __init__(self, content: Union[bytes, str]):
-        self.content = content
+    def cleanup(self):
+        pass
 
     @property
     def size(self) -> int:
-        return len(self.content)
+        raise NotImplementedError
+
+    @property
+    def content(self) -> bytes:
+        raise NotImplementedError
+
+
+class MultipartUploadedFile(UploadedFile):
+    def __init__(self, file_field: FileField, max_size: int):
+        self.file_field = file_field
+        self.max_size = max_size
+        self._temp_file = tempfile.NamedTemporaryFile(delete=False)
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        super().__exit__(exc_type, exc_value, traceback)
+        os.unlink(self._temp_file.name)
+
+    @property
+    def content(self) -> bytes:
+        total_read = 0
+        chunk_size = 8192
+        while total_read < self.max_size:
+            chunk = self.file_field.file.read(chunk_size)
+            if not chunk:
+                break
+            total_read += len(chunk)
+            if total_read > self.max_size:
+                raise web.HTTPRequestEntityTooLarge(
+                    reason="File size exceeds the maximum limit.",
+                    max_size=self.max_size,
+                    actual_size=total_read,
+                )
+            self._temp_file.write(chunk)
+        self._temp_file.seek(0)
+        return self._temp_file.read()
+
+    @property
+    def size(self) -> int:
+        return os.path.getsize(self._temp_file.name)
+
+
+class RawUploadedFile(UploadedFile):
+    def __init__(self, content: Union[bytes, str]):
+        self._temp_file = tempfile.NamedTemporaryFile(delete=False, mode='w+b')
+        # Encode only if the content is a string
+        if isinstance(content, str):
+            content = content.encode('utf-8')  # Explicitly encode str to bytes
+        self._temp_file.write(content)
+        self._temp_file.seek(0)
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        super().__exit__(exc_type, exc_value, traceback)
+        os.unlink(self._temp_file.name)
+
+    @property
+    def content(self) -> bytes:
+        self._temp_file.seek(0)
+        return self._temp_file.read()
+
+    @property
+    def size(self) -> int:
+        self._temp_file.seek(0, os.SEEK_END)
+        return self._temp_file.tell()
 
 
 async def _check_and_add_file(
@@ -197,9 +224,6 @@ async def _check_and_add_file(
 
     # TODO: this can still reach 1 GiB in memory. We should look into streaming.
     file_content = file.content
-    file_bytes = (
-        file_content.encode("utf-8") if isinstance(file_content, str) else file_content
-    )
     file_hash = get_sha256(file_content)
 
     if message_content:
@@ -210,7 +234,7 @@ async def _check_and_add_file(
 
     await storage_service.add_file_content_to_local_storage(
         session=session,
-        file_content=file_bytes,
+        file_content=file_content,
         file_hash=file_hash,
     )
 
@@ -224,9 +248,9 @@ async def _check_and_add_file(
 
 
 async def _make_mq_queue(
-    request: web.Request,
-    sync: bool,
-    routing_key: Optional[str] = None,
+        request: web.Request,
+        sync: bool,
+        routing_key: Optional[str] = None,
 ) -> Optional[aio_pika.abc.AbstractQueue]:
     if not sync:
         return None
@@ -282,21 +306,22 @@ async def storage_add_file(request: web.Request):
         message = None
         sync = False
 
-    if uploaded_file.size > max_upload_size:
-        raise web.HTTPRequestEntityTooLarge(
-            actual_size=uploaded_file.size, max_size=max_upload_size
-        )
+    with uploaded_file:
+        if uploaded_file.size > max_upload_size:
+            raise web.HTTPRequestEntityTooLarge(
+                actual_size=uploaded_file.size, max_size=max_upload_size
+            )
 
-    with session_factory() as session:
-        file_hash = await _check_and_add_file(
-            session=session,
-            signature_verifier=signature_verifier,
-            storage_service=storage_service,
-            message=message,
-            file=uploaded_file,
-            grace_period=grace_period,
-        )
-        session.commit()
+        with session_factory() as session:
+            file_hash = await _check_and_add_file(
+                session=session,
+                signature_verifier=signature_verifier,
+                storage_service=storage_service,
+                message=message,
+                file=uploaded_file,
+                grace_period=grace_period,
+            )
+            session.commit()
 
     if message:
         broadcast_status = await broadcast_and_process_message(
--- a/src/aleph/web/controllers/storage.py
+++ b/src/aleph/web/controllers/storage.py
@@ -121,9 +121,9 @@ class StorageMetadata(pydantic.BaseModel):
 class UploadedFile:
     def __init__(self, max_size: int):
         self.max_size = max_size
+        self.hash = ""
+        self.size = 0
         self._hasher = hashlib.sha256()
-        self._hash = ""
-        self._size = 0
         self._temp_file_path = None
         self._temp_file = None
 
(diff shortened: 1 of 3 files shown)


hoh:
@philogicae this should fix the file upload issue you reported.

1yam:
https://github.com/aleph-im/pyaleph/blob/885c59cd709f33ec5c54dd2abcd807147a600dfb/src/aleph/web/controllers/storage.py#L272

Based on aiohttp post method looks like it's made for MultipartForm:

```
If method is not POST, PUT, PATCH, TRACE or DELETE or [content_type](https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.BaseRequest.content_type) is not empty or application/x-www-form-urlencoded or multipart/form-data returns empty multidict.
```

Maybe should add a check and use read() for RawUploadedFile ?

aliel:
You might have noticed a big warning in the example above. The general issue is that [aiohttp.web.BaseRequest.post()](https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.BaseRequest.post) reads the whole payload in memory, resulting in possible OOM errors. To avoid this, for multipart uploads, you should use [aiohttp.web.BaseRequest.multipart()](https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.BaseRequest.multipart) which returns a [multipart reader](https://docs.aiohttp.org/en/stable/multipart.html#aiohttp-multipart):
=>
https://docs.aiohttp.org/en/stable/web_quickstart.html#file-uploads

src/aleph/web/controllers/storage.py
```diff
@@ -114,59 +117,168 @@ class StorageMetadata(pydantic.BaseModel):
...
+            if not chunk:
+                break
+            total_read += len(chunk)
+            if total_read > self.max_size:
+                raise web.HTTPRequestEntityTooLarge(
+                    reason="File size exceeds the maximum limit.",
+                    max_size=self.max_size,
+                    actual_size=total_read,
+                )
+            self._temp_file.write(chunk)
```
Psycojoker, line 188: you might want to use https://github.com/Tinche/aiofiles/ to do async IO operation here to avoid blocking the loop (even through I'm not sure writing something this small will have a great performance impact)

src/aleph/web/controllers/storage.py
```diff
@@ -114,59 +117,168 @@ class StorageMetadata(pydantic.BaseModel):
...
+        chunk_size = 8192
+        hash_sha256 = hashlib.sha256()
+
+        while total_read < self.max_size:
+            chunk = await self.file_field.read_chunk(chunk_size)
+            if not chunk:
+                break
+            total_read += len(chunk)
+            if total_read > self.max_size:
+                raise web.HTTPRequestEntityTooLarge(
```
Psycojoker, line 183: Shouldn't you close and delete the temporary file here before raising?

src/aleph/web/controllers/storage.py
```diff
@@ -114,59 +117,168 @@ class StorageMetadata(pydantic.BaseModel):
...
+        return bytes(self._file_content)
 
-        return self._content
+    @property
+    def file(self) -> str:
+        return self._temp_file.name
+
+    def get_hash(self) -> str:
+        if self._hash is None:
+            raise ValueError("Hash has not been computed yet")
```
Psycojoker, line 208: For easier debugging you should add more information in this error like what you need to do for the hash to be computed

src/aleph/web/controllers/storage.py
```diff
@@ -114,59 +117,168 @@ class StorageMetadata(pydantic.BaseModel):
...
-    def __init__(self, file_field: FileField, size: int):
+class MultipartUploadedFile(UploadedFile):
+    def __init__(self, file_field: BodyPartReader, max_size: int):
         self.file_field = file_field
-        self.size = size
-        self._content = None
+        self.max_size = max_size
+        self._hash = None
+        try:
+            self._temp_file = tempfile.NamedTemporaryFile(delete=False)
```
Psycojoker, line 155: I think it would make more sens to create the temporary file in `__enter__` to only create it when you need it and having hanging temporary files before the workflow has effectively started.

src/aleph/web/controllers/storage.py
```diff
@@ -114,59 +117,168 @@ class StorageMetadata(pydantic.BaseModel):
...
+        self._temp_file.seek(0)
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        try:
+            self._temp_file.close()
+            if os.path.exists(self._temp_file.name):
+                os.unlink(self._temp_file.name)
+        except Exception as e:
+            web.HTTPInternalServerError(reason="Cannot create tempfile")
```
Psycojoker, line 170: It looks like you forgot to modify the error message here :sweat_smile:

src/aleph/web/controllers/storage.py
```diff
@@ -114,59 +117,168 @@ class StorageMetadata(pydantic.BaseModel):
...
+                raise web.HTTPRequestEntityTooLarge(
+                    reason="File size exceeds the maximum limit.",
+                    max_size=self.max_size,
+                    actual_size=total_read,
+                )
+            self._temp_file.write(chunk)
+            self._file_content.extend(chunk)
+            hash_sha256.update(chunk)
+        self._hash = hash_sha256.hexdigest()
+        self._temp_file.seek(0)
```
Psycojoker, line 192: Is there a reason on why you do that? You probably want to add a comment here explaining this because it doesn't seem obvious to me so I guess other people won't guess it neither.

src/aleph/web/controllers/storage.py
```diff
@@ -114,59 +117,168 @@ class StorageMetadata(pydantic.BaseModel):
...
+        except Exception as e:
+            web.HTTPInternalServerError(reason="Cannot create tempfile")
+
+    async def read_and_validate(self):
+        total_read = 0
+        chunk_size = 8192
+        hash_sha256 = hashlib.sha256()
+
+        while total_read < self.max_size:
+            chunk = await self.file_field.read_chunk(chunk_size)
```
Psycojoker, line 178: You probably want to do an "async for" like you did here https://github.com/aleph-im/pyaleph/pull/565/files#diff-3ac5ca0d72a6758698e8f83939e455e54c6d97047c7c4310b7f9e3dbd963b2d7R222 since the same condition is being taken care of in the if line 182 https://github.com/aleph-im/pyaleph/pull/565/files#diff-3ac5ca0d72a6758698e8f83939e455e54c6d97047c7c4310b7f9e3dbd963b2d7R182

src/aleph/web/controllers/storage.py
```diff
@@ -114,59 +117,168 @@ class StorageMetadata(pydantic.BaseModel):
     sync: bool
 
 
-class UploadedFile(Protocol):
+class UploadedFile:
+    def __enter__(self):
+        raise NotImplementedError
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.cleanup()
```
Psycojoker, line 125: I haven't seen this cleanup used anywhere, maybe we don't need it?

src/aleph/web/controllers/storage.py
```diff
@@ -114,59 +117,168 @@ class StorageMetadata(pydantic.BaseModel):
...
+        self.request = request
+        self.max_size = max_size
+        self._temp_file = tempfile.NamedTemporaryFile(delete=False)
+        self._hasher = hashlib.sha256()
+        self._size = 0
+        self._hash = None
+
+    async def read_and_validate(self):
+        async for chunk in self.request.content.iter_chunked(8192):
+            self._temp_file.write(chunk)
```
Psycojoker, line 223: Same comment than with aiofiles here https://github.com/aleph-im/pyaleph/pull/565/files#r1618964708

src/aleph/web/controllers/storage.py
```diff
@@ -114,59 +117,168 @@ class StorageMetadata(pydantic.BaseModel):
...
+            self._hasher.update(chunk)
+            self._size += len(chunk)
+            if self._size > self.max_size:
+                raise web.HTTPRequestEntityTooLarge(
+                    reason="File size exceeds the maximum limit.",
+                    max_size=self.max_size,
+                    actual_size=self._size,
+                )
+        self._temp_file.seek(0)
```
Psycojoker, line 232: Same comment than here https://github.com/aleph-im/pyaleph/pull/565/files#r1618976201

src/aleph/web/controllers/storage.py
```diff
@@ -114,59 +117,168 @@ class StorageMetadata(pydantic.BaseModel):
...
+        self._size = 0
+        self._hash = None
+
+    async def read_and_validate(self):
+        async for chunk in self.request.content.iter_chunked(8192):
+            self._temp_file.write(chunk)
+            self._hasher.update(chunk)
+            self._size += len(chunk)
+            if self._size > self.max_size:
+                raise web.HTTPRequestEntityTooLarge(
```
Psycojoker, line 227: same comment than here https://github.com/aleph-im/pyaleph/pull/565/files#r1618967347

src/aleph/web/controllers/storage.py
```diff
@@ -114,59 +117,168 @@ class StorageMetadata(pydantic.BaseModel):
...
+        self._temp_file.seek(0)
+        self._hash = self._hasher.hexdigest()
+
+    def __enter__(self):
+        self._temp_file.seek(0)
+        return self._temp_file
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self._temp_file.close()
+        os.unlink(self._temp_file.name)
```
Psycojoker, line 241: Is it really needed? (genuine question)

src/aleph/web/controllers/storage.py
```diff
@@ -114,59 +117,168 @@ class StorageMetadata(pydantic.BaseModel):
...
+    def size(self) -> int:
+        return self._size
 
-class RawUploadedFile:
-    def __init__(self, content: Union[bytes, str]):
-        self.content = content
+    @property
+    def content(self) -> bytes:
+        self._temp_file.seek(0)
+        return self._temp_file.read()
```
Psycojoker, line 250: Same comment with aiofiles https://github.com/aleph-im/pyaleph/pull/565/files#r1618964708

src/aleph/web/controllers/storage.py
```diff
@@ -114,59 +117,168 @@ class StorageMetadata(pydantic.BaseModel):
...
 
     @property
-    def size(self) -> int:
-        return len(self.content)
+    def file(self):
```
Psycojoker, line 253: You probably want to name this `file_name`

src/aleph/web/controllers/storage.py
```diff
@@ -114,59 +117,168 @@ class StorageMetadata(pydantic.BaseModel):
...
+        return self._temp_file.name
+
+    def get_hash(self) -> str:
```
Psycojoker, line 256: Maybe this can move in the parent class? This is the same code on both classes

src/aleph/web/controllers/storage.py
```diff
@@ -177,22 +289,17 @@ async def _check_and_add_file(
...
-        if message_content.item_hash != file_hash:
-            raise web.HTTPUnprocessableEntity(
-                reason=f"File hash does not match ({file_hash} != {message_content.item_hash})"
-            )
+    if isinstance(file_content, bytes):
+        file_bytes = file_content
+    elif isinstance(file_content, str):
+        file_bytes = file_content.encode("utf-8")
+    else:
+        raise web.HTTPUnprocessableEntity(reason="Invalid file content type")
```
Psycojoker, line 302: Here, for easier debugging, you want to show the effective file type and the accepted file types

src/aleph/web/controllers/storage.py
```diff
@@ -230,19 +337,27 @@ async def storage_add_file(request: web.Request):
...
+            elif part.name == 'metadata':
+                metadata = await part.read(decode=True)
     else:
-        uploaded_file = RawUploadedFile(file_field)
+        uploaded_file = RawUploadedFile(request=request, max_size=MAX_UNAUTHENTICATED_UPLOAD_FILE_SIZE)
+        await uploaded_file.read_and_validate()
 
-    metadata = post.get("metadata")
+    if uploaded_file is None:
+        raise web.HTTPBadRequest(reason="No file uploaded")
```
Psycojoker, line 356: This one looks like a hard error to debug with very little indication of the effective problem :/

src/aleph/web/controllers/storage.py
```diff
@@ -230,19 +337,27 @@ async def storage_add_file(request: web.Request):
...
+        reader = await request.multipart()
+        async for part in reader:
+            if part.name == 'file':
+                uploaded_file = MultipartUploadedFile(part, MAX_FILE_SIZE)
+                await uploaded_file.read_and_validate()
+            elif part.name == 'metadata':
+                metadata = await part.read(decode=True)
     else:
-        uploaded_file = RawUploadedFile(file_field)
+        uploaded_file = RawUploadedFile(request=request, max_size=MAX_UNAUTHENTICATED_UPLOAD_FILE_SIZE)
```
Psycojoker, line 352: same comment than above here https://github.com/aleph-im/pyaleph/pull/565/files#r1619000450

src/aleph/web/controllers/storage.py
Psycojoker: And there is probably more code that can move to the parent class I think (but I haven't reviewed with this in mind)

Psycojoker:
I did a quick general review because I don't know the business code very well and some of them are more questions/suggestions, don't hesitate if you have questions :)

src/aleph/web/controllers/storage.py
1yam: Normally this is be done by __exit__

src/aleph/web/controllers/storage.py
1yam: The .post() method we were using before did the same things (split and write chunks to a temp file) so probably still a good way to do, we could also use aiofiles.

The second point is that reading by chunk avoids loading the full file into memory, allowing us to check the size while reading the file data.

src/aleph/web/controllers/storage.py
1yam: Well, I think this is parts can just be removed, cause the only way the hash would not be set would be in the case no file was sent and so this error would never be ever raise

src/aleph/web/controllers/storage.py
Psycojoker: Yes, I know this kind of pattern, the problem is that a file writing is still a blocking IO operation and we probably don't want to block the whole loop here

1yam:
@Psycojoker I did some updates:
- Refactored `UploadFile` to avoid code duplication
- Used `aiofiles`
- Fixing most of you'r comment

The main thing I didn't change is that we're still reading and writing the file to a temporary file first to avoid launching the workflow if the file is too large and we need to calculate the hash early for safety check.

I wasn't using the manager correctly, that should be fixed now.

src/aleph/web/controllers/storage.py
```diff
@@ -230,19 +337,27 @@ async def storage_add_file(request: web.Request):
...
-    if isinstance(file_field, FileField):
-        uploaded_file: UploadedFile = MultipartUploadedFile(file_field, len(file_field.file.read()))
+    metadata = None
+    uploaded_file: Optional[UploadedFile] = None
+
+    if request.content_type == "multipart/form-data":
+        reader = await request.multipart()
+        async for part in reader:
+            if part.name == 'file':
+                uploaded_file = MultipartUploadedFile(part, MAX_FILE_SIZE)
```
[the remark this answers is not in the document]
Psycojoker: Did you forgot to move to a context manager here or forget to push a commit '-'? Maybe "context manager" is not an obvious term ^^' 

This could should look like something like this:

```python
async with MultipartUploadedFile(part, MAX_FILE_SIZE) as uploaded_file:
    await uploaded_file.read_and_validate()
```

And same line 291:

```python
async with RawUploadedFile(request=request, max_size=MAX_UNAUTHENTICATED_UPLOAD_FILE_SIZE) as uploaded_file:
    await uploaded_file.read_and_validate()
```

src/aleph/web/controllers/storage.py
Psycojoker: Reading [BodyPartReader source code](https://docs.aiohttp.org/en/stable/_modules/aiohttp/multipart.html#BodyPartReader), it does support the async iterator protocol so you should be able to to rewrite this code like this:

```python
async for chunk in self.file_field.read_chunk(chunk_size):
    yield chunk
```

Edit: turns out you did already use `async for here https://github.com/aleph-im/pyaleph/pull/565/files?diff=split#diff-3ac5ca0d72a6758698e8f83939e455e54c6d97047c7c4310b7f9e3dbd963b2d7R198-R200

So maybe this method isn't really needed? You can directly use `async for` [line 154](https://github.com/aleph-im/pyaleph/pull/565/files?diff=split#diff-3ac5ca0d72a6758698e8f83939e455e54c6d97047c7c4310b7f9e3dbd963b2d7R154) and instead and remove the 3 `_read_chunks` methods to simplify code

src/aleph/web/controllers/storage.py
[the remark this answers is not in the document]
Psycojoker: Which actually wouldn't make that much sens here since uploaded_file would be deleted once you get out of the context manager.

I don't think using a context manager here make a lot of sens or the code logic needs to be changed :/

Why did you chose to use a context manager?

tests/api/test_storage.py
```diff
@@ -101,6 +136,7 @@ async def add_file(
 ):
     form_data = aiohttp.FormData()
     form_data.add_field("file", file_content)
+    print(file_content)
```
Psycojoker, line 139: This looks like a debug print, shouldn't it be removed?

Psycojoker:
Hugo suggested that we could do a call at some point if you want me to explain some of my comments.

Psycojoker:
We did a pair programming session and that should be good on my side regard good python practices.

Lyam told me he needed to do one last test with the frontend before confirming it's good.

Psycojoker (approved):

1yam:
@hoh @Psycojoker Just finish my last test, everything looks working as intended, I would say ready for merge

PR #566, opened 2024-05-16
Solve Flow pricing calculation issues

nesitor:
Fix: Required flow price calculation was failing, now it is fixed also with tests.


olethanh (approved):
haven't tested it but LGTM

PR #570, opened 2024-07-03
a lot of fixes and improvement for running local tests using nix-shell because it was just broken

Psycojoker:
List of modifications and improvements: 

* add pkgs.glibcLocales because otherwise locales aren't handles and postgresql fails
* ensure that we have at least a default config.yml otherwise tests won't launch
* display information information at the end of the output log of nix-shell and in bold otherwise it's hidden in the logs
* wrap shellHooks in `set -eu` otherwise you can have error, nix won't complain and you'll end up in a broken nix-shell not understanding why
* make tests fixtures take into account config.yml otherwise it's simply ignored and only the default are used
* launch postgresql on port 5434 because users can have postgresql on 5432 (or even 5433) and nix-shell will just fails 
* and remove an unused import

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -46,14 +46,6 @@ pkgs.mkShell {
     ipfs daemon &
     echo "IPFS Kubo started. Data directory is $IPFS_PATH"
 
-    echo
-    echo "PostgreSQL started. Data directory is $PGDATA, Socket directory is $PG_SOCKET_DIR"
-    echo "Redis started. Data directory is $REDIS_DATA_DIR"
-    echo "Use 'psql -h $PG_SOCKET_DIR' to connect to the database."
-    echo "Use 'redis-cli -p 6379' to connect to the Redis server."
-    echo "To stop PostgreSQL: 'pg_ctl -D $PGDATA stop'"
-    echo "To manually stop Redis: 'redis-cli -p 6379 shutdown'"
-
     # Trap the EXIT signal to stop services when exiting the shell
     trap 'echo "Stopping PostgreSQL..."; pg_ctl -D "$PGDATA" stop; echo "Stopping Redis..."; redis-cli -p 6379 shutdown; echo "Stopping IPFS Kubo..."; ipfs shutdown; deactivate' EXIT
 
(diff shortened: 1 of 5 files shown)


Psycojoker:
Anyway, I just wanted to run tests locally to review another PR

tenor

shell.nix
```diff
@@ -65,5 +62,20 @@ pkgs.mkShell {
 
     # Activate the virtual environment
     source venv/bin/activate
+
+    [ -e config.yml ] || echo -e "postgres:\n  port: 5434" > config.yml
```
hoh, line 66:
```suggestion
    # If config.yml does not exist, create it with the port specified in this shell. 
    [ -e config.yml ] || echo -e "postgres:\n  port:$PG_PORT" > config.yml
```

tests/conftest.py
```diff
@@ -98,6 +98,16 @@ def session_factory(mock_config):
 def mock_config(mocker):
     config = Config(aleph.config.get_defaults())
 
+    config_file = Path.cwd() / "config.yml"
+
+    if config_file.exists():
+        with config_file.open() as f:
+            user_config = f.read()
```
hoh, line 105:
```suggestion
        user_config = config_file.read_text()
```

tests/conftest.py
```diff
@@ -98,6 +98,16 @@ def session_factory(mock_config):
...
+
+        # Little trick to allow empty config files
+        if user_config:
+            config.yaml.loads(user_config)
```
hoh, line 109: Would it be worth renaming one of these variables for typing coherence ?
( `user_config_raw: str` and `user_config: dict` for example)

hoh (commented):
Looks good overall, left a few minor comments / suggestions.

tests/conftest.py
Psycojoker: yeah that would make sens, let's do it

Psycojoker: It should be better now, I took the occasion to fix a bug because cli_args was returning a string a not a Path object

hoh:
I see this error many times in the CI:

```
ERROR tests/web/controllers/test_pub_json.py::test_pub_invalid_aleph_message - AttributeError: 'str' object has no attribute 'items'
```

```
 >       for name, value in dictionary.items():
E       AttributeError: 'str' object has no attribute 'items'
```

Psycojoker:
This should be all good now.

Psycojoker:
> I see this error many times in the CI:
> 
> ```
> ERROR tests/web/controllers/test_pub_json.py::test_pub_invalid_aleph_message - AttributeError: 'str' object has no attribute 'items'
> ```
> 
> ```
>  >       for name, value in dictionary.items():
> E       AttributeError: 'str' object has no attribute 'items'
> ```

Oups, miss you comment.

It's fix now, it was because you removed the space after the ":" here because yaml is weird:

image

hoh:
Approved. Wanna squash ?

Psycojoker:
> Approved. Wanna squash & merge ?

Like you want, for me those are independent commits and you might wanna revert them at some point but I don't mind if you prefer to squash them 👍

PR #571, opened 2024-07-03
UX improvments for nix-shell

Psycojoker:
- **chore(nix-shell/ux): display usefull information at the end of nix-shell output**
- **chore(nix-shell/ux): add instructions in bold so people don't miss it**

It looks like that now:

image


Psycojoker:
Merged everything in #570 because there is just too much stuff to fix and improve.

PR #572, opened 2024-07-03
chore(nix-shell/fix): ensure at least a default config.yml exist otherwise tests won't run

Psycojoker:
This also need the PR #570 to be merged, otherwise just running:

```shell
nix-shell --run "pytest"
```

Simply won't work.

I'll also add a PR to handle the situation where postgresql port can colide
with an existing locally running postgresql database because this can often be
the case.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -65,5 +65,7 @@ pkgs.mkShell {
 
     # Activate the virtual environment
     source venv/bin/activate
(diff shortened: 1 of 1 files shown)


Psycojoker:
Merged everything in #570 because there is just too much stuff to fix and improve.

PR #573, opened 2024-07-08
Feature: Allow processing of Confidential message

nesitor:


Psycojoker (approved):
I can't tell business knowledge wide but code wide this PR LGTM.

I've rebased merged and reorder commits and split this into 2 PR with https://github.com/aleph-im/pyaleph/pull/578 being only the mypy part so this should be easier to review.

This can either be squased or rebase.

PR #574, opened 2024-07-09
ALEPH-7 project modernization: migrate project to python 3.12, hatch, pyproject.toml, upgrade dependencies, migrate to latest pytezos and add a bunch of other improvements and some documentation improvement.

Psycojoker:
Content of this PR (for now):

* migrate to hatch and pyproject.toml
* modify the workflows so they match the ones of the others projects
* add instructions on how to run tests outside of nix/docker
* on ci add a code quality job
* disabled nix CI jobs because they just don't work anymore and it's not a priority right now

I suggest to read this PR commit by commit.

Split with another small PR https://github.com/aleph-im/pyaleph/pull/584


Psycojoker:
Ok, this PR should FINALLY be good, I've fixed a lot of small things and tried to move as close as possible to other projects (like the python SDK) to uniformise things. 

There still might be weird things in the github workflows but some of them come from the old workflow or from the python SDK ones. I think we can refine those later if needed.

hoh:
See my fixes in another PR based on this branch, and the annoying requirement to rebase on `main` :roll_eyes:

Psycojoker:
Current status and next steps:

- nix-os is being extremely complicated to work on and debug, we might drop it for now since it's not a priority
- tests are working, everything seems ready on that side
- mypy is raising weird issues that we need to fix
- I've used the newly released version of pytezos (3.13.3) that works the way we want but it's more likely that we want to update [our fork](https://github.com/aleph-im/aleph-pytezos) instead of avoid pulling too many dependencies

hoh:
As discussed, we can disable the `shell.nix` requirement in the CI if that is blocking us.

Psycojoker:
Those mypy bugs are extremely annoying:

* pydantic wants a `Literal[something]`
* we uses `Literal[SomeEnum.key]` which works for pydantic
* but mypy considers that `SomeEnum.key` IS a `Literal` and thus think that doing `Literal[Literal]` makes : no sens at all
* we probably should migrate to pydantic `2.8` because we are still using `1.10` but this will need some code modification

And: this whole issue is actually absolutely not related to migrating to pyproject, it has been introduced by other commits  (I haven't took the time to find which ones) but turns out this PR enforce using mypy in the CI.

Psycojoker:
This PR should be ready to review

philogicae (approved):
Stable now, no detected error during runtime, LGTM

Psycojoker:
I've convert it to draft since we want to wait for a release of pyaleph before merging this PR.

hoh (approved):

PR #580, opened 2024-07-15
Fix issues of #574 regarding dependencies

hoh:
Attempt to fix the issues of #574

- **Fix: Pyproject requires `hatch` > 1.9.7 (nixpkgs 24.05)**
- **Fix: nix-shell did not use latest stable channel**
- **Fix: pyproject.toml used an old version of aleph-message**
- **Fix: pyproject.toml used fork of aioipfs**

Remains one issue with hatch since I updated to use the unstable channel of Nixpkgs, which is not available by default in the CI.


Psycojoker:
I've cherry-picked your 4 commits directly on the pyproject branch because rebasing it was way too complicated.

Thx for your work!

Issue #576, opened 2024-07-09
Error 500 on /api/v0/price


hoh:
scheduler_1  | aiohttp.client_exceptions.ClientResponseError: 500, message='Internal Server Error', url=URL('https://api2.aleph.im/api/v0/price/a203afe59e02421cf992c24889aea24261452c1940612098dfd8a59fe6b1f664')

1yam:
The endpoints seem working, look like the volume is missing

```
RuntimeError: Could not find entry in file tags for d4a9f4abb451edb361504cc093e78e2d507fb3bb9244ffd746ababaf4c8537a9.
```

But still it should return us more details and not just a code 500

PR #581, opened 2024-07-16
Fix: Handle no volume found instead of error 500

1yam:
[Issue](https://github.com/aleph-im/pyaleph/issues/576)

Problem:
When a volume is not found, it's returning error 500 instead of returning a web response 

Solution:
Catch runtime error and return the error as web error

### DIFF
--- a/src/aleph/web/controllers/prices.py
+++ b/src/aleph/web/controllers/prices.py
@@ -81,11 +81,13 @@ async def message_price(request: web.Request):
         message = await get_executable_message(session, request.match_info["item_hash"])
 
         content: ExecutableContent = message.parsed_content
-
-        if content.payment and content.payment.is_stream:
-            required_tokens = compute_flow_cost(session=session, content=content)
-        else:
-            required_tokens = compute_cost(session=session, content=content)
+        try:
+            if content.payment and content.payment.is_stream:
+                required_tokens = compute_flow_cost(session=session, content=content)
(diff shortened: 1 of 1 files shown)


1yam:
https://github.com/aleph-im/pyaleph/issues/576

src/aleph/web/controllers/prices.py
```diff
@@ -81,11 +81,13 @@ async def message_price(request: web.Request):
...
-        if content.payment and content.payment.is_stream:
-            required_tokens = compute_flow_cost(session=session, content=content)
-        else:
-            required_tokens = compute_cost(session=session, content=content)
+        try:
+            if content.payment and content.payment.is_stream:
+                required_tokens = compute_flow_cost(session=session, content=content)
+            else:
+                required_tokens = compute_cost(session=session, content=content)
+        except RuntimeError as e:
```
hoh, line 89: Can you explain why is a `RuntimeError` error expected here ? Is that error not too generic ?

1yam: [Where it can failed](https://github.com/aleph-im/pyaleph/blob/09b6d5b5ed50a625dcaf82fe3eaaef7f48a4cf63/src/aleph/services/cost.py#L70)
It's in order to handle this error if volume is removed but not the program

nesitor (approved):

PR #584, opened 2024-07-24
Allow to connect postgresql in a passwordless mode and with a unix socket

Psycojoker:
Before that, it was only possible to connect pyaleph to postgresql using an ip/domain and only with a password.

I've also fix a bug in the tests that was overwriting some configuration.

This PR has been extracted from #574 

This PR can be "rebase and merge" since all commits are standalone.


philogicae (approved):
LGTM

PR #585, opened 2024-07-25
doc: remove references to MongoDB

Psycojoker:
We still have reference to MongoDB here which is not that great and professional https://pyaleph.readthedocs.io/en/latest/guides/install.html since it's linked here https://account.aleph.im/earn/ccn/


philogicae (approved):
LGTM

PR #589, opened 2024-08-01
feat(ci): add a new job to test the keys generation

Psycojoker:
This also has the side effect of testing if we manage to launch pyaleph.


philogicae (approved):
LGTM

PR #591, opened 2024-08-06
Fix: check if requirements exist on message

1yam:
Problem:
In some case (hold instance) there is no requirements field

Solutions:
Add check if requirements field exist


Psycojoker:
If possible, add a test on it

1yam:
We already have a test that covers instance message processing without specific requirements, which is test_process_instance. You can find it [here](https://github.com/aleph-im/pyaleph/blob/0e44b8ef2c69feffdc15a7efe40fbc67f52d77e6/tests/message_processing/test_process_instances.py#L252).

I have added a check for the node_hash field to ensure it aligns with the configuration of the instance. As the confidential one do

olethanh (approved):

PR #594, opened 2024-08-13
Fix: nix-shell raised 'pop_var_context'

hoh:
Using nix-shell with the latest version of Nixpkgs raised the following
error:

```
bash: pop_var_context: head of shell_variables not a function context
```

Changing the order of the instructions fixed this.


Psycojoker (approved):

Psycojoker:
Thx!

PR #595, opened 2024-08-13
feat(pyproject): add check-dist to ensure we ship all the needed files

Psycojoker:
I wanted to use check-manifest at first but it doesn't support pyproject.toml
yet so I've switch to [check-dist](https://github.com/henryiii/check-sdist) which seems to do the job.

I'm not 100% convinced by it (it doesn't take into account the exclude
directive of hatch and you need to use a custom config to tell him to ignore
some files) but it should do the trick.

Like that we'll avoid error like releasing aleph-message without including "py.typed" file.


hoh:
Thanks ! Should this be added to other repositories ?

hoh (approved):

Psycojoker:
> Thanks! Should this be added to other repositories?

Yes, that's the idea, same for other tools like pyproject-fmt, yamlix etc ... so we can spend less time on all of that.

PR #596, opened 2024-08-14
Provide a template for new PRs

olethanh:


Psycojoker (approved):
I don't think this is the same template that on other repos but let's move forward and merge this since it's already good enough and we'll improve it later on if needed.

PR #602, opened 2024-09-19
fix: mock.called_once is not a valid test

Psycojoker:


Psycojoker:
@hoh so, I could actually never fix this broken test on #574 because : it actually never worked as expected, just that the mock library started adding method to ensure you don't do erroneous call like using `called_once` instead of `assert_called_once` 

Do you have an idea if that's a real bug or a bad test?

Psycojoker:
Well, according to https://github.com/aleph-im/pyaleph/pull/603 this test never worked, is it a real bug then?

1yam (approved):
LGTM

PR #611, opened 2024-09-26
Added large ints test case

nesitor:
Fix: Add a test for large ints to ensure don't raises a TypeError

- [X] Are my files well typed
- [X] Are there enough tests

## Changes

Added a test case ensuring that large ints doesn't break the json load and dump

## How to test

Execute the test suite

### DIFF
--- a/tests/toolkit/test_json.py
+++ b/tests/toolkit/test_json.py
@@ -46,3 +46,22 @@ def test_serialized_json_type():
 
     actual = json.loads(serialized_json)
     assert actual == expected
+
+
+def test_loads_large_ints_json():
+    """
+    Check that the output of dumps and loads don't raise TypeError errors caused by large ints on orjson library.
+    """
+
+    expected = {
+        "0x3E1aba4ad853Dd7Aa531aB59F10bd9f4d89aebaF": 498729072221377800000,
+        "0x525C49BF83Ce3a1AAf425ac1A463537dB68c8bd7": 8059602048250472000000,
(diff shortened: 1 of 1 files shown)


hoh (approved):

hoh:
What strategy should be used to avoid breaking Javascript and non-Python clients ?

> Javascript integers are limited to about 2^53

There are already have a few messages stored in the database with too large integers, and loading these from JSON in any language other than Python would result in a client crash if these are served.

Would it make more sense to handle this problem on the server side and not serve these messages ?
(as well as not accepting new ones)

nesitor:
> What strategy should be used to avoid breaking Javascript and non-Python clients ?
> 
> > Javascript integers are limited to about 2^53
> 
> There are already have a few messages stored in the database with too large integers, and loading these from JSON in any language other than Python would result in a client crash if these are served.
> 
> Would it make more sense to handle this problem on the server side and not serve these messages ? (as well as not accepting new ones)

At the moment Javascript don't crash handling this messages with large ints, the unique thing that does is loose precision (see https://stackoverflow.com/questions/34561827/how-do-i-deal-with-number-greater-than-253-in-javascript#:~:text=The%20%C2%B1253%20range,use%20a%20Javascript%20bignum%20library.). And the unique clients that we have for the API is JavaScript and Python.
I think that the API should not assume the task to check if the ints are so large or not, only if it can handle it and store it and not crash. The responsability about what is saved and what is loaded is from the user that saves the message, so if the API can manage, save and serve it without crash it is fine.

PR #652, opened 2024-11-21
Feature: Add new endpoint to list message hashes

aliel:

- [ x] Is my code clear enough and well documented
- [ x] Are my files well typed
- [ ] Are there enough tests

## Changes

## How to test

Explain how to test your PR.

## Print screen / video

## Notes

### DIFF
--- a/src/aleph/web/controllers/messages.py
+++ b/src/aleph/web/controllers/messages.py
@@ -6,12 +6,21 @@ import aio_pika.abc
 import aiohttp.web_ws
 from aiohttp import WSMsgType, web
 from aleph_message.models import Chain, ItemHash, MessageType
-from pydantic import BaseModel, Field, ValidationError, root_validator, validator
+from pydantic import (
+    BaseModel,
+    Field,
+    ValidationError,
+    parse_obj_as,
+    root_validator,
+    validator,
+)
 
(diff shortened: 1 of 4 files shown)


src/aleph/web/controllers/messages.py
```diff
@@ -585,3 +639,36 @@ async def view_message_status(request: web.Request):
 
     status_info = MessageStatusInfo.from_orm(message_status)
     return web.json_response(text=status_info.json())
+
+
+async def view_message_hashes(request: web.Request):
+    print("run once?")
```
nesitor, line 645: This print can be removed?

nesitor (commented):
As I understand, this new endpoint is to obtain just the item_hashes between some dates or by status, but why don't integrate it on the same `/messages` endpoint only adding this new `hash_only` parameter?
Or maybe I'm missing another functionality that we don't integrate in the other endpoint?

src/aleph/web/controllers/messages.py
aliel: its better to keep them separate,
here we use only "message_status" table which doesn't support all filtering like "/messages"
we should also avoid having different outputs for models and keep "/messages" relaying on "message" model

Psycojoker:
It would have been nice to have at least a basic test for this view

PR #653, opened 2024-11-22
Aleph-302: fix messages.json start_block and end_block params

Psycojoker:
Query like `http://0.0.0.0:4024/api/v0/messages.json?msgType=POST&tags=mainnet&contentTypes=balances-update&pagination=50&sort_order=1&start_block=21135862&addresses=0xa1B3bb7d2332383D96b7796B908fB7f7F3c2Be10` was broken and generating a 500 error

Related Clickup or Jira tickets : ALEPH-302

- [x] Are my files well typed

## Changes

So, basically until [this commit](https://github.com/aleph-im/pyaleph/commit/eb3b65369c6daea23616d90e038b02caa837e151) the `start_block/end_block` part of the code was never reached and was not working (except if you call the html query params `StartBlock`)

- first, a select query tried to access the `height` attribute on the table `message_confirmations` that doesn't have any height
- so using a join we grab the transaction `height` value 
- add it to the `GROUP BY`
- then we fix a priority operator because `|` has less precedence than `>=`
- and then we had a default value to `order_by_columns` because in some branches of the `if/elif/else` block it can be uninitialized

## How to test

Launch pyaleph then run this query: `http://0.0.0.0:4024/api/v0/messages.json?msgType=POST&tags=mainnet&contentTypes=balances-update&pagination=50&sort_order=1&start_block=21135862&addresses=0xa1B3bb7d2332383D96b7796B908fB7f7F3c2Be10`

## Notes

I'm not 100% sure that the modification of the `GROUP BY` and of the default value for `order_by_columns` are the behavior we want but this part of the code just never worked before :confused: 

We also probably really want to have some tests here that handle that situation but I understand that this PR is urgent and 301 is also so I'm prioritizing this.

### DIFF
--- a/src/aleph/db/accessors/messages.py
+++ b/src/aleph/db/accessors/messages.py
@@ -138,8 +138,7 @@ def make_matching_messages_query(
         if start_block:
             select_stmt = select_stmt.where(
                 select_earliest_confirmation.c.height.is_(None)
-                | select_earliest_confirmation.c.height
-                >= start_block
+                | (select_earliest_confirmation.c.height >= start_block)
             )
         if end_block:
             select_stmt = select_stmt.where(
--- a/src/aleph/db/accessors/messages.py
+++ b/src/aleph/db/accessors/messages.py
@@ -118,7 +118,7 @@ def make_matching_messages_query(
     if channels:
         select_stmt = select_stmt.where(MessageDb.channel.in_(channels))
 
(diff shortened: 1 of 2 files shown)


nesitor (approved):

Psycojoker:
@nesitor chan you approve this PR again plz? I've just solved the merge conflict

nesitor (approved):

PR #654, opened 2024-11-23
Update settings.py

digvijay-y:
Thread-Safe Singleton:
    Uses a lock to ensure only one instance of Settings is created, even when accessed in a multithreaded environment.

Extendability:
    Additional configuration parameters can easily be added as fields in the Settings dataclass.

Immutability:
    The frozen=True parameter ensures the settings remain immutable after initialization.

- [ ] Are my files well typed
- [ ] Are there enough tests

## Changes

## How to test

Explain how to test your PR.

## Print screen / video

## Notes

### DIFF
--- a/src/aleph/settings.py
+++ b/src/aleph/settings.py
@@ -1,10 +1,27 @@
-from dataclasses import dataclass
+from dataclasses import dataclass, field
+from threading import Lock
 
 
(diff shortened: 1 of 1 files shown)


hoh:
Thanks for your proposal. Can you please fill in the PR template and follow the git commit guidelines of the project ?

PR #663, opened 2024-12-03
feat: upgrade to new version of aleph-message

Psycojoker:


nesitor:
I think that you need to also apply the changes on some tests.

PR #670, opened 2024-12-22
fix: avoid exposing messages/posts without channels

Psycojoker:
We have a bug in production where apparently someone sent messages without a
channel in them and exposing them seems to produce bugs.


aliel:
we should not avoid exposing messages without channels, as its an optional param.

PR #690, opened 2025-01-24
feat: cache all metrics data for 10 seconds for performances reasons

Psycojoker:
The metrics building ends up taking a lot of performance and we wanted to cache it to reducing the CCN load.

This cache all heavy operations for 10 seconds (arbitrary value to discuss).

Those decisions are based on this line_profiling of the function:

(gosh it tooks so much time to get docker-compose is a PITA)

Related Clickup or Jira tickets : ALEPH-362

- [ ] Are my files well typed
- [ ] Are there enough tests

## How to test

Just launch pyaleph and go on "http://localhost:4024/metrics.json"

### DIFF
--- a/src/aleph/config.py
+++ b/src/aleph/config.py
@@ -39,6 +39,13 @@ def get_defaults():
                     "max_concurrency": 20,
                 },
             },
+            "cache": {
+                "ttl": {
+                    "total_aleph_messages": 120,
+                    "eth_height": 600,
+                    "metrics": 10,
+                },
+            },
         },
         "p2p": {
             # Port used for HTTP communication between nodes.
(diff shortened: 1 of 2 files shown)


nesitor (commented):
I see so good the changes, but either we defined if before, I think that we need to test it on staging to ensure if 10 seconds is enough cache time, or we need to increment it before merging it.
Also, if you can, please add this cache setting on the API configuration to allow customizing from the config.yml file without need a deployment, something like a cache_ttl for example.

Psycojoker:
Done.

I've completely changed the code because it didn't made that much sens to make it that complicated in the end and I've also made all cache TTL of this file configurable.

nesitor (approved):

PR #692, opened 2025-01-29
Fix: Ignore duplicated messages of already forgotten messages

amalcaraz:

- [x] Are my files well typed
- [x] Database migrations file are included
- [x] Are there enough tests

## Changes

- There is a race condition scenario in which regular message confirmations are processed after a FORGET message referencing them has been processed, resulting in inconsistent final states (wrong aggregates, posts amends, and vm executions). As a fix we are checking if the message has already  been forgotten to skip the process method in that cases.

- dev and test environment docker configurations have been also added

## Notes

We still need to migrate (delete duplicates) from aggregate_elements, aggregates, file_tags and posts tables.
The issue that was causing this inconsistent state has been fixed and we have considered that it doesn't worth it to clean
this tables for now as there are less than 1k orphan rows

### DIFF
--- a/src/aleph/handlers/message_handler.py
+++ b/src/aleph/handlers/message_handler.py
@@ -418,7 +418,7 @@ class MessageHandler(BaseMessageHandler):
             )
             return RejectedMessage(
                 pending_message=pending_message,
-                error_code=ErrorCode.FORGET_TARGET_NOT_FOUND,
+                error_code=ErrorCode.FORGOTTEN_DUPLICATE,
             )
 
         message = await self.verify_and_fetch(
--- a/src/aleph/handlers/message_handler.py
+++ b/src/aleph/handlers/message_handler.py
@@ -13,6 +13,7 @@ from sqlalchemy.dialects.postgresql import insert
 from aleph.chains.signature_verifier import SignatureVerifier
 from aleph.db.accessors.files import insert_content_file_pin, upsert_file
 from aleph.db.accessors.messages import (
+    get_forgotten_message,
     get_message_by_item_hash,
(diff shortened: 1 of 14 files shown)


src/aleph/handlers/message_handler.py
```diff
@@ -407,6 +405,22 @@ async def process(
             )
             return ProcessedMessage(message=existing_message, is_confirmation=True)
 
+        # Note: Check if message is already forgotten (and confirm it)
+        # this is to avoid race conditions when a confirmation arrives after the FORGET message has been preocessed
```
nesitor, line 409:
```suggestion
        # this is to avoid race conditions when a confirmation arrives after the FORGET message has been processed
```

src/aleph/handlers/message_handler.py
```diff
@@ -407,6 +405,22 @@ async def process(
...
+        )
+        if forgotten_message:
+            await self.confirm_existing_forgotten_message(
+                session=session,
+                forgotten_message=forgotten_message,
+                pending_message=pending_message,
+            )
+            return RejectedMessage(
+                pending_message=pending_message,
+                error_code=ErrorCode.FORGET_TARGET_NOT_FOUND,
```
nesitor, line 421: I think that we need to create a specific error code for that situation.

nesitor (approved):

src/aleph/handlers/message_handler.py
amalcaraz: I have created a new code called `FORGET_FORGOTTEN` for this scenario

nesitor (approved):

PR #693, opened 2025-01-31
feat: cache aggregates by owner results until they are updated

Psycojoker:
Optimize aggregates API route using caching

Related Clickup or Jira tickets : ALEPH-363

- [x] Are my files well typed
- [x] New translations have been added or updated if new strings have been introduced in the frontend
- [x] Database migrations file are included
- [x] Are there enough tests
- [x] Documentation has been included (for new feature)

## Changes

Introduce a small in memory cache based on defaultdict because aiocache requires async and most db related methods aren't async so we can't use it.

We cache on get_aggregates_by_owner call and invalidate the cache on PostgreSQL `after_update` and `after_delete` events to ensure we don't miss any update/delete.

## How to test

Running the tests should be enough, otherwise you just hit the aggregates API endpoint several time to raise the cache and wait for an aggregate to be updated to see if it works.

### DIFF
--- a/src/aleph/cache.py
+++ b/src/aleph/cache.py
@@ -1,0 +1,28 @@
+from collections import defaultdict
+
+
+class Cache:
+    def __init__(self):
+        self._cache = defaultdict(dict)
+
+    def get(self, key, namespace):
+        return self._cache[namespace].get(key)
+
+    def set(self, key, value, namespace):
+        self._cache[namespace][key] = value
+
+    def exists(self, key, namespace):
+        return key in self._cache[namespace]
+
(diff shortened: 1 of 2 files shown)


Psycojoker:
Right now this is using ORM events, it should be good but I'm not 100 sure, maybe we need to switch to PostgreSQL event triggers instead to ensure that event if another process works on an aggregate the cache will be invalidated?

nesitor (approved):

Psycojoker:
We'll need to run this one on testing for some time to ensure that it works as expected.

PR #694, opened 2025-01-31
Feature: Cost service now uses pricing aggregate

amalcaraz:

- [x] Are my files well typed
- [x] Are there enough tests

## Changes

Hardcoded magic numbers for calculating prices has been replaced by an official product pricing aggregate created by a logic and verified account in the network.

The cost calculator service has been refactored to get the prices form the aggregate instead of using hardcoded values.

https://api3.aleph.im/api/v0/aggregates/0xA07B1214bAe0D5ccAA25449C3149c0aC83658874.json?keys=pricing

## How to test

You can use the `/api/v0/price/{hash}` for getting the price of a vm or instance and it should return the correct price as it was before the change

## Notes

The owner of the aggregate may change before the next release


nesitor (approved):

pyproject.toml
```diff
@@ -202,6 +202,7 @@ git-only = [
   ".coveragerc",
   ".dockerignore",
   "shell.nix",
+  "mypy.ini",
```
Psycojoker, line 205: fyi mypy configuration is already in the pyproject.toml file (see line 258), the syntax is just a little bit different https://mypy.readthedocs.io/en/stable/config_file.html

PR #703, opened 2025-02-17
Estimate costs update

amalcaraz:

- [x] Are my files well typed
- [x] Are there enough tests

## Changes

The price estimation endpoint now uses a different Message schema, less restrictive and accepting `estimated_size_mib` props passed from the client to calculate costs without uploading files up front.

There is also a tiny refactor on the cost service, and a performance improvement on storage controller checking the account balance before uploading the volume

### DIFF
--- a/src/aleph/schemas/base_messages.py
+++ b/src/aleph/schemas/base_messages.py
@@ -6,15 +6,69 @@ import datetime as dt
 from hashlib import sha256
 from typing import Any, Generic, Mapping, Optional, TypeVar, cast
 
-from aleph_message.models import BaseContent, Chain, ItemType, MessageType
 from pydantic import root_validator, validator
 from pydantic.generics import GenericModel
 
-from aleph.toolkit.timestamp import timestamp_to_datetime
+from aleph_message.models import BaseContent, Chain, ItemType, MessageType
 from aleph.utils import item_type_from_hash
 
 MType = TypeVar("MType", bound=MessageType)
 ContentType = TypeVar("ContentType", bound=BaseContent)
+
+
+def base_message_validator_check_item_type(values):
+    """
+    Checks that the item hash of the message matches the one inferred from the hash.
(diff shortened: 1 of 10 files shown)


src/aleph/schemas/base_messages.py
```diff
@@ -10,95 +10,89 @@
...
-    chain: Chain
-    signature: Optional[str]
-    type: MType
-    item_content: Optional[str]
+    sender: Optional[str] = None
+    chain: Optional[Chain] = None
+    signature: Optional[str] = None
+    type: Optional[MType] = None
+    item_content: Optional[str] = None
     item_type: ItemType
```
1yam, line 86: Why is this becoming optional ?
Shouldn't these fields be required in every Message ?

src/aleph/schemas/cost_estimation_messages.py
```diff
@@ -0,0 +1,151 @@
...
+    volumes: List[CostEstimationMachineVolume] = Field(
+        default=[], description="Volumes to mount on the filesystem"
+    )
+
+
+class CostEstimationStoreContent(StoreContent):
+    estimated_size_mib: Optional[int] = None
+
+
+# ----
```
1yam, line 50:
```suggestion
```

src/aleph/services/cost.py
```diff
@@ -450,6 +457,26 @@ def _calculate_executable_costs(
...
+
+    if not pricing.price.compute_unit:
+        raise ValueError(
+            "compute_unit not defined for type '{}' in pricing aggregate".format(
+                pricing.type.value
+            )
+        )
+
+    compute_unit_cost = pricing.price.compute_unit.holding
+    compute_unit_cost_second = pricing.price.compute_unit.payg / HOUR
```
nesitor, line 478: This seems to be duplicated code from the above one

nesitor (approved):

1yam (approved):
LGTM

PR #708, opened 2025-02-20
fix docker compose dev allow git to read tags for pyaleph to be able to start

Psycojoker:
- **pyaleph does not provide the lint subdependencies category**
- **fix(docker-compose-dev): allow git to read tags for pyaleph to be able to start"**


Psycojoker:
@amalcaraz hey,

Do you have a feedback on this PR? This is the dockerfile you've wrote and idk if you really wanted the user root to run the programs or if the user aleph, like in our suggested docker-compose, is good enough?

PR #711, opened 2025-02-21
ALEPH-408 Storage on a CCN should be an option

Psycojoker:
Related Clickup or Jira tickets : [ALEPH-408](https://aleph-im.atlassian.net/browse/ALEPH-408)

- [x] Are my files well typed
- [x] Are there enough tests

## Changes

I'm not 100% sure about this approach but I've changed the behavior from doing
nothing to pretending we are storing a file in the database with a size of -1
while not doing the standard lookup nor talking to IPFS.

I'm wondering if we still shouldn't talk to IPFS but prevent it from pinning
the file?  It's unclear to me if we actually want to pin the file here.

## How to test

The tests should take care of it but I haven't been able to find how to
simulate a new message asking to store a file yet.

### DIFF
--- a/src/aleph/handlers/content/store.py
+++ b/src/aleph/handlers/content/store.py
@@ -107,9 +107,6 @@ class StoreMessageHandler(ContentHandler):
         # This check is essential to ensure that files are not added to the system
         # or the current node when the configuration disables storing of files.
         config = get_config()
-        if not config.storage.store_files.value:
-            return  # Ignore if files are not to be stored.
-
         content = message.parsed_content
         assert isinstance(content, StoreContent)
 
@@ -122,6 +119,7 @@ class StoreMessageHandler(ContentHandler):
         do_standard_lookup = True
 
         # Sentinel value, the code below always sets a value but mypy does not see it.
+        # otherwise if config.storage.store_files is False, this will be the database value
         size: int = -1
 
         if engine == ItemType.ipfs and ipfs_enabled:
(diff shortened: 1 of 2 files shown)


src/aleph/handlers/content/store.py
```diff
@@ -122,9 +119,14 @@ async def fetch_related_content(
         do_standard_lookup = True
 
         # Sentinel value, the code below always sets a value but mypy does not see it.
+        # otherwise if config.storage.store_files is False, this will be the database value
         size: int = -1
 
-        if engine == ItemType.ipfs and ipfs_enabled:
+        if (
+            config.storage.store_files.value
```
hoh, line 126: There are 2 types of file storage: IPFS and local on the node. My understanding is that `config.storage.store_files.value` is for the local storage and `config.ipfs.enabled.value -> ipfs_enabled` for IPFS storage.

Is it therefore justified to diable IPFS storage when `config.storage.store_files.value == False` and `config.ipfs.enabled.value == True` ?

src/aleph/handlers/content/store.py
```diff
@@ -175,7 +177,7 @@ async def fetch_related_content(
                 )
                 do_standard_lookup = True
 
-        if do_standard_lookup:
+        if config.storage.store_files.value and do_standard_lookup:
```
hoh, line 180: Is it relevant to still register the file in the database with `upsert_file` in this case or should that be omitted as well ? :thinking:

src/aleph/handlers/content/store.py
Psycojoker: Probably not yeah, I'll remove that.

src/aleph/handlers/content/store.py
Psycojoker: The instructions that I got is that the current behavior doesn't mark the message as acknowledge which we don't want so I believe that registering this in the db would do it?

But the code is really unclear and not documented so I'm quite confuse on how to achieve that :/

hoh (approved):

PR #757, opened 2025-03-17
Feat: store and program messages are not free anymore

amalcaraz:
Now any user can pin or upload STORE files on Aleph network without having tokens hold, and this thing also applies on programs, as anyone can create a program without having tokens hold.

We need to enable the balance check on Pyaleph on creation moment for programs, and add the check on the garbage collector and on the pin/upload action for STORE messages. 

This should only apply for new PROGRAM or STORE messages, so we can use the release block height or date to take into account for that check.

- [x] Are my files well typed
- [x] Are there enough tests

## Changes

- Store and ephemeral program messages now take into account the free and locked account token balance before being processed

## How to test

- Try to upload/pin a volume with a low/empty balance account and check that an error is raised
- Try to set up an ephemeral program with a low/empty balance account and check that an error is raised

## Notes

- `STORE_AND_PROGRAM_COST_DEADLINE_HEIGHT` and `STORE_AND_PROGRAM_COST_DEADLINE_TIMESTAMP` variables are open to be defined with a different value depending on the release date


src/aleph/web/controllers/storage.py
```diff
@@ -111,12 +111,12 @@ async def _verify_user_balance(
...
             session=session, address=content.address
         )
-        required_storage_cost, _ = get_total_and_detailed_costs(session, content, "")
+        message_cost, _ = get_total_and_detailed_costs(session, content, "")
 
-        required_balance = current_cost_for_user + required_storage_cost
+        required_balance = current_cost + message_cost
 
         if current_balance < required_balance:
             raise web.HTTPPaymentRequired()
```
nesitor, line 122: I think that we also need to avoid this 25MB limit of individual file

PR #792, opened 2025-05-15
Garbage collection failure proof

nesitor:
Fix: Solve garbage collection deleting issues that raises individually and stops the garbage process.

- [X] Are my files well typed
- [X] Database migrations file are included
- [X] Are there enough tests

## Changes

- Added a try catch for every file to avoid aborting the entire garbage collection process only caused by one file error.

## How to test

Execute pyaleph and check that garbage collection runs smoothly and well.


amalcaraz (approved):
LGTM

PR #799, opened 2025-05-29
Added balance pre-check on file pins

nesitor:
This PR solves the issue when a user tries to pin a big file on IPFS without having a balance.

- [X] Are my files well typed
- [X] New translations have been added or updated if new strings have been introduced in the frontend
- [X] Database migrations file are included
- [X] Are there enough tests
- [X] Documentation has been included (for new feature)

## Changes

Added a pre-check balance checking the IPFS file without having to download it. It will work only on unifs type pins, but will save us a lot of useless disk usage.

## How to test

Try to pin a big file on the network with an empty wallet and check that file is not saved either on the database and the message is rejected so faster.

### DIFF
--- a/src/aleph/services/cost.py
+++ b/src/aleph/services/cost.py
@@ -344,22 +344,21 @@ def _get_execution_volumes_costs(
                 isinstance(content, CostEstimationProgramContent)
                 and content.data.estimated_size_mib
             ):
-                if content.data.estimated_size_mib:
-                    volumes.append(
-                        SizedVolume(
-                            CostType.EXECUTION_PROGRAM_VOLUME_DATA,
-                            Decimal(content.data.estimated_size_mib),
-                            content.data.ref,
-                        )
+                volumes.append(
+                    SizedVolume(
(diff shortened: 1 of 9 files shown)


src/aleph/services/cost.py
```diff
@@ -344,28 +344,29 @@ def _get_execution_volumes_costs(
                 isinstance(content, CostEstimationProgramContent)
                 and content.data.estimated_size_mib
             ):
-                volumes.append(
-                    SizedVolume(
-                        CostType.EXECUTION_PROGRAM_VOLUME_DATA,
-                        Decimal(content.data.estimated_size_mib),
-                        content.data.ref,
+                if content.data.estimated_size_mib:
```
amalcaraz, line 347: this `if` is not necessary and can cause bad cost calculation for `ProgramContent` instances

amalcaraz (changes requested):
I think there could be a problem calculating real costs for `ProgramContent` DATA volume

PR #820, opened 2025-07-08
Fix: remove I/O Blocking code on message processing parts

1yam:
This PR remove I/O blocking operations on parts used by message processing to bring back message parallelization

Related Clickup or Jira tickets : ALEPH-526

- [x] Are my files well typed
- [x] Are there enough tests

## Changes
This pull request introduces significant changes to improve asynchronous handling across multiple modules, enhance efficiency, and ensure non-blocking operations. Key updates include replacing synchronous file and I/O operations with asynchronous alternatives, introducing semaphores to limit concurrent requests, and refactoring code to use shared `aiohttp.ClientSession` instances for HTTP requests.

### Asynchronous Enhancements:
* [`src/aleph/handlers/content/aggregate.py`](diffhunk://#diff-fdcc8dbfb6b9f02a3edf4722670b7d71418e42a1cafa283d6b9829bd8c6c7131L90-R92): Updated `_prepend_to_aggregate` and `_update_aggregate` to use `asyncio.to_thread` for non-blocking execution of `merge_aggregate_elements`. [[1]](diffhunk://#diff-fdcc8dbfb6b9f02a3edf4722670b7d71418e42a1cafa283d6b9829bd8c6c7131L90-R92) [[2]](diffhunk://#diff-fdcc8dbfb6b9f02a3edf4722670b7d71418e42a1cafa283d6b9829bd8c6c7131L133-R134)

* [`src/aleph/services/storage/fileystem_engine.py`](diffhunk://#diff-26a3c1e17d71e9d0aa34cae935c4ce244e31380a3d64408e8e7f4e32b6b419daL19-R44): Replaced synchronous file operations with asynchronous equivalents using `aiofiles`, including `read`, `write`, `delete`, and `exists` methods.

### HTTP Request Optimization:
* [`src/aleph/services/p2p/http.py`](diffhunk://#diff-2350fd5801f1cbda69cf07db21c8fe367b9ab069bbc6bdfd2daf27f14b7eb254L15-R22): Refactored `api_get_request` to use a shared `aiohttp.ClientSession` instead of maintaining separate session instances, improving resource management. Added semaphores to limit concurrent calls in `get_peer_hash_content` and `request_hash`. [[1]](diffhunk://#diff-2350fd5801f1cbda69cf07db21c8fe367b9ab069bbc6bdfd2daf27f14b7eb254L15-R22) [[2]](diffhunk://#diff-2350fd5801f1cbda69cf07db21c8fe367b9ab069bbc6bdfd2daf27f14b7eb254L47-L56) [[3]](diffhunk://#diff-2350fd5801f1cbda69cf07db21c8fe367b9ab069bbc6bdfd2daf27f14b7eb254L65-R103)

* [`src/aleph/services/p2p/jobs.py`](diffhunk://#diff-575c251d6b8c284c71309da3a25bc1ffc613fbcc29acfc8be5f0456d7d49ed2fR75-R91): Introduced `request_version` to handle peer version checks using `aiohttp.ClientSession` and asynchronous tasks. Refactored `tidy_http_peers_job` to utilize `request_version` for better scalability. [[1]](diffhunk://#diff-575c251d6b8c284c71309da3a25bc1ffc613fbcc29acfc8be5f0456d7d49ed2fR75-R91) [[2]](diffhunk://#diff-575c251d6b8c284c71309da3a25bc1ffc613fbcc29acfc8be5f0456d7d49ed2fL80-R109)

### Code Cleanup:
* [`src/aleph/handlers/content/post.py`](diffhunk://#diff-2c6d5e01065f96f5fd5131f7cd49c43fe34aad31e0ffc70506053c36b5d2061aL52-L57): Removed debug print statements from `update_balances` for cleaner logging.

These changes collectively enhance the system's scalability, responsiveness, and maintainability by leveraging asynchronous programming patterns effectively.

## How to test
This PR doesn't add any breaking change, it's just running things in Non I/O blockign way

### DIFF
--- a/src/aleph/db/accessors/balances.py
+++ b/src/aleph/db/accessors/balances.py
@@ -154,7 +154,6 @@ async def update_balances(
     )
 
     # Convert floats to str to avoid having issue with float to decimal conversion
-
     records = [
         (address, chain.value, dapp or "", str(balance), eth_height, last_update)
         for address, balance in balances.items()
--- a/src/aleph/handlers/content/aggregate.py
+++ b/src/aleph/handlers/content/aggregate.py
@@ -1,3 +1,4 @@
+import asyncio
 import itertools
 import logging
 from typing import List, Sequence, Set, cast
@@ -87,8 +88,8 @@ class AggregateMessageHandler(ContentHandler):
         aggregate: AggregateDb,
         elements: Sequence[AggregateElementDb],
     ):
-        # That will need to run in no IO blocking way
-        new_content = merge_aggregate_elements(elements)
+        # We run in to thread to avoid IO blocking if huge aggregate pass
+        new_content = await asyncio.to_thread(merge_aggregate_elements, elements)
 
         await update_aggregate(
             session=session,
@@ -130,7 +131,7 @@ class AggregateMessageHandler(ContentHandler):
         if not aggregate_metadata:
             LOGGER.info("%s/%s does not exist, creating it", key, owner)
 
-            content = merge_aggregate_elements(elements)
+            content = await asyncio.to_thread(merge_aggregate_elements, elements)
             await insert_aggregate(
                 session=session,
                 key=key,
@@ -239,5 +240,4 @@ class AggregateMessageHandler(ContentHandler):
 
         LOGGER.debug("Refreshing aggregate %s/%s...", owner, key)
         await refresh_aggregate(session=session, owner=owner, key=str(key))
-
         return set()
--- a/src/aleph/handlers/content/post.py
+++ b/src/aleph/handlers/content/post.py
@@ -49,12 +49,8 @@ async def update_balances(session: AsyncDbSession, content: Mapping[str, Any]) -
     dapp = content.get("dapp")
 
     LOGGER.info("Updating balances for %s (dapp: %s)", chain, dapp)
-    print(f"Chain type: {type(chain)}, Chain value: {chain.value}, Full chain: {chain}")
 
     balances: Dict[str, float] = content["balances"]
-    print(f"Number of balances to update: {len(balances)}")
-    for addr, bal in list(balances.items())[:3]:  # Print first 3 for debug
-        print(f"  {addr}: {bal} bal type {type(bal)}")
 
     await update_balances_db(
         session=session,
--- a/src/aleph/services/p2p/http.py
+++ b/src/aleph/services/p2p/http.py
@@ -66,40 +66,65 @@ async def get_peer_hash_content(
 async def request_hash(
     api_servers: Sequence[str], item_hash: str, timeout: int = 1
 ) -> Optional[bytes]:
+    """
+    Request a hash from available API servers Concurrently over the network.
+    We take the first valid respond and close other task
+    """
     uris: List[str] = sample(api_servers, k=len(api_servers))
-
-    # Avoid too  much request at the same time
     semaphore = asyncio.Semaphore(5)
-
     connector = aiohttp.TCPConnector(limit_per_host=5)
     timeout_conf = aiohttp.ClientTimeout(total=timeout)
 
-    async with aiohttp.ClientSession(
-        connector=connector, timeout=timeout_conf
-    ) as session:
-        # Use Task instead of
-        tasks = [
+    # Use a dedicated session for each get_peer_hash_content call
+    tasks = []
+    for url in uris:
+        session = aiohttp.ClientSession(connector=connector, timeout=timeout_conf)
+        tasks.append(
             asyncio.create_task(
-                get_peer_hash_content(
+                get_peer_hash_content_with_session(
                     session=session,
                     base_uri=url,
                     item_hash=item_hash,
                     semaphore=semaphore,
                 )
             )
-            for url in uris
-        ]
+        )
 
-        for completed_task in asyncio.as_completed(tasks):
-            try:
-                result = await completed_task
-                if result:
-                    # We cancel other Task
-                    for task in tasks:
-                        if not task.done():
-                            task.cancel()
-                    return result
-            except Exception:
-                continue
+    done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
+    result = None
+    for t in done:
+        try:
+            data = t.result()
+        except Exception:
+            continue
+        if data:
+            result = data
+            break
 
-    return None  # Nothing found...
+    for p in pending:
+        p.cancel()
+    await asyncio.gather(*pending, return_exceptions=True)
+
+    return result
+
+
+async def get_peer_hash_content_with_session(
+    session: aiohttp.ClientSession,
+    base_uri: str,
+    item_hash: str,
+    semaphore,
+    timeout: int = 1,
+) -> Optional[bytes]:
+    """Wrapper to ensure session is properly closed"""
+    try:
+        async with session:
+            return await get_peer_hash_content(
+                session=session,
+                base_uri=base_uri,
+                item_hash=item_hash,
+                semaphore=semaphore,
+                timeout=timeout,
+            )
+    except Exception as e:
+        LOGGER.exception(f"Error in get_peer_hash_content_with_session: {e}")
+        return None
--- a/src/aleph/services/p2p/http.py
+++ b/src/aleph/services/p2p/http.py
@@ -12,23 +12,14 @@ import aiohttp
 
 LOGGER = logging.getLogger("P2P.HTTP")
 
-SESSIONS = dict()
 
-
-async def api_get_request(base_uri, method, timeout=1):
-    if timeout not in SESSIONS:
-        connector = aiohttp.TCPConnector(limit_per_host=5)
-        SESSIONS[timeout] = aiohttp.ClientSession(
-            read_timeout=timeout, connector=connector
-        )
-
+async def api_get_request(session: aiohttp.ClientSession, base_uri, method, timeout=1):
     uri = f"{base_uri}/api/v0/{method}"
     try:
-        async with SESSIONS[timeout].get(uri) as resp:
+        async with session.get(uri) as resp:
             if resp.status != 200:
-                result = None
-            else:
-                result = await resp.json()
+                return None
+            return await resp.json()
     except (
         TimeoutError,
         asyncio.TimeoutError,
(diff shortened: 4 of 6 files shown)


foxpatch-aleph (changes requested):
The PR introduces several genuine bugs alongside the legitimate async improvements. The most critical issue is in `request_hash`: multiple `aiohttp.ClientSession` instances share a single `TCPConnector` without setting `connector_owner=False`, so the first session to close will close the connector and break all remaining sessions. Additionally, using `asyncio.wait(return_when=FIRST_COMPLETED)` and then cancelling all pending tasks means that if the fastest server returns `None` (failure), the function gives up on potentially successful slower servers — a correctness regression from the original sequential approach. There is also an inconsistency in aggregate.py where `_append_to_aggregate` still calls `merge_aggregate_elements` synchronously on the event loop while `_prepend_to_aggregate` and `_update_aggregate` have been moved to `asyncio.to_thread`; the justification for threading is weak anyway since `merge_aggregate_elements` is a trivial in-memory dict merge with no I/O. The `aiofiles.ospath.wrap` usage for `Path.unlink` is fragile (internal API), and the `timeout` parameter in `api_get_request` is now dead code. The filesystem and jobs.py changes are directionally correct, but these bugs need addressing before merge.

**`src/aleph/services/p2p/http.py`** (line 81): **Bug: shared connector closed by the first session to exit.**

A single `TCPConnector` is created and then passed to multiple `ClientSession` instances. When `get_peer_hash_content_with_session` does `async with session:`, the session's `close()` also closes the connector (it is the `connector_owner` by default). The first task to finish will close the connector, causing `ConnectorError` or `RuntimeError` for all remaining sessions.

Fix: either create one shared session and share it across all tasks (closing it after `asyncio.gather`), or pass `connector_owner=False` to every `ClientSession` and close the connector explicitly afterwards.

```python
# Option A – single shared session
connector = aiohttp.TCPConnector(limit_per_host=5)
timeout_conf = aiohttp.ClientTimeout(total=timeout)
async with aiohttp.ClientSession(connector=connector, timeout=timeout_conf) as session:
    tasks = [
        asyncio.create_task(get_peer_hash_content(session, url, item_hash, semaphore))
        for url in uris
    ]
    ...
```

**`src/aleph/services/p2p/http.py`** (line 93): **Bug: correctness regression — valid responses from slower peers are silently discarded.**

`asyncio.wait(return_when=FIRST_COMPLETED)` returns as soon as *any* task finishes. If that first task returns `None` (timeout, HTTP error, etc.), the loop exhausts `done` without finding valid data, then cancels all `pending` tasks — even though they might succeed. The old code tried every server sequentially and only gave up once all were exhausted.

To preserve the original semantics while still being concurrent, either:
- Use `return_when=ALL_COMPLETED` and pick the first non-`None` result, or
- Loop with `return_when=FIRST_COMPLETED` in a `while tasks:` loop, continuing only when `result` is still `None`.

```python
while tasks:
    done, tasks = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
    for t in done:
        try:
            data = t.result()
        except Exception:
            continue
        if data:
            for p in tasks:
                p.cancel()
            await asyncio.gather(*tasks, return_exceptions=True)
            return data
return None
```

**`src/aleph/services/p2p/http.py`** (line 16): **Dead parameter: `timeout` is no longer used.**

The `timeout` argument was previously used to key into `SESSIONS`. Now that sessions are passed in by callers, `timeout` does nothing in this function and is misleading. Remove it from the signature.

**`src/aleph/handlers/content/aggregate.py`** (line 74): **Inconsistency: `_append_to_aggregate` still calls `merge_aggregate_elements` on the event loop.**

`_prepend_to_aggregate` (line 92) and `_update_aggregate` (line 134) have been moved to `asyncio.to_thread`, but `_append_to_aggregate` at line 74 still calls the function synchronously. Since `_append_to_aggregate` is the *best-case* (most common) code path, this inconsistency means the supposed non-blocking fix barely applies in practice.

Either apply the same treatment to `_append_to_aggregate`, or reconsider whether the threading is warranted at all (see next comment).

**`src/aleph/handlers/content/aggregate.py`** (line 92): **Questionable use of `asyncio.to_thread` for a trivial in-memory operation.**

`merge_aggregate_elements` is a pure CPU dict-merge loop with no I/O whatsoever:
```python
def merge_aggregate_elements(elements):
    content = {}
    for element in elements:
        content.update(element.content)
    return content
```
For typical aggregate sizes this runs in microseconds. Offloading it to a thread pool adds scheduling overhead and — more importantly — passes SQLAlchemy ORM objects across thread boundaries, which can be unsafe if lazy-load attributes are accessed in the worker thread.

If this operation genuinely needs to be non-blocking for very large aggregates, the comment should document the threshold/rationale, and the SQLAlchemy objects should be converted to plain dicts before being handed to the thread.

**`src/aleph/services/storage/fileystem_engine.py`** (line 34): **Fragile use of `aiofiles.ospath.wrap` (internal API).**

`aiofiles.ospath.wrap` is an internal helper, not part of the public aiofiles API, and its signature or behaviour could change without notice. The equivalent using the public API is:

```python
async def delete(self, filename: str):
    file_path = self.folder / filename
    await asyncio.to_thread(file_path.unlink, missing_ok=True)
```

This is clearer, avoids internal APIs, and the intent is immediately obvious.

**`src/aleph/services/p2p/jobs.py`** (line 75): **Missing return type annotation on `request_version`.**

All other functions in this module are annotated. `request_version` should be:
```python
async def request_version(peers: Sequence[str], my_ip: str, timeout: int = 1) -> Sequence[PeerStatus]:
```

**`src/aleph/services/p2p/http.py`** (line 41): **Untyped `semaphore` parameter.**

Both `get_peer_hash_content` and `get_peer_hash_content_with_session` receive `semaphore` typed as plain `Any`. Use `asyncio.Semaphore` for clarity and IDE support.

foxpatch-aleph (changes requested):
The PR makes good progress on removing I/O-blocking operations, but introduces two correctness bugs in `request_hash` that would cause failures in production. First, a single `TCPConnector` is shared across multiple `ClientSession` instances, each of which owns and closes the connector on exit — so the second session to close will see a already-closed connector and raise. Second, `asyncio.wait(FIRST_COMPLETED)` cancels all remaining tasks the moment any task finishes, even when that task returned `None` (server didn't have the hash), meaning the function can return `None` when other servers still in flight would have succeeded. These two issues together make the concurrent hash-fetching logic less reliable than the original sequential code. There are also smaller issues: `timeout` in `api_get_request` is now a dead parameter, `_append_to_aggregate` was missed in the `asyncio.to_thread` refactor, and `aiofiles.ospath.wrap` uses a semi-internal API.

**`src/aleph/services/p2p/http.py`** (line 75): **Bug: shared connector across multiple sessions will cause failures.**

A single `TCPConnector` is created and then passed to N different `ClientSession` objects. `aiohttp.ClientSession` defaults to `connector_owner=True`, so the first session that closes (via `async with session:` in `get_peer_hash_content_with_session`) will close and finalize the connector. Every subsequent session that tries to use the now-closed connector will raise `RuntimeError: Session is closed` or a connector-closed error.

Options:
- Create one shared session for all tasks (pass it in from the caller or create it in `request_hash` and pass it through).
- Or pass `connector_owner=False` to each `ClientSession` and close the connector explicitly after all tasks are done.

The simplest fix is to lift session creation out of the loop and share a single session, since the connector's `limit_per_host=5` already limits concurrency.

**`src/aleph/services/p2p/http.py`** (line 93): **Bug: `FIRST_COMPLETED` cancels pending tasks even when the first result is `None`.**

`asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)` returns as soon as *any* task finishes — including tasks that return `None` (server didn't have the hash) or raise an exception. The loop over `done` will find no valid data, `result` stays `None`, and then all pending tasks are cancelled. This means we can return `None` even when servers still in flight would have returned the content.

The old sequential code always tried all servers. To preserve that guarantee while still being concurrent, consider using `asyncio.FIRST_EXCEPTION` with a loop, or collecting all results with `asyncio.gather` and picking the first non-None, or implementing a manual approach where tasks notify a queue when they produce a valid result.

**`src/aleph/services/p2p/http.py`** (line 16): The `timeout` parameter of `api_get_request` is now dead code — it is accepted but never used. The actual timeout is configured on the `ClientSession` passed in by the caller. Remove the parameter (or at minimum remove the default value and add a note) to avoid misleading callers into thinking they can control the timeout here.

**`src/aleph/handlers/content/aggregate.py`** (line 74): `_append_to_aggregate` (line 74) still calls `merge_aggregate_elements` synchronously, while `_prepend_to_aggregate` and `_update_aggregate` were updated to use `asyncio.to_thread`. All three paths call the same function on potentially large element sets, so they should be consistent. Either apply `asyncio.to_thread` here too, or reconsider whether it's needed at all — `merge_aggregate_elements` is a simple `dict.update` loop with no I/O, so it is not blocking in the usual sense.

**`src/aleph/services/storage/fileystem_engine.py`** (line 34): `aiofiles.ospath.wrap` is not part of `aiofiles`'s documented public API and is not guaranteed to be stable across versions. Prefer the stdlib-idiomatic `asyncio.to_thread(file_path.unlink, missing_ok=True)` which achieves the same result (run the blocking call in a thread pool) without relying on internal `aiofiles` helpers.

**`src/aleph/services/p2p/http.py`** (line 93): Minor: if `api_servers` is empty, `tasks` will be empty and `asyncio.wait([])` raises `ValueError`. Add an early-return guard (`if not tasks: return None`) before the `asyncio.wait` call. This is a pre-existing edge case surfaced by the refactor.

foxpatch-aleph (changes requested):
The PR successfully converts synchronous I/O to async equivalents and removes debug prints, which are solid improvements. However, there are two genuine bugs in `request_hash`: (1) multiple `aiohttp.ClientSession` instances share a single `TCPConnector` without `connector_owner=False`, so the first session to close will destroy the connector and cause all remaining sessions to fail; (2) the `asyncio.FIRST_COMPLETED` logic does not retry when the first-to-finish task returns `None` — it immediately cancels all other tasks and returns `None`, which is a regression from the prior sequential strategy that tried all servers. There are also several minor issues worth addressing.

**`src/aleph/services/p2p/http.py`** (line 81): **Bug: shared connector ownership.** A single `connector` is created on line 75 and then passed into a separate `aiohttp.ClientSession` for each URL. By default `connector_owner=True`, so when the first session exits its `async with session:` block in `get_peer_hash_content_with_session`, it will close the shared connector. Every other in-flight session then loses its connection pool and will raise `RuntimeError: Session is closed` or `connector: Cannot connect to host`. Either (a) create a dedicated connector per session, (b) pass `connector_owner=False` to every session and close the connector explicitly after all tasks finish, or (c) use one shared session for all tasks (like `request_version` does).

**`src/aleph/services/p2p/http.py`** (line 93): **Bug: early bail-out on `None` result causes regression.** `asyncio.wait(..., return_when=asyncio.FIRST_COMPLETED)` returns as soon as *at least one* task finishes. If that task returned `None` (the peer doesn't have the content), the loop over `done` finds no truthy `data`, so `result` stays `None` — and then all pending tasks are cancelled. The previous sequential code kept trying every server until one succeeded. To preserve that behaviour the wait should loop (`while pending`) or use `return_when=asyncio.ALL_COMPLETED` with a first-non-None scan.

**`src/aleph/services/p2p/http.py`** (line 16): The `timeout=1` parameter on `api_get_request` is now dead code — it is never used inside the function body. Timeout is correctly governed by the `ClientSession` passed in. The parameter should be removed to avoid confusion; callers like `check_peer` (jobs.py:64) currently pass `timeout=timeout` which silently has no effect.

**`src/aleph/services/p2p/http.py`** (line 41): The `semaphore` parameter (here and in `get_peer_hash_content_with_session`) has no type annotation. Should be `asyncio.Semaphore` for clarity and static analysis.

**`src/aleph/services/storage/fileystem_engine.py`** (line 34): `aiofiles.ospath.wrap` is designed for `os.path`-style functions, not `pathlib.Path` instance methods. While it happens to work here (passing `file_path` as the implicit `self`), it is non-obvious and fragile. The idiomatic alternative is `await asyncio.to_thread(file_path.unlink, missing_ok=True)`, which is both clearer and officially supported.

**`src/aleph/handlers/content/aggregate.py`** (line 89): Comment says "to avoid IO blocking" but `asyncio.to_thread` offloads CPU-bound work to a thread pool — it does not help with async I/O. The accurate description is "to avoid blocking the event loop with CPU-bound work for large aggregates".

PR #827, opened 2025-07-23
Add Aleph Credits system with balance tracking and API endpoints

amalcaraz:
Summary

  - Implement Aleph Credits system to track user credit balances from token distributions
  - Add new database table credit_balances with migration to store credit transactions
  - Create API endpoints for querying credit balances at account and system levels
  - Integrate credit balance updates through POST messages from authorized addresses

  Changes

  - Database: New AlephCreditBalanceDb model and migration for credit balance tracking
  - API Endpoints: Added /accounts/{address}/credits and /credits/balances endpoints
  - Message Processing: Enhanced POST message handler to process credit distribution messages
  - Balance Management: Credit balance accessors with expiration date handling and bulk updates
  - Configuration: Added credit balance configuration with authorized addresses

  Test plan

  - Verify credit balance database operations (insert, query, expiration handling)
  - Test API endpoints for individual and paginated credit balance queries
  - Validate POST message processing for credit distributions
  - Confirm proper handling of expired credits in balance calculations
  - Test bulk credit balance updates and conflict resolution

### DIFF
--- a/src/aleph/db/models/balances.py
+++ b/src/aleph/db/models/balances.py
@@ -45,18 +45,20 @@ class AlephBalanceDb(Base):
 class AlephCreditBalanceDb(Base):
     __tablename__ = "credit_balances"
 
-    id: int = Column(BigInteger, primary_key=True)
+    id: int = Column(BigInteger, autoincrement=True)
 
     address: str = Column(String, nullable=False, index=True)
     amount: Decimal = Column(DECIMAL, nullable=False)
-    ratio: Decimal = Column(DECIMAL, nullable=False)
-    tx_hash: str = Column(String, nullable=False)
-    token: str = Column(String, nullable=False)
-    chain: str = Column(String, nullable=False)
-    provider: str = Column(String, nullable=False)
+    ratio: Optional[Decimal] = Column(DECIMAL, nullable=True)
+    tx_hash: Optional[str] = Column(String, nullable=True)
+    token: Optional[str] = Column(String, nullable=True)
+    chain: Optional[str] = Column(String, nullable=True)
+    provider: Optional[str] = Column(String, nullable=True)
     origin: Optional[str] = Column(String, nullable=True)
-    ref: Optional[str] = Column(String, nullable=True)
+    payment_ref: Optional[str] = Column(String, nullable=True)
     payment_method: Optional[str] = Column(String, nullable=True)
+    distribution_ref: str = Column(String, nullable=False, primary_key=True)
+    distribution_index: int = Column(Integer, nullable=False, primary_key=True)
     expiration_date: Optional[dt.datetime] = Column(
         TIMESTAMP(timezone=True), nullable=True
     )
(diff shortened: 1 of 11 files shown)


deployment/migrations/versions/0033_1c06d0ade60c_calculate_costs_statically.py
```diff
@@ -112,12 +70,15 @@ def do_calculate_costs() -> None:
 
     logger.debug("INIT: CALCULATE COSTS FOR: %r", msg_item_hashes)
 
+    # Build the initial pricing model from DEFAULT_PRICE_AGGREGATE
+    initial_pricing_model = build_default_pricing_model()
```
nesitor, line 74: Change a migration already executed it's not a good thing, as it can create differences with the nodes that have the previous migration.

src/aleph/db/models/balances.py
```diff
@@ -40,3 +40,33 @@ class AlephBalanceDb(Base):
...
+    )
+    last_update: dt.datetime = Column(
+        TIMESTAMP(timezone=True),
+        nullable=False,
+        server_default=func.now(),
+        onupdate=func.now(),
+    )
+
+    __table_args__ = (
+        UniqueConstraint("tx_hash", name="credit_balances_tx_hash_uindex"),
```
nesitor, line 71: This unique constraint is only valid for provider transactions, but not for "free" credit transactions that don't have transactions itself.

src/aleph/schemas/api/accounts.py
```diff
@@ -69,10 +74,31 @@ class AddressBalanceResponse(BaseModel):
...
+    min_balance: int = Field(
+        default=0, ge=1, description="Minimum Credit Balance needed"
+    )
+
+
+class AddressCreditBalanceResponse(BaseModel):
+    model_config = ConfigDict(from_attributes=True)
+
+    address: str
+    credits: FloatDecimal
```
nesitor, line 99: I think that we agreed to not allow decimals on credits.

amalcaraz: It is an open door just in case we change our mind on the while. For now we are always sending integer values on the credit distribution message, so there won't be any decimal value anywhere for credits

nesitor (dismissed):

nesitor (commented):
I think that we need to fix the constraint before merging it

src/aleph/schemas/api/accounts.py
```diff
@@ -69,10 +74,31 @@ class AddressBalanceResponse(BaseModel):
     model_config = ConfigDict(from_attributes=True)
 
     address: str
-    balance: str
+    balance: FloatDecimal
```
nesitor, line 77: Why changing this to float? We put as a string because the token balance can have more decimals than the allowed overflow for float values.

amalcaraz: This model is the response schema for the endpoint `https://api.aleph.im/api/v0/balances` which is not being use anywhere. It is currently broken in production nodes, becase the type casting from Decimal to str is not working.

I have fixed and aligned this schema to have the same format than `/api/v0/addresses/{address}/balance` which is  `FloatDecimal => float`
https://github.com/aleph-im/pyaleph/blob/main/src/aleph/schemas/api/accounts.py#L24

nesitor (approved):

amalcaraz:
This has been merged into:
https://github.com/aleph-im/pyaleph/pull/836

PR #919, opened 2026-01-08
feature: use EIP-1559 to optimize gas costs for Ethereum sync

odesenfans:
The packer was still using the legacy fee mechanism. While EIP-1559 yield similar results when there's no contention, EIP-1559 enables cheaper inclusion in case of contention.

We now also monitor the current gas price on-chain and wait if the current gas price is above target.

- [ ] Are my files well typed
- [ ] Are there enough tests

## Changes

## How to test

Explain how to test your PR.

## Print screen / video

## Notes

### DIFF
--- a/src/aleph/chains/ethereum.py
+++ b/src/aleph/chains/ethereum.py
@@ -2,7 +2,17 @@ import asyncio
 import importlib.resources
 import json
 import logging
-from typing import Any, AsyncIterator, Dict, List, Literal, Self, Tuple, Union
+from typing import (
+    Any,
+    AsyncIterator,
+    Dict,
+    List,
+    Literal,
+    Self,
+    Tuple,
+    TypedDict,
+    Union,
+)
 
 from aleph_message.models import Chain
 from configmanager import Config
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (approved):
This PR introduces EIP-1559 gas estimation for Ethereum transaction broadcasting, replacing the legacy `gasPrice`-only approach. The `_get_fee_estimates` method uses `eth.fee_history` with a conservative 25th-percentile priority fee averaged over 5 blocks, falling back to legacy gas pricing when fee history is unavailable. A documented web3 7.15.0 workaround strips a bogus `gasPrice` coroutine from EIP-1559 transactions. The gas price ceiling check in `broadcast_messages` correctly handles both EIP-1559 and legacy paths. The code is correct and well-structured; missing test coverage is noted but is acceptable for this stage of development.

**`src/aleph/chains/ethereum.py`** (line 367): The algorithm uses `fee_history(5, "latest", [25])` to get the 25th percentile priority fee per block, then averages those values. The docstring says "deriving the priority fee from the 25th percentile" which is accurate but could clarify that it's the average of per-block 25th percentile values, not the 25th percentile of a pooled sample.

**`src/aleph/chains/ethereum.py`** (line 380): Suggestion: the `2x` headroom factor comment says "leaves headroom for base-fee growth" — consider noting that this provides roughly 6 blocks (~1 min) of headroom given the 12.5% per-block max increase, to make the rationale self-documenting.

**`src/aleph/chains/ethereum.py`** (line 429): The `or` here is safe in practice (maxFeePerGas will never be 0 on a live network), but if `maxFeePerGas` were 0 it would incorrectly fall through to `gasPrice`. Consider using a more explicit sentinel pattern if this edge case matters.

PR #1093, opened 2026-04-10
Fix version detection picking stale v-prefixed tags

aliel:


foxpatch-aleph (changes requested):
The PR correctly fixes the version detection to filter out v-prefixed tags and improves the publish workflow with proper tagging strategy. However, there's a bug in the pre-release regex that only matches single-digit suffixes (e.g., rc9) but fails for multi-digit versions (e.g., rc10).

**`deployment/docker-build/publish.sh`** (line 55): The regex `^[0-9]+\.[0-9]+\.[0-9]+-(rc|a|alpha|beta)[0-9]+$` only matches single-digit pre-release suffixes. For versions like `0.10.1-rc10` or `0.5.3-beta12`, this would fail to match and fall through to the dev build case. Change `[0-9]+$` to `[0-9]+$` (add the `+` quantifier) to handle multi-digit suffixes.

**`deployment/docker-build/build.sh`** (line 20): Duplicate comment line. Line 19 says `# Sets IMAGE_TAG and PEP440_VERSION` and line 20 says `# Sets IMAGE, IMAGE_TAG, and PEP440_VERSION`. Remove line 19 since line 20 is the updated version.

foxpatch-aleph (changes requested):
The PR correctly filters v-prefixed tags in the exact-match case, but the fallback pattern `--match '[0-9]*'` in get_version.sh line 25 will still match v-prefixed tags like 'v0.10.0' (since they contain digits). This defeats the purpose of the fix. Change the pattern to `'[0-9].*'` to ensure tags start with a digit.

**`deployment/docker-build/get_version.sh`** (line 25): Bug: The pattern `--match '[0-9]*'` matches any tag containing digits, including `v0.10.0`. This means if a v-prefixed tag exists, it could be picked as the fallback. Change to `--match '[0-9].*'` to require tags to START with a digit, excluding v-prefixed tags.

foxpatch-aleph (commented):
This PR correctly fixes the v-prefixed tag issue by filtering git tags to only match semantic version patterns (X.Y.Z). The new tagging strategy in publish.sh is well-documented and handles stable releases, pre-releases, and dev builds appropriately. The deletion of publish-alpha.sh is correct as it was producing incorrect tags. The docker-compose.yml update to use :stable is appropriate for production deployments. There is one minor issue: a duplicate comment line in build.sh that should be cleaned up.

**`deployment/docker-build/build.sh`** (line 19): Duplicate comment - line 19 says "Sets IMAGE_TAG and PEP440_VERSION" and line 20 says "Sets IMAGE, IMAGE_TAG, and PEP440_VERSION". Remove line 19 since line 20 is more accurate (includes IMAGE).

**`deployment/docker-build/get_version.sh`** (line 25): The `--match '[0-9]*'` pattern is a bit permissive (matches any tag starting with a digit). Consider using `--match '[0-9]*.[0-9]*.[0-9]*'` to be more specific about version-like tags, though the current approach works for the stated goal.

foxpatch-aleph (approved):
This PR correctly fixes the version detection issue by filtering to only X.Y.Z-style tags, avoiding stale v-prefixed tags. The new publish.sh tagging strategy is well-documented and correctly distinguishes between stable releases, pre-releases, and dev builds using appropriate regex patterns. The deletion of publish-alpha.sh is appropriate as its functionality is now handled by the improved publish.sh logic.

**`deployment/docker-build/publish.sh`** (line 55): The pre-release regex `^[0-9]+\.[0-9]+\.[0-9]+-(rc|a|alpha|beta)[0-9]+$` requires a digit after the pre-release type. This correctly excludes dev builds like `0.10.0-a1b2c3d`, but would also reject pre-releases without a trailing digit like `0.10.1-rc`. If your release workflow ever uses such tags, the pattern would need adjustment. Currently this appears intentional and correct.

**`deployment/docker-build/get_version.sh`** (line 17): The `sort -rV | head -1` pattern is a nice touch for handling edge cases where multiple version tags might point to the same commit (e.g., if a tag was retagged). This ensures consistent behavior.

PR #1099, opened 2026-04-15
feat: add filters and sorting to /credit_history endpoint

amalcaraz:
# Summary

  Adds three new capabilities to `GET /api/v0/addresses/{address}/credit_history`:

  - **`has_expiration` filter** (boolean) — filter entries by whether they have an expiration date (`true` = only expiring credits, `false` = only non-expiring)
  - **`exclude_payment_method` filter** (comma-separated list) — exclude entries matching given payment methods (e.g. `exclude_payment_method=credit_expense` to see only distributions and transfers)
  - **`sort_by` / `sort_order` params** — sort results by any field (`message_timestamp`, `expiration_date`, `payment_method`, `amount`, `origin`, `tx_hash`, `provider`) in ascending or descending order. Default remains `message_timestamp DESC`. Nullable columns use `NULLS LAST` ordering.

  Cursor-based pagination is fully supported with custom sort fields — the cursor encodes the sort field name and value, with backward compatibility for existing cursors.

  ## Changes

  - **`src/aleph/types/sort_order.py`** — new `SortByCreditHistory` enum
  - **`src/aleph/schemas/api/accounts.py`** — new query params on `GetAccountCreditHistoryQueryParams`
  - **`src/aleph/toolkit/cursor.py`** — sort-aware cursor encode/decode functions
  - **`src/aleph/db/accessors/balances.py`** — dynamic sorting, generalized keyset pagination, extracted shared filter logic into `_apply_credit_history_filters()`
  - **`src/aleph/web/controllers/accounts.py`** — wired new params in handler, cursor sort validation
  - **`tests/db/test_credit_balances.py`** — 8 new tests covering filters, sorting, nulls ordering, and combinations

  ## Test plan

  - [x] `has_expiration=true` returns only entries with expiration_date
  - [x] `has_expiration=false` returns only entries without expiration_date
  - [x] `exclude_payment_method` filters out specified payment methods
  - [x] `count_address_credit_history` respects `exclude_payment_method`
  - [x] `sort_by=amount` with ASC/DESC ordering
  - [x] `sort_by=expiration_date` places NULLs last in both directions
  - [x] Combined `has_expiration` + `exclude_payment_method` filters
  - [x] All existing tests still pass

### DIFF
--- a/src/aleph/db/accessors/balances.py
+++ b/src/aleph/db/accessors/balances.py
@@ -883,7 +883,6 @@ CREDIT_HISTORY_SORT_COLUMN_MAP = {
 # Columns that are nullable and need NULLS LAST handling
 _NULLABLE_SORT_COLUMNS = {
     SortByCreditHistory.EXPIRATION_DATE,
-    SortByCreditHistory.PAYMENT_METHOD,
     SortByCreditHistory.ORIGIN,
     SortByCreditHistory.TX_HASH,
     SortByCreditHistory.PROVIDER,
@@ -1035,6 +1034,7 @@ def get_address_credit_history(
                         )
                     )
                 )
+                # NULLs come after all non-NULLs due to NULLS LAST ordering
                 | (primary_col.is_(None))
             )
         else:
@@ -1050,6 +1050,7 @@ def get_address_credit_history(
                         )
                     )
                 )
+                # NULLs come after all non-NULLs due to NULLS LAST ordering
                 | (primary_col.is_(None))
             )
 
--- a/src/aleph/db/accessors/balances.py
+++ b/src/aleph/db/accessors/balances.py
@@ -18,6 +18,7 @@ from aleph.toolkit.constants import (
 )
 from aleph.toolkit.timestamp import timestamp_to_datetime, utc_now
 from aleph.types.db_session import DbSession
+from aleph.types.sort_order import SortByCreditHistory, SortOrder
 
 
 def _apply_credit_precision_multiplier(
@@ -869,11 +870,28 @@ def validate_credit_transfer_balance(
     return current_balance >= total_transfer_amount
 
 
-def get_address_credit_history(
-    session: DbSession,
-    address: str,
-    page: int = 1,
-    pagination: int = 0,
+CREDIT_HISTORY_SORT_COLUMN_MAP = {
+    SortByCreditHistory.MESSAGE_TIMESTAMP: AlephCreditHistoryDb.message_timestamp,
+    SortByCreditHistory.EXPIRATION_DATE: AlephCreditHistoryDb.expiration_date,
+    SortByCreditHistory.PAYMENT_METHOD: AlephCreditHistoryDb.payment_method,
+    SortByCreditHistory.AMOUNT: AlephCreditHistoryDb.amount,
+    SortByCreditHistory.ORIGIN: AlephCreditHistoryDb.origin,
+    SortByCreditHistory.TX_HASH: AlephCreditHistoryDb.tx_hash,
+    SortByCreditHistory.PROVIDER: AlephCreditHistoryDb.provider,
+}
+
+# Columns that are nullable and need NULLS LAST handling
+_NULLABLE_SORT_COLUMNS = {
+    SortByCreditHistory.EXPIRATION_DATE,
+    SortByCreditHistory.PAYMENT_METHOD,
+    SortByCreditHistory.ORIGIN,
+    SortByCreditHistory.TX_HASH,
+    SortByCreditHistory.PROVIDER,
+}
+
+
+def _apply_credit_history_filters(
+    query: Select,
     tx_hash: Optional[str] = None,
     token: Optional[str] = None,
     chain: Optional[str] = None,
@@ -881,44 +899,10 @@ def get_address_credit_history(
     origin: Optional[str] = None,
     origin_ref: Optional[str] = None,
     payment_method: Optional[str] = None,
-    after_time: Optional[dt.datetime] = None,
-    after_credit_ref: Optional[str] = None,
-    after_credit_index: Optional[int] = None,
-    cursor_mode: bool = False,
-) -> Sequence[AlephCreditHistoryDb]:
-    """
-    Get paginated credit history entries for a specific address, ordered from newest to oldest.
-
-    Args:
-        session: Database session
-        address: Address to get credit history for
-        page: Page number (starts at 1)
-        pagination: Number of entries per page (0 for all entries)
-        tx_hash: Filter by transaction hash
-        token: Filter by token
-        chain: Filter by chain
-        provider: Filter by provider
-        origin: Filter by origin
-        origin_ref: Filter by origin reference
-        payment_method: Filter by payment method
-        after_time: Cursor-based: only return entries older than this timestamp
-        after_credit_ref: Cursor-based: tiebreaker credit_ref for entries with the same timestamp
-        after_credit_index: Cursor-based: tiebreaker credit_index for entries with the same timestamp and credit_ref
-
-    Returns:
-        List of credit history entries ordered by message_timestamp desc
-    """
-    query = (
-        select(AlephCreditHistoryDb)
-        .where(AlephCreditHistoryDb.address == address)
-        .order_by(
-            AlephCreditHistoryDb.message_timestamp.desc(),
-            AlephCreditHistoryDb.credit_ref.desc(),
-            AlephCreditHistoryDb.credit_index.desc(),
-        )
-    )
-
-    # Apply filters
+    has_expiration: Optional[bool] = None,
+    exclude_payment_method: Optional[List[str]] = None,
+) -> Select:
+    """Apply common filters to a credit history query."""
     if tx_hash is not None:
         query = query.where(AlephCreditHistoryDb.tx_hash == tx_hash)
     if token is not None:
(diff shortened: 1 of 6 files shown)


foxpatch-aleph (changes requested):
The PR adds useful filtering and sorting capabilities to the credit history endpoint with good test coverage. However, there are three bugs that need fixing: (1) a pagination bug where NULL values appear on every page instead of just the final page when sorting by nullable columns, (2) `payment_method` is incorrectly listed in `_NULLABLE_SORT_COLUMNS` since it always has a value, and (3) the cursor encoding should include `sort_order` to prevent incorrect pagination when switching between ASC/DESC directions.

**`src/aleph/db/accessors/balances.py`** (line 899): **BUG: Pagination with NULL values is incorrect.**

The condition `| (primary_col.is_(None))` on line 911 causes NULL values to appear on *every* page when `after_sort_value` is not None. This breaks pagination because:

1. Page 1 returns: non-NULL values + NULLs
2. Page 2 (with cursor pointing to a non-NULL value) would also return: remaining non-NULL values + NULLs again

The fix is to remove `| (primary_col.is_(None))` from the non-NULL cursor case. The NULL-only case (lines 874-897) already handles fetching NULLs correctly when the cursor is at a NULL value.

Same issue exists on line 926 for the ASC case.

Correct logic:
```python
elif is_desc:
    query = query.where(
        (primary_col < after_sort_value)
        | (
            (primary_col == after_sort_value)
            & (
                (AlephCreditHistoryDb.credit_ref < after_credit_ref)
                | (
                    (AlephCreditHistoryDb.credit_ref == after_credit_ref)
                    & (AlephCreditHistoryDb.credit_index < after_credit_index)
                )
            )
        )
    )
```

Add a pagination test to verify this works correctly across multiple pages.

**`src/aleph/db/accessors/balances.py`** (line 759): `payment_method` should not be in `_NULLABLE_SORT_COLUMNS`.

Looking at all code paths that insert credit history:
- `update_credit_balances_distribution`: sets `payment_method` from entry (line 550)
- `update_credit_balances_expense`: always sets `payment_method = "credit_expense"` (line 633)
- `update_credit_balances_transfer`: always sets `payment_method = "credit_transfer"` (lines 693, 718)

`payment_method` always has a value in practice. Only `expiration_date` is genuinely nullable (can be None for distributions). Remove `SortByCreditHistory.PAYMENT_METHOD` from the set.

**`src/aleph/toolkit/cursor.py`** (line 93): **Cursor should include `sort_order` for correct pagination.**

The current cursor encoding doesn't include `sort_order`. This means:
1. A cursor created with `sort_order=DESC` is identical to one created with `sort_order=ASC` for the same entry
2. If a user switches from DESC to ASC and reuses a cursor, they'd get incorrect results

Update `encode_credit_history_sort_cursor` to include `sort_order`:
```python
def encode_credit_history_sort_cursor(
    sort_by: str,
    sort_value: Any,
    sort_order: SortOrder,
    credit_ref: str,
    credit_index: int,
) -> str:
    value = (
        sort_value.isoformat() if isinstance(sort_value, dt.datetime) else sort_value
    )
    return encode_cursor({
        "s": sort_by, 
        "v": value, 
        "o": sort_order,  # Add sort_order
        "r": credit_ref, 
        "i": credit_index
    })
```

Also update `decode_credit_history_sort_cursor` and the controller to validate `sort_order` matches the request (similar to the existing `sort_by` validation).

**`tests/db/test_credit_balances.py`** (line 1585): Consider adding a test for cursor-based pagination with nullable sort columns to catch the pagination bug described above. Test that when sorting by `expiration_date` DESC, NULL values appear only on the final page, not on every page.

amalcaraz:
Thanks for the thorough review! Here's what I addressed:

  **Fixed:**

  - **`payment_method` removed from `_NULLABLE_SORT_COLUMNS`** — Agreed, it always has a value across all insertion paths (`credit_distribution`, `credit_expense`, `credit_transfer`). Removed.
  - **`sort_order` added to cursor encoding** — Good catch. The cursor now encodes `sort_order` alongside `sort_by`, and the handler validates both match the request. This prevents incorrect pagination if a client reuses a cursor with a different sort direction.

  **Regarding the NULL pagination bug — I respectfully disagree:**

  The `| (primary_col.is_(None))` clause in the non-NULL cursor case is actually required for correct behavior with `NULLS LAST` ordering. Here's why:

  With `ORDER BY expiration_date DESC NULLS LAST`, the ordering is:
  2026-09-01  ← largest non-NULL first
  2026-07-01
  2026-05-01  ← smallest non-NULL last
  NULL        ← NULLs come after all non-NULLs
  NULL

  When the cursor points to a non-NULL value (e.g. `2026-07-01`), the keyset condition needs to match everything that comes *after* it in this ordering:
  - `primary_col < 2026-07-01` → matches `2026-05-01` ✓
  - `primary_col IS NULL` → matches the NULLs at the end ✓

  Without `| (primary_col.is_(None))`, SQL's `NULL < 2026-07-01` evaluates to `NULL` (falsy), so NULLs would **never** be fetched — they'd be permanently unreachable once the cursor passes any non-NULL value.

  The reviewer's concern was that NULLs would appear on every page, but that doesn't happen because the `ORDER BY` + `LIMIT` ensures results come back in order. NULLs only appear once the non-NULL entries before the limit are exhausted.

  I added a multi-page cursor pagination test that walks through all 5 entries (3 with dates + 2 NULLs) in pages of 2, verifying:
  - NULLs don't appear on page 1
  - NULLs appear starting on page 2 (after the non-NULL entries)
  - Every entry appears exactly once across all pages

foxpatch-aleph (approved):
This PR implements a well-designed feature for filtering and sorting credit history entries. The implementation correctly handles nullable columns with NULLS LAST ordering, maintains backward compatibility with existing cursors, and includes comprehensive tests covering filters, sorting, and cursor pagination with nullable sort fields. The code is clean, well-documented, and follows existing patterns in the codebase.

**`src/aleph/db/accessors/balances.py`** (line 841): Minor nit: For non-nullable columns, you could apply `.nullslast()` consistently for uniformity, though this is not strictly necessary since non-nullable columns won't have NULLs anyway.

**`src/aleph/toolkit/cursor.py`** (line 129): The backward compatibility fallback assumes `-1` (DESC) order for old cursors. This is correct given the default, but consider adding a comment explaining this assumption.

**`src/aleph/web/controllers/accounts.py`** (line 836): Good validation of cursor sort parameters against request parameters. This prevents pagination inconsistencies when users change sort settings mid-pagination.

**`tests/db/test_credit_balances.py`** (line 1587): Excellent test coverage for the complex case of cursor pagination with nullable sort columns. This test validates that NULLs appear only on the last page when sorting DESC with NULLS LAST ordering.

foxpatch-aleph (changes requested):
The PR adds well-designed filtering and sorting capabilities to the credit history endpoint with comprehensive tests. However, there's a correctness issue: `payment_method` is nullable in the database model but missing from `_NULLABLE_SORT_COLUMNS`, which could cause incorrect pagination behavior when sorting by payment_method with NULL values. Additionally, the old `encode_credit_history_cursor` and `decode_credit_history_cursor` functions are now unused and should be removed for clarity.

**`src/aleph/db/accessors/balances.py`** (line 844): **BUG**: `SortByCreditHistory.PAYMENT_METHOD` is missing from `_NULLABLE_SORT_COLUMNS`. According to `src/aleph/db/models/balances.py:53`, `payment_method` is nullable (`Mapped[Optional[str]]`). This means sorting by `payment_method` with NULL values won't use NULLS LAST ordering, potentially causing pagination issues where NULL entries appear inconsistently. Add `SortByCreditHistory.PAYMENT_METHOD` to the set.

**`src/aleph/toolkit/cursor.py`** (line 77): **Cleanup**: The old `encode_credit_history_cursor` and `decode_credit_history_cursor` functions (lines 77-90) are now unused since the PR switched to the sort-aware versions. Remove these to avoid confusion. The backward compatibility logic in `decode_credit_history_sort_cursor` handles old cursors, so these functions aren't needed.

**`src/aleph/db/accessors/balances.py`** (line 852): **Suggestion**: Consider adding a docstring to `_apply_credit_history_filters()` explaining the filter semantics, especially for `has_expiration` (True = entries WITH expiration, False = entries WITHOUT expiration) and `exclude_payment_method` (NOT IN filter).

foxpatch-aleph (dismissed):
This PR correctly implements filtering and sorting for the credit history endpoint. The implementation handles nullable columns with NULLS LAST ordering, cursor-based pagination with sort-aware cursors, and backward compatibility for existing cursors. The code is clean, well-structured, and the test coverage is comprehensive, including edge cases for NULL pagination and combined filters.

**`src/aleph/db/accessors/balances.py`** (line 928): The ordering logic for non-nullable columns could be simplified to use the same pattern as nullable columns (i.e., always use `.asc()` or `.desc()`), but the current implementation is correct and clear.

**`src/aleph/toolkit/cursor.py`** (line 102): Consider handling the case where `sort_value` might be `Decimal` or other non-datetime types that have special serialization needs. Currently only `datetime` is handled specially.

**`src/aleph/schemas/api/accounts.py`** (line 171): The `split_exclude_payment_method` validator could filter out empty strings from the result to avoid potential issues with malformed input like `"credit_expense,"` producing `['credit_expense', '']`.

**`tests/db/test_credit_balances.py`** (line 1550): The test helper could include all fields of `AlephCreditHistoryDb` (like `token`, `chain`, `price`, `bonus_amount`) for completeness, even if they're nullable and not required for these tests.

foxpatch-aleph (dismissed):
The PR is well-implemented with correct cursor-based pagination logic, proper NULLS LAST handling for nullable sort columns, and good backward compatibility for old cursors. The extracted `_apply_credit_history_filters()` function cleanly deduplicates filter logic between the count and query accessors. Test coverage is comprehensive for the core functionality (filters, sorting, nulls ordering, pagination across NULL boundary). One minor redundant check and one small naming nit; no bugs or security issues found.

**`src/aleph/web/controllers/accounts.py`** (line 840): Redundant check: `if cursor:` on line 840 is inside the `if cursor is not None:` block (line 834), so it's always True. The inner `if` can be removed and the code block dedented one level.

**`src/aleph/web/controllers/accounts.py`** (line 861): Since `SortOrder` is an `IntEnum` (ASCENDING=1, DESCENDING=-1), `query_params.sort_order` is already an int. The `int()` cast is unnecessary but harmless.

**`src/aleph/schemas/api/accounts.py`** (line 181): The `split_exclude_payment_method` validator uses `LIST_FIELD_SEPARATOR` (",") but does not strip whitespace from individual values. If a caller passes "credit_expense, credit_distribution" (with spaces), the list would contain ' credit_distribution' with a leading space. Consider adding `.strip()` to each element: `return [s.strip() for s in v.split(LIST_FIELD_SEPARATOR)]`. This may be intentional, but worth noting.

**`src/aleph/schemas/api/accounts.py`** (line 180): The validator lacks a `@classmethod` decorator on the first parameter. While Pydantic v2 handles this via `model_validator` or `field_validator` with `mode='before'`, the conventional pattern is `def split_exclude_payment_method(cls, v):`. This works as-is but may trigger linter warnings depending on config.

**`tests/db/test_credit_balances.py`** (line 1954): Good test coverage for `exclude_payment_method` via the accessor. Consider adding a small integration test that exercises the comma-separated string parsing path in the `split_exclude_payment_method` validator (i.e., going through the API layer), since the current tests pass the list directly.

foxpatch-aleph (commented):
The implementation is correct and well-tested. The only notable gap is the reduced docstring on get_address_credit_history, which lost its parameter documentation during refactoring. Everything else — filter logic, NULLS LAST ordering, cursor encoding/decoding, backward compatibility, and validation — is solid.

**`src/aleph/db/accessors/balances.py`** (line 951): The docstring for get_address_credit_history is now a single line. The original function had full parameter documentation. Consider restoring at least the key new parameters (has_expiration, exclude_payment_method, sort_by, sort_order, after_sort_value) to aid maintainability.

**`src/aleph/schemas/api/accounts.py`** (line 173): The sort_by Field description is just 'Field to sort by.' — consider listing the allowed values for discoverability, e.g. 'One of: message_timestamp, expiration_date, payment_method, amount, origin, tx_hash, provider.'

**`src/aleph/toolkit/cursor.py`** (line 101): encode_credit_history_sort_cursor only converts datetime values via isoformat(); non-datetime values (amount as int, strings like payment_method) are stored as-is. This works because int and str both serialize fine in the cursor JSON, but worth noting that sort_value is not type-checked here.

odesenfans (approved):

PR #1100, opened 2026-04-15
feat: add include_credit_details param to /balance endpoint

amalcaraz:
## Summary

  Adds an optional `include_credit_details` boolean parameter to `GET /api/v0/addresses/{address}/balance`. When enabled, the response includes a `credit_balance_details` field with a FIFO-accurate breakdown of the remaining credit balance grouped by expiration date.

  Example response with `include_credit_details=true`:
  ```json
  {
    "address": "0x...",
    "balance": 10.5,
    "locked_amount": 2.0,
    "credit_balance": 1200,
    "credit_balance_details": [
      { "expiration_date": null, "amount": 300 },
      { "expiration_date": "2026-06-01T00:00:00+00:00", "amount": 600 },
      { "expiration_date": "2026-12-01T00:00:00+00:00", "amount": 300 }
    ]
  }
```

  The amounts reflect actual remaining credits after FIFO consumption — partially consumed credits are correctly accounted for. Non-expiring credits (null) are listed first, followed by expiring buckets sorted by date ascending.

  When include_credit_details is omitted or false, the response is unchanged (uses the existing cached balance path).

##  Changes

  - src/aleph/schemas/api/accounts.py — add include_credit_details to GetAccountQueryParams, new CreditBalanceDetailItem model, extend GetAccountBalanceResponse with optional credit_balance_details
  - src/aleph/db/accessors/balances.py — refactor FIFO logic into shared _apply_fifo_consumption helper, add CreditBalanceDetail dataclass and get_credit_balance_with_details() public function
  - src/aleph/web/controllers/accounts.py — wire include_credit_details in get_account_balance handler, update OpenAPI docstring
  - tests/db/test_credit_balances.py — 7 new tests for the details calculation

##  Test plan

  - Non-expiring credits only → single bucket with null expiration
  - Mixed expiration dates → grouped correctly, sorted null first then by date ASC
  - Partial FIFO consumption → remaining amounts are accurate per bucket
  - Fully consumed credits → empty details list, total = 0
  - Expired credits excluded from details
  - No credit history → total = 0, empty details
  - Sum of detail amounts equals total balance (invariant check)
  - All existing tests still pass

### DIFF
--- a/src/aleph/db/accessors/balances.py
+++ b/src/aleph/db/accessors/balances.py
@@ -768,9 +768,7 @@ def update_credit_balances_transfer(
     # Compute sender's remaining credits once for all entries in this transfer
     sender_remaining: list[PositiveCredit] = []
     if not is_whitelisted:
-        sender_remaining = _apply_fifo_consumption(
-            session, sender_address, last_update
-        )
+        sender_remaining = _apply_fifo_consumption(session, sender_address, last_update)
 
     for credit_entry in credits_list:
         recipient_address = credit_entry["address"]
--- a/src/aleph/db/accessors/balances.py
+++ b/src/aleph/db/accessors/balances.py
@@ -5,7 +5,7 @@ import time
 from dataclasses import dataclass
 from decimal import Decimal
 from io import StringIO
-from typing import Any, Dict, Mapping, Optional, Sequence
+from typing import Any, Dict, List, Mapping, Optional, Sequence, Tuple
 
 from aleph_message.models import Chain
 from sqlalchemy import func, select, text
@@ -240,7 +240,13 @@ class NegativeAmount:
     timestamp: dt.datetime
 
 
-def _get_remaining_credits_fifo(
+@dataclass
+class CreditBalanceDetail:
+    expiration_date: Optional[dt.datetime]
+    amount: int
+
+
+def _apply_fifo_consumption(
     session: DbSession, address: str, now: Optional[dt.datetime] = None
 ) -> list[PositiveCredit]:
     """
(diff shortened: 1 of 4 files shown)


foxpatch-aleph (approved):
This PR correctly implements the `include_credit_details` parameter for the `/balance` endpoint. The implementation is clean and well-structured: (1) The FIFO consumption logic is properly refactored into a shared `_apply_fifo_consumption` helper that both the existing balance calculation and new details calculation use, ensuring consistency. (2) The new `get_credit_balance_with_details` function correctly groups credits by expiration date and sorts them with non-expiring first, then by date ascending. (3) The API schema changes are minimal and follow existing patterns. (4) The controller properly conditionally calls the details endpoint only when requested, preserving the cached path for the common case. (5) The 7 new tests comprehensively cover: non-expiring only, mixed expiration dates, partial FIFO consumption, fully consumed credits, expired credits exclusion, no history, and an invariant check that details sum equals total. The code is correct, well-tested, and ready to merge.

**`src/aleph/db/accessors/balances.py`** (line 363): Minor nit: The sort key `lambda x: (x[0] is not None, x[0] or dt.datetime.min)` is clever but could be slightly clearer. The `x[0] or dt.datetime.min` fallback is only reached when `x[0]` is not None (since the first tuple element handles the None case), so it's effectively just `x[0]`. Consider simplifying to `lambda x: (x[0] is not None, x[0])` for clarity.

**`src/aleph/web/controllers/accounts.py`** (line 357): The check `if credits is None: credits = 0` is unreachable since `get_credit_balance_with_details` always returns an int (not None). However, this is a defensive pattern that doesn't hurt and may be useful if the function signature changes in the future.

foxpatch-aleph (dismissed):
This PR cleanly implements the `include_credit_details` feature for the balance endpoint. The implementation correctly uses FIFO-accurate credit consumption logic, properly groups credits by expiration date, and sorts results with non-expiring credits first. The code is well-structured, well-documented, and includes 7 comprehensive tests covering all key scenarios including partial consumption, expired credits, and the invariant that detail amounts sum to total balance.

**`src/aleph/web/controllers/accounts.py`** (line 357): The check `if credits is None: credits = 0` is unreachable since both `get_credit_balance_with_details` and `get_credit_balance` always return `int` (never `None`). Consider removing this dead code.

**`src/aleph/db/accessors/balances.py`** (line 414): The sorting key uses `dt.datetime.min` as a fallback for `None` values. While this works correctly, a more explicit approach would be to separate keys into two lists (None vs datetime) and concatenate, which would be clearer. However, the current implementation is correct and efficient.

**`src/aleph/db/accessors/balances.py`** (line 252): Good refactoring of the FIFO logic into `_apply_fifo_consumption`. The docstring correctly notes that expired credits are included in the return and callers must filter them - this is an important contract that's well documented.

foxpatch-aleph (approved):
This PR correctly implements an optional `include_credit_details` parameter that provides a FIFO-accurate breakdown of credit balances by expiration date. The implementation properly reuses the existing FIFO consumption logic via the refactored `_apply_fifo_consumption` helper, correctly groups and sorts credits (non-expiring first, then by date ascending), and excludes expired credits. The controller correctly only calls the expensive details function when the flag is enabled, preserving the cached path for normal requests. The 7 new tests comprehensively cover edge cases including partial consumption, fully consumed credits, expired exclusions, and the invariant that detail amounts sum to total balance.

**`src/aleph/web/controllers/accounts.py`** (line 357): The check `if credits is None: credits = 0` is dead code. Both `get_credit_balance_with_details` and `get_credit_balance` always return `int`, never `None`. This check can be safely removed.

PR #1101, opened 2026-04-15
fix: resolve consumed_credits for volume resources in /costs endpoint

amalcaraz:
## Summary

  - Fix `consumed_credits` showing 0 for volume (STORE) resources in `/api/v0/costs?include_details=1`
  - Root cause: volume credit expenses store the resource hash in `origin_ref` (from `ref` field), not `origin` (from `execution_id`), but the queries only checked `origin`
  - Both `get_total_consumed_credits()` and `get_consumed_credits_by_resource()` now use `COALESCE(NULLIF(origin, ''), origin_ref)` to fall back to `origin_ref` when `origin` is empty

  ## Test plan

  - [x] 3 new unit tests covering the origin_ref fallback for volumes
  - [x] All existing credit_balances and costs tests pass (23 + 15 = 38 tests)
  - [x] Linting and mypy pass
  - [ ] Manual verification: query `/api/v0/costs?address=0xB6B5358493AF8159B17506C5cC85df69193444BC&include_details=1` and confirm volume resources show non-zero `consumed_credits`

### DIFF
--- a/src/aleph/db/accessors/balances.py
+++ b/src/aleph/db/accessors/balances.py
@@ -1034,7 +1034,11 @@ def get_total_consumed_credits(
     if address:
         query = query.where(AlephCreditHistoryDb.address == address)
     if item_hash:
-        query = query.where(AlephCreditHistoryDb.origin == item_hash)
+        effective_origin = func.coalesce(
+            func.nullif(AlephCreditHistoryDb.origin, ""),
+            AlephCreditHistoryDb.origin_ref,
+        )
+        query = query.where(effective_origin == item_hash)
 
     result = session.execute(query).scalar()
     return result or 0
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
This PR correctly fixes the bug where volume resources showed 0 consumed_credits. The fix uses COALESCE(NULLIF(origin, ''), origin_ref) to fall back to origin_ref when origin is empty, which correctly handles both instance resources (with execution_id → origin) and volume resources (with ref → origin_ref). The three new unit tests provide good coverage for the fix, including a mixed scenario test. The code is clean, follows existing patterns, and has no security concerns.

odesenfans (dismissed):

foxpatch-aleph (approved):
The fix correctly resolves consumed_credits for volume (STORE) resources by using COALESCE(NULLIF(origin, ''), origin_ref) to fall back to origin_ref when origin is empty. The root cause analysis is sound—update_credit_balances_expense stores execution_id in origin and ref in origin_ref, so volume expenses (which lack execution_id) have origin='' and origin_ref set to the resource hash. The fix is minimal, focused, and properly parameterized through SQLAlchemy.

**`src/aleph/db/accessors/balances.py`** (line 1037): Consider extracting the effective_origin expression into a module-level SQLAlchemy column element or helper function, since it's now duplicated in both get_total_consumed_credits and get_consumed_credits_by_resource. This would prevent drift if the logic ever needs to change again. Not urgent for this fix.

**`src/aleph/db/accessors/balances.py`** (line 1096): The effective_origin.in_(item_hashes) filter is safe—SQLAlchemy parameterizes the list values. No SQL injection risk from query parameters flowing through the controller.

**`tests/db/test_credit_balances.py`** (line 1973): Good coverage in the mixed test (test_get_consumed_credits_by_resource_uses_origin_ref_for_volumes) that verifies both instance (origin set) and volume (origin_ref fallback) resources resolve correctly in the same query.

PR #1107, opened 2026-04-21
Allow /messages/{hash}/content to return non-POST message content

aliel:
The endpoint was POST-only and rejected all other message types with a
  confusing 422 "Invalid message hash status" error, the status was
  actually PROCESSED, but the real failure was that the handler's
  isinstance(message, PostMessage) check refused any non-POST type.

  Extend to all message types:
  - POST: return content.content (unwrapped user payload - unchanged)
  - Others (PROGRAM, INSTANCE, STORE, AGGREGATE, FORGET): return full message.content, since those types don't double-wrap their payload.

### DIFF
--- a/src/aleph/web/controllers/messages.py
+++ b/src/aleph/web/controllers/messages.py
@@ -884,9 +884,7 @@ async def view_message_content(request: web.Request):
             session=session, status_db=message_status_db
         )
 
-    if message_with_status.status != MessageStatus.PROCESSED or not isinstance(
-        message_with_status, ProcessedMessageStatus
-    ):
+    if not isinstance(message_with_status, ProcessedMessageStatus):
         raise web.HTTPUnprocessableEntity(
             text=(
                 f"Message {item_hash} is not processed "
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (commented):
## PR Review: #1107 — Allow /messages/{hash}/content to return non-POST message content

```json
{
  "verdict": "APPROVE",
  "summary": "This PR correctly extends the /messages/{hash}/content endpoint to support all message types. The refactored logic is cleaner and more maintainable. The isinstance check for ProcessedMessageStatus properly replaces the multi-part validation, and the content extraction logic correctly handles POST's nested content versus other types' direct content. Tests are updated to verify both code paths.",
  "comments": [
    {
      "path": "src/aleph/web/controllers/messages.py",
      "line": 887,
      "body": "The refactored validation is cleaner than the previous multi-part check. Using isinstance(message_with_status, ProcessedMessageStatus) properly ensures we have a processed message with a .message attribute, and the error message now correctly reports the actual status instead of the misleading 'Invalid message hash status'."
    },
    {
      "path": "src/aleph/web/controllers/messages.py",
      "line": 903,
      "body": "Using model_dump(mode='json') for non-POST types ensures all nested models are properly serialized to plain dicts before json.dumps(). This is consistent with the POST case where json.dumps serializes the BaseModel result."
    },
    {
      "path": "tests/api/test_get_message.py",
      "line": 216,
      "body": "The test correctly covers both POST (which unwraps content.content) and non-POST types (FORGET in the fixture, which returns full content). The test structure is improved by checking the response first, then branching on message type."
    }
  ]
}
```

foxpatch-aleph (approved):
This PR correctly extends the /messages/{hash}/content endpoint to support all message types, not just POST. The implementation is clean and efficient: it serializes the message once, then conditionally unwraps the nested content for POST messages while returning the full content for other types. The error message is improved to be more descriptive. Tests cover both POST (expects unwrapped content) and non-POST (expects full content) cases. No security issues or bugs identified.

**`tests/api/test_get_message.py`** (line 216): Minor: The test fixture only includes FORGET and POST message types for PROCESSED status. Consider adding test cases for PROGRAM, INSTANCE, STORE, and AGGREGATE types to fully cover the PR description's claim of supporting 'all message types'. This is not blocking - the implementation is generic enough that it will work correctly.

PR #1110, opened 2026-04-23
feat: authenticated IPFS uploads via /api/v0/ipfs/add_file

odesenfans:
## Summary

Extends `POST /api/v0/ipfs/add_file` to accept a signed STORE message in the multipart form, mirroring `/storage/add_file`. Users can now upload and anchor an IPFS pin in a single request, rather than pinning via the unauthenticated endpoint and submitting a STORE message separately.

Also adds an explicit size limit to the existing unauthenticated mode (was relying on aiohttp's default only) and rationalizes post-pin error handling so no failure mode leaves an orphaned pin on the local IPFS daemon.

## Endpoint contract (new)

```
POST /api/v0/ipfs/add_file
Content-Type: multipart/form-data

file:     <binary>                                      # required
metadata: {"message": <PendingInlineStoreMessage>,      # optional
           "sync": <bool>}
```

- Unauthenticated mode (no `metadata`): unchanged shape, now capped at `ipfs.max_unauthenticated_upload_file_size` (default 25 MiB) instead of aiohttp's implicit ~100 MiB. Pin gets a 24 h grace period.
- Authenticated mode: signature + `item_type=ipfs` + content schema verified **before** pinning; CID/balance checked **after**. On any post-pin failure, a 24 h grace period is applied so the pin is never stranded. The STORE message goes through `broadcast_and_process_message`, same as `/storage/add_file`.

## Config (new)

| Key | Default |
|---|---|
| `ipfs.max_upload_file_size` | 100 MiB |
| `ipfs.max_unauthenticated_upload_file_size` | 25 MiB |

Defaults match the equivalent `storage.*` values; separate keys so IPFS limits can be tuned independently going forward.

## Out of scope (tracked as follow-ups)

- The < 1 MiB "fetch from network instead of pin" quirk in `/messages` STORE processing — kept as-is; deserves its own async pin-queue design.
- Streaming uploads — only relevant if the 100 MiB cap is raised significantly.
- Hardening the open unauthenticated write surface.

## Design docs

- Spec: `docs/superpowers/specs/2026-04-23-authenticated-ipfs-uploads-design.md`
- Plan: `docs/superpowers/plans/2026-04-23-authenticated-ipfs-uploads.md`

## Test plan

- [x] `tests/api/test_ipfs.py` — 11 tests covering unauth regression, authenticated happy path, all documented error paths (bad signature, wrong item_type, size caps, CID mismatch, balance shortfall, malformed metadata, stat timeout), and the sub-threshold balance-check skip.
- [x] `tests/api/test_storage.py` — 17 existing tests still green, including the legacy `test_ipfs_add_file` regression guard.
- [x] `black` + `isort` + `ruff` clean on touched files.
- [ ] Review the spec wording about "no file read yet before signature": the implementation reads the file part during multipart iteration (same as `/storage/add_file`), then verifies the signature. The hard invariant ("no orphan pin") holds; the spec text overstates the cheap-fail-fast property. Worth a spec clarification post-merge.


foxpatch-aleph (dismissed):
The PR is well-implemented and matches the spec. The controller correctly handles the request lifecycle: parse multipart, validate message before pinning, pin to IPFS, post-pin checks (CID match, balance), persist with grace-period fallback, and broadcast. The 11 tests cover all documented paths including unauth regression, authenticated happy path, bad signature, wrong item_type, size caps, CID mismatch, balance shortfall, small-file balance skip, malformed metadata, and stat timeout. The stat timeout status code change from 404 to 504 is a semantic improvement. The config additions are clean. No bugs or security issues found.

foxpatch-aleph (commented):
The PR is well-designed and thoroughly tested. The authenticated IPFS upload feature correctly mirrors the existing /storage/add_file pattern, with solid post-pin error handling that ensures no orphaned pins. The 11 tests cover all documented error paths and regression scenarios. There are a couple of minor dead-code and stylistic nits that don't affect correctness but are worth cleaning up.

**`src/aleph/web/controllers/ipfs.py`** (line 139): Dead code: `metadata` is always bytes (from `part.read(decode=True)` at line 117), so the `isinstance(metadata, FileField)` branch at line 140 is never taken. Consider simplifying to just `metadata_bytes = metadata`.

**`src/aleph/web/controllers/ipfs.py`** (line 175): Dead code: `temp_file.read()` on a binary-mode file (`'w+b'` via `MultipartUploadedFile`) will always return bytes. The `isinstance(file_content, str)` check at line 175 is defensive but never triggers.

**`src/aleph/web/controllers/ipfs.py`** (line 194): Mutating `message_content.estimated_size_mib` in place on a pydantic model. This is consistent with the same pattern in storage.py line 287, so it's not a bug — just worth noting. Consider whether setting this via a model copy or a separate variable would be cleaner long-term.

**`src/aleph/web/controllers/ipfs.py`** (line 250): After successful pin + DB persist, the broadcast step can still fail (e.g., message rejected → 422). The file is already persisted in the DB at that point, so the client gets an error despite the file being stored. This matches the existing /storage/add_file behavior, but worth confirming this is intentional.

**`src/aleph/web/controllers/ipfs.py`** (line 106): The multipart iteration order is flexible (metadata can arrive before or after file). If metadata arrives first, signature validation runs before the file is fully read. This is fine functionally, but the comment at line 134 ('fail-fast') is slightly misleading in that ordering — the file is always fully read before pinning regardless.

**`tests/api/test_ipfs.py`** (line 251): Minor: `await response.text()` is called twice on lines 250-251. This works fine in aiohttp (response body is buffered), but could be collapsed into a single call for cleanliness.

foxpatch-aleph (commented):
Here's the review summary:

**Overall: Solid implementation with good test coverage.** The code correctly mirrors the `/storage/add_file` patterns — pre-pin validation, post-pin grace period fallback, and balance checks all behave as documented.

**Key findings:**

1. **Dead code (ipfs.py:143-145)** — `isinstance(metadata, FileField)` will never be true. `part.read(decode=True)` at line 117 already returns `bytes`, not a `FileField`. This is harmless but should be cleaned up.

2. **Content-type comparison (ipfs.py:95)** — Uses exact equality with `"multipart/form-data"`. This is consistent with the storage controller, but could be brittle if aiohttp's `content_type` includes parameters. Worth a quick sanity check.

3. **Missing `add_bytes.assert_not_called()` in the authenticated-cap test** — The 101 MiB test (line 278) returns 413 from the pre-pin size check but doesn't verify no pin was attempted. The bad_signature test does this correctly; the authenticated-cap test should too.

4. **Response body assertions in tests** — Several tests assert status codes but don't check the response body for the expected error reason. This makes debugging slower when error messages change.

5. **Grace period exception handler** — The nested `try/except` at lines 227-252 is correct but the broad `except Exception` could confuse reviewers. A one-line comment explaining the intent would help.

None of these are bugs that would cause runtime issues, so the code is safe to merge with the minor cleanup items above.

PR #1111, opened 2026-04-23
build: exclude .git from docker image, require VERSION build-arg

odesenfans:
## Summary

- `.git/` was being shipped inside the runtime Docker image via `COPY . .` in the builder stage, then re-copied into the final stage via `COPY --from=builder /opt/pyaleph /opt/pyaleph`. On a CI checkout where history is expanded with `git fetch --prune --unshallow --tags`, this can contribute a noticeable amount to the image size and also drifts between builds.
- Exclude `.git`, `.gitignore`, and `.gitattributes` from the build context via `.dockerignore`.
- With `.git/` gone, hatch-vcs can no longer derive the version during `pip install -e .`, so the already-declared `VERSION` build-arg (wired to `SETUPTOOLS_SCM_PRETEND_VERSION`) is now required. A guard `RUN : "${VERSION:?...}"` fails the build early with a clear message if it's missing.
- CI computes the version via `hatch version` and passes it as `--build-arg VERSION=...`.
- The `RUN git config --global --add safe.directory /opt/pyaleph` line is removed — no longer needed since setuptools-scm/hatch-vcs never runs against the in-image tree.

Context: investigating why `0.10.1-rc7` ended up several hundred MB larger than `0.10.1-rc6` despite a trivial source diff. This change removes one of the non-determinism sources (varying `.git/` size across builds) and is a first step; separate follow-ups could look at pinning transitive pip deps and the rolling `ubuntu:24.04` base.

## Test plan

- [x] Local build with `--build-arg VERSION=$(hatch version)` succeeds; final image is 754 MB
- [x] `/opt/pyaleph/.git` does not exist in the resulting image
- [x] `importlib.metadata.version('pyaleph')` inside the image returns the version passed via `--build-arg`
- [x] Build without `--build-arg VERSION=...` fails early with the guard message
- [ ] CI build green on this PR
- [ ] Verify image size on CI is smaller than the most recent `main`/`dev` build

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -52,6 +52,11 @@ cover/*
 MANIFEST
 .github/
 
+# Git metadata — version is injected via --build-arg VERSION
+.git
+.gitignore
+.gitattributes
+
 # Per-project virtualenvs
 .venv*/
 venv*/
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,10 +26,19 @@ jobs:
       - name: Download Docker cache image (if available)
         run: docker pull ghcr.io/$GITHUB_REPOSITORY/build-cache || true
 
-      - name: Build the Docker image
+      - name: Compute version
+        id: version
         run: |
           git fetch --prune --unshallow --tags
-          docker build . -t pyaleph-node:${GITHUB_REF##*/} -f deployment/docker-build/pyaleph.dockerfile --cache-from=ghcr.io/$GITHUB_REPOSITORY/build-cache
+          python3 -m pip install --quiet --user hatch hatch-vcs
+          echo "value=$(hatch version)" >> "$GITHUB_OUTPUT"
+
+      - name: Build the Docker image
+        run: |
+          docker build . -t pyaleph-node:${GITHUB_REF##*/} \
+            --build-arg VERSION=${{ steps.version.outputs.value }} \
+            -f deployment/docker-build/pyaleph.dockerfile \
+            --cache-from=ghcr.io/$GITHUB_REPOSITORY/build-cache
 
       - name: Push the image to the cache
         # It's not possible to push packages from fork PRs.
--- a/.github/workflows/test-keygen.yml
+++ b/.github/workflows/test-keygen.yml
@@ -24,10 +24,19 @@ jobs:
       - name: Download Docker cache image (if available)
         run: docker pull ghcr.io/$GITHUB_REPOSITORY/build-cache || true
 
-      - name: Build the Docker image
+      - name: Compute version
+        id: version
         run: |
           git fetch --prune --unshallow --tags
-          docker build . -t alephim/pyaleph-node:${GITHUB_REF##*/} -f deployment/docker-build/pyaleph.dockerfile --cache-from=ghcr.io/$GITHUB_REPOSITORY/build-cache
+          python3 -m pip install --quiet --user hatch hatch-vcs
+          echo "value=$(hatch version)" >> "$GITHUB_OUTPUT"
+
+      - name: Build the Docker image
+        run: |
+          docker build . -t alephim/pyaleph-node:${GITHUB_REF##*/} \
+            --build-arg VERSION=${{ steps.version.outputs.value }} \
+            -f deployment/docker-build/pyaleph.dockerfile \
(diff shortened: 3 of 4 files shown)


foxpatch-aleph (commented):
The changes are well-reasoned and correctly implement the goal of excluding .git from the Docker image while requiring an explicit VERSION build-arg. The approach is sound, the guard is a good safety net, and the CI changes are straightforward. No bugs or security issues found — a few minor notes below.

**`.dockerignore`** (line 58): Excluding `.gitattributes` is safe but worth noting: if any tooling (e.g. git hooks or CI) depends on `.gitattributes` being present elsewhere in the repo, it's fine since only the Docker build context is affected. No issue, just documenting the side effect.

**`.github/workflows/build.yml`** (line 33): Consider adding `--break-system-packages` to the `pip install` command or using `pipx` / a venv, since `pip --user` can emit deprecation warnings on Python 3.12+ and may fail in stricter environments. Not urgent, but a minor robustness improvement.

**`.github/workflows/build.yml`** (line 34): Consider adding a fallback for when `hatch version` fails (e.g. `set -euo pipefail` at the top of the script block). The current `run:` block doesn't set `set -e` explicitly, though GitHub Actions does abort on non-zero exit. Adding `set -euo pipefail` would also catch unset variable issues.

**`deployment/docker-build/pyaleph.dockerfile`** (line 52): The VERSION guard is clean. One minor suggestion: consider making this a `RUN echo "Using version: $VERSION"` so the version is visible in the build logs, which aids debugging when reproducing builds.

**`deployment/docker-build/pyaleph.dockerfile`** (line 50): The `COPY . .` in the builder stage now excludes .git thanks to .dockerignore. This means `hatch-vcs` / `setuptools-scm` won't find git history, which is exactly why the VERSION guard exists. The ordering (ARG before RUN) is correct — the guard fires before pip install, failing fast.

foxpatch-aleph (approved):
The PR correctly excludes .git/ from the Docker image via .dockerignore and enforces VERSION as a required build-arg with a clear guard message. The CI workflows are updated consistently to compute and pass the version. The changes are minimal, well-reasoned, and logically consistent across all three files. Minor nits around CI robustness (shallow-clone edge case, pip install latency) are noted but do not block approval.

**`.github/workflows/build.yml`** (line 32): git fetch --prune --unshallow --tags will error if the local clone is already unshallow. Consider adding || true or using a more robust fetch strategy.

**`.github/workflows/build.yml`** (line 33): Installing hatch/hatch-vcs via pip on every CI run adds latency. Consider caching these packages or using a pre-built helper image.

**`.github/workflows/test-keygen.yml`** (line 18): This workflow uses actions/checkout@v4 while build.yml still uses @v2. Inconsistent versions may produce different shallow-clone behavior.

**`.github/workflows/build.yml`** (line 33): pip install --quiet --user relies on ~/.local being on PATH. In containerized CI this usually works, but a temp venv or --break-system-packages would be more robust.

foxpatch-aleph (approved):
The PR is well-structured and correctly addresses the goal of excluding .git from the Docker image. The .dockerignore changes properly exclude git metadata, the Dockerfile guard ensures VERSION is provided with a clear error message, and both CI workflows are updated consistently. The approach of using hatch to compute the version and passing it via --build-arg is sound.

**`.dockerignore`** (line 56): Consider adding a comment explaining why .gitignore and .gitattributes are excluded (e.g., 'Git metadata and version-control files — not needed at runtime'). This makes the intent clearer for future maintainers.

**`deployment/docker-build/pyaleph.dockerfile`** (line 52): The error message could be slightly more specific to help developers. Consider: '--build-arg VERSION is required. Compute with: hatch version'

**`.github/workflows/build.yml`** (line 33): Consider pinning hatch and hatch-vcs versions (e.g., hatch==0.12.0 hatch-vcs==0.4.0) for reproducible version computation across CI runs. This is a minor nit since the current approach works, but pinning prevents unexpected changes if hatch releases a breaking version.

**`.github/workflows/test-keygen.yml`** (line 31): Same version pinning suggestion as build.yml (line 33). The duplicate compute-version step in both workflows is acceptable since they are independent CI jobs.

**`deployment/docker-build/pyaleph.dockerfile`** (line 51): Minor: consider adding a comment after ARG VERSION explaining its purpose, e.g., 'ARG VERSION # Version injected from CI (hatch version)'. This helps future maintainers understand why this ARG exists.

PR #1112, opened 2026-04-27
build: switch docker base image to python:3.12-slim-bookworm

odesenfans:
## Summary

Replace `ubuntu:24.04` + `deadsnakes/ppa` + manual `python3.12` install with the official `python:3.12-slim-bookworm` image, which already ships Python 3.12 (with headers and venv).

Removes from the Dockerfile:
- `software-properties-common` + `add-apt-repository ppa:deadsnakes/ppa`
- `python3.12`, `python3.12-dev`, `python3.12-venv` apt packages
- Adds `patch` as an explicit dep in the final stage (slim-bookworm doesn't include it; it was previously pulled in transitively by the Ubuntu base layer)

Local build: **754 MB → 558 MB (-196 MB)**.

## Stack

Based on `od/docker-slim-no-git` (PR #1111). One of four parallel follow-ups to that PR; merge order doesn't matter, but the others may need a trivial rebase if this lands first.

## Test plan

- [x] Local build succeeds with `--build-arg VERSION=$(hatch version)`
- [x] `importlib.metadata.version('pyaleph')` returns the build-arg value
- [x] `.git` absent from final image (inherited from #1111)
- [ ] CI build green
- [ ] Smoke-test the image actually starts (entrypoint runs)


foxpatch-aleph (dismissed):
The migration from ubuntu:24.04 + deadsnakes PPA to python:3.12-slim-bookworm is correct and well-executed. All Python references are consistently updated from python3.12 to python3. Build dependencies (libpq-dev, pkg-config, rust) are properly placed in the builder stage. Runtime libraries (libpq5, libsodium23) are in the final stage. The explicit addition of patch as a dependency is correct since slim images don't include it. The 196 MB size reduction is a meaningful improvement. No bugs or security issues found.

foxpatch-aleph (changes requested):
The Dockerfile change from ubuntu:24.04 + deadsnakes to python:3.12-slim-bookworm is straightforward and correct. However, this PR also includes two very large, unrelated documentation files (~1,500+ combined lines) for an "Authenticated IPFS Uploads" feature design and implementation plan. These should not be in this PR — they appear to be accidental checkins or belong in a separate PR. Additionally, libgmp-dev is installed in both the base and final stages; since the builder inherits from base, the final-stage install is redundant but harmless.

**`docs/superpowers/plans/2026-04-23-authenticated-ipfs-uploads.md`** (line 1): This 1,345-line implementation plan for "Authenticated IPFS Uploads" is unrelated to the Docker base image change. It should either be removed from this PR or in a separate PR. Committing a feature design/plan in the same PR as an infrastructure change makes the PR harder to review, bisect, and revert.

**`docs/superpowers/specs/2026-04-23-authenticated-ipfs-uploads-design.md`** (line 1): Same issue — this ~202-line design doc is unrelated to the Docker base image change. Remove from this PR.

**`deployment/docker-build/pyaleph.dockerfile`** (line 59): libgmp-dev is already installed in the base stage (line 9) and the final stage inherits from base. The redundant install here is harmless but unnecessary — it can be removed from the final-stage apt install.

foxpatch-aleph (approved):
The Dockerfile change to switch from ubuntu:24.04 + deadsnakes PPA to python:3.12-slim-bookworm is correct and well-reasoned. All python3.12 references are properly updated to python3. The explicit addition of 'patch' in the final stage is correct since the slim image doesn't include it transitively. The removal of build-only packages from the final stage is appropriate. No new bugs or security issues introduced.

**`deployment/docker-build/pyaleph.dockerfile`** (line 59): Nit (pre-existing, not introduced by this PR): libsodium-dev is installed in the final runtime image. This is a build dependency — only libsodium23 (the runtime shared library) and libgmp-dev should be needed. Consider removing libsodium-dev to shrink the image further in a follow-up.

**`deployment/docker-build/pyaleph.dockerfile`** (line 58): Nit (pre-existing): libgmp-dev is already installed in the 'base' stage (line 9) and is reinstalled here. Since the final stage starts from 'base', this install is redundant. The copy-only approach in the final stage could skip it.

PR #1114, opened 2026-04-27
build: stop running apt-get upgrade in dockerfile

odesenfans:
## Summary

`apt-get -y upgrade` pulls the latest version of every preinstalled package in `ubuntu:24.04` at build time, which makes the image size non-deterministic across builds: rebuilding the same git ref weeks apart can produce a materially different image because Ubuntu pushed package updates in between. This is one of the suspected contributors to the rc6 → rc7 size regression.

Drop both `apt-get -y upgrade` calls in the base stage. Specific packages we install via `apt-get install` will still be at the latest version (apt always installs the latest matching the indexes), but pre-installed packages in the `ubuntu:24.04` image stay at whatever version the base image ships.

Local build: **754 MB → 714 MB (-40 MB)**.

## Tradeoff

We give up automatically picking up Ubuntu security updates between base-image refreshes. The mitigation is to bump the base image when a new ubuntu:24.04 point is published (or pin to a digest and bump deliberately).

## Stack

## Test plan

- [x] Local build succeeds
- [x] `importlib.metadata.version('pyaleph')` returns expected version
- [ ] CI build green


foxpatch-aleph (dismissed):
The diff correctly removes both `apt-get -y upgrade` calls from the Dockerfile, which should produce deterministic, smaller images. The change is minimal and well-reasoned: removing upgrade preserves the latest versions for explicitly installed packages while preventing pre-installed Ubuntu packages from drifting between builds. The acknowledged tradeoff (manual base image bumps for security updates) is standard practice in container engineering. No issues found.

foxpatch-aleph (dismissed):
Clean, low-risk change that removes both `apt-get -y upgrade` calls from the base stage of the Dockerfile. This makes image builds deterministic by avoiding pulling in arbitrary package updates between base-image refreshes. `apt-get update` is retained before each `apt-get install`, so newly installed packages still get the latest repo version. The tradeoff of not auto-upgrading pre-installed packages is acknowledged and acceptable.

**`deployment/docker-build/pyaleph.dockerfile`** (line 5): The diff also adds two large, unrelated documentation files (`docs/superpowers/plans/2026-04-23-authenticated-ipfs-uploads.md` at 1345 lines and `docs/superpowers/specs/2026-04-23-authenticated-ipfs-uploads-design.md` at 202 lines). These don't affect correctness but should ideally be submitted as a separate PR to keep the change set focused.

foxpatch-aleph (dismissed):
The change is correct and the rationale is sound. Removing `apt-get upgrade` makes image builds reproducible by keeping pre-installed base-image packages at their shipped versions. `apt-get update` + `apt-get install` still installs the latest available versions for explicitly requested packages, so there's no functional regression. The security tradeoff (missing interim Ubuntu security updates for pre-installed packages) is acknowledged in the PR description with a reasonable mitigation (bump the base image tag when Ubuntu publishes a new point release). The diff is minimal and clean with no issues.

PR #1115, opened 2026-04-27
build: install git only in builder stage

odesenfans:
## Summary

`git` is needed at build time to clone the `aleph-p2p-client` git URL dependency, but not at runtime. Move it from the shared `base` stage into the `builder` stage so the final runtime image doesn't ship git.

Also explicitly install `patch` in the final stage: it was previously pulled in transitively as a Recommends of git, and is needed for the openssl.cnf legacy-providers patch step. Without this, removing git also removed patch and the build failed at the patch step.

Local build: **754 MB → 717 MB (-37 MB)**. Verified `which git` returns nothing and `dpkg -l git` shows the package is not installed in the final image.

## Stack

## Test plan

- [x] Local build succeeds
- [x] `git` not installed in the final image (`which git` empty, `dpkg -l git` reports nothing)
- [x] `aleph-p2p-client` git URL dep still installs in the builder stage
- [x] openssl.cnf patch step still works (patch installed explicitly)
- [ ] CI build green


foxpatch-aleph (dismissed):
The Dockerfile changes correctly move git from the shared base stage to the builder stage (where it's needed for cloning aleph-p2p-client) and explicitly install patch in the final stage (needed for the openssl.cnf patch step). The change is minimal, correct, and reduces the final image size by 37 MB while also slightly reducing the attack surface by removing git from the runtime image. The only minor observation is that libgmp-dev is installed in both the base and final stages, which is harmless (apt is idempotent) and appears intentional.

**`deployment/docker-build/pyaleph.dockerfile`** (line 66): libgmp-dev is installed in both the base stage (line 10) and the final runtime stage (line 66). This is harmless since apt-get install is idempotent, but worth noting — the base stage copy may be redundant if the final stage always installs it anyway.

foxpatch-aleph (commented):
Let me check if other Dockerfiles in the repo reference the base stage or expect git to be available.

foxpatch-aleph (dismissed):
The PR correctly moves git from the shared base stage to the builder-only stage, eliminating it from the final runtime image. The explicit patch dependency is added to the final stage to replace the transitive dependency that git was providing. The aleph-p2p-client git URL dependency (confirmed in pyproject.toml line 35) ensures git is needed during pip install in the builder. The openssl.cnf.patch at line 74 confirms patch is needed in the final stage. All changes are sound and align with the stated goal of reducing the runtime image size.

PR #1117, opened 2026-05-05
refactor: lift denormalized-column list to MessageDb.DENORMALIZED_COLUMNS

odesenfans:
## Summary

- Both `message_to_dict` (in `src/aleph/web/controllers/messages.py`) and `test_owners_filter_ws_logic` (in `tests/api/test_list_messages.py`) carry their own hardcoded copy of the columns to strip from a `MessageDb` payload before validating it against `aleph_message` schemas (which use `extra="forbid"`).
- Adding a new denormalized column requires remembering to update both lists; forgetting either leaks the column onto the wire (and breaks SDK clients that validate responses) or fails the WebSocket-filtering test.
- Lift the set onto the model as `MessageDb.DENORMALIZED_COLUMNS` (`ClassVar[FrozenSet[str]]`) so future denormalizations only need to extend a single source of truth. Behavior is unchanged.

## Test plan

- [ ] `hatch run testing:test tests/api/test_list_messages.py::test_owners_filter_ws_logic`
- [ ] `hatch run testing:test tests/api/test_list_messages.py -v`
- [ ] `hatch run linting:all`
- [ ] Spot-check `/messages.json` response shape unchanged (no denormalized columns leak)

### DIFF
--- a/src/aleph/db/models/messages.py
+++ b/src/aleph/db/models/messages.py
@@ -1,5 +1,5 @@
 import datetime as dt
-from typing import Any, Dict, List, Mapping, Optional, Type
+from typing import Any, ClassVar, Dict, FrozenSet, List, Mapping, Optional, Type
 
 from aleph_message.models import (
     AggregateContent,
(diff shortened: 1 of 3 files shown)


foxpatch-aleph (approved):
Clean, low-risk refactor that centralizes the denormalized-column list on the model as a single source of truth. All 10 column names in the new constant match the model's denormalized fields. Both usage sites (controller + test) are correctly updated. No behavior change.

PR #1118, opened 2026-05-11
Fix WS active-connection counter leak in MessageBroadcaster

aliel:
_on_message discarded dead clients from the in-memory set without
  decrementing pyaleph_ws_messages_connections_active. The handler's
  finally block then called remove(), which early-returned because
  the client was no longer in _clients — the Redis counter never
  came back down.

  Collect dead clients during the broadcast batch and route them
  through remove() so the decrement runs and the MQ consumer is
  torn down when the last client drops.

### DIFF
--- a/src/aleph/web/controllers/messages.py
+++ b/src/aleph/web/controllers/messages.py
@@ -244,6 +244,7 @@ class MessageBroadcaster:
         )
 
         # Send in batches to avoid event loop starvation with many clients
+        dead: List[_WsClient] = []
         for i in range(0, len(clients), _SEND_BATCH_SIZE):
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (approved):
Correctly fixes a WS active-connection counter leak by routing dead clients through the remove() method instead of a bare set discard. remove() properly decrements the Redis counter and cleans up the MQ consumer when the last client drops. The existing double-remove guard prevents a concurrent-disconnect race from causing a double-decrement. Clean, minimal, well-commented change.

**`src/aleph/web/controllers/messages.py`** (line 267): Consider adding a regression test for the counter leak: e.g., add two clients, close one, broadcast, then assert the counter is decremented exactly once.

PR #1120, opened 2026-05-11
feat: maintain eager running_balance on credit_balances writes

odesenfans:
## Summary
- Add `running_balance` column to `credit_balances`, maintained atomically by the three `credit_history` writers (distribution, expense, transfer) via a new `_apply_balance_deltas` UPSERT helper inside the existing `_bulk_insert_credit_history` chokepoint.
- The existing FIFO-derived `balance` column and the lazy recompute in `get_credit_balance` are left untouched, so the `/api/v0/addresses/{address}/balance` read-path semantics are preserved during this groundwork commit.
- Add `deployment/scripts/backfill_credit_balances_running_balance.py` to populate `running_balance` for existing addresses from `credit_history` (one transaction per address, idempotent via UPSERT).

## Motivation
The current balance endpoint triggers an O(N^2) Python FIFO recompute over every `credit_history` row for the address on every cache miss, which is why CLI lookups for VM-heavy addresses are slow. This PR is step 1 of the rework: eager-maintain a per-address running sum on every write, so a follow-up can switch the read path to an O(1) cache lookup once expiration handling is added.

Full design and migration order: `docs/superpowers/plans/2026-05-08-credit-balance-eager-cache.md`.

## Scope of this PR
- Migration `0057_c7d8e9f0a1b2_add_credit_balances_running_balance` adds the nullable `running_balance` BIGINT column.
- Model `AlephCreditBalanceDb` maps the new column.
- `_apply_balance_deltas(session, deltas)` performs a `pg_insert ... ON CONFLICT DO UPDATE` against `credit_balances`, summing into the existing `running_balance` (with `COALESCE(running_balance, 0)` to handle pre-backfill rows).
- The three writer functions accumulate per-address signed deltas while building their CSV rows and pass them to `_bulk_insert_credit_history`, which applies them in the same transaction as the history insert.

## Out of scope (next PR)
- Expiration cron / `next_expiration_at` column.
- Switching `get_credit_balance` to read from `running_balance`.
- Removing the FIFO recompute from the read path.
- Behaviour of `validate_credit_transfer_balance`.

## Test plan
- [ ] `tests/db/test_credit_balances.py` covers the new helper (insert, increment, negative delta, null initial, empty mapping, zero delta) and the eager updates from each writer (distribution, repeated-address, expense, dist-then-expense, non-whitelisted transfer, whitelisted transfer).
- [ ] `tests/db/test_backfill_credit_balances_running_balance.py` covers backfill from history, idempotence, and creation of `credit_balances` rows for addresses that don't yet have one.
- [ ] Pre-existing balance/credit/transfer tests should remain green (no behaviour change on read path).
- [ ] Manual: run the backfill script against staging once the migration is applied; verify the resulting `running_balance` matches what the existing FIFO recompute would produce for a sample of addresses.

### DIFF
--- a/src/aleph/db/accessors/balances.py
+++ b/src/aleph/db/accessors/balances.py
@@ -9,6 +9,7 @@ from typing import Any, Dict, List, Mapping, Optional, Sequence, Tuple
 
 from aleph_message.models import Chain
 from sqlalchemy import func, select, text
+from sqlalchemy.dialects.postgresql import insert as pg_insert
 from sqlalchemy.sql import Select
 
 from aleph.db.models import AlephBalanceDb, AlephCreditBalanceDb, AlephCreditHistoryDb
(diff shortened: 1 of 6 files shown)


foxpatch-aleph (dismissed):
Cleanly structured groundwork PR that eagerly maintains a running_balance column on credit_balances via a single UPSERT chokepoint. The three writer functions accumulate per-address deltas while building their CSV rows, and _apply_balance_deltas atomically applies them in the same transaction. Migration, model, backfill script, and 15+ new tests are thorough and well-organized. Read-path semantics are preserved (the existing FIFO-derived `balance` column is untouched), which is correctly scoped. No bugs or security issues — the design is sound and the implementation matches the spec.

**`src/aleph/db/accessors/balances.py`** (line 554): The `_apply_balance_deltas` UPSERT inserts rows with only `address` and `running_balance`, relying on the ORM default of `balance=0`. This is fine (the FIFO path handles `balance` independently) but worth being explicit about in the docstring: "New rows are created with `balance=0` — the eager path only maintains `running_balance`."

**`deployment/scripts/backfill_credit_balances_running_balance.py`** (line 54): The backfill `_upsert_running_balance` also sets `balance=0` for newly created rows. This is a placeholder that gets overwritten by the next FIFO recompute, but for the orphan-address case (credit_history exists but no credit_balances row) it means `balance=0` is temporarily stored. Consider adding a comment noting this is intentional and harmless.

**`deployment/migrations/versions/0057_c7d8e9f0a1b2_add_credit_balances_running_balance.py`** (line 1): Migration numbering is 0057, correctly following the previous 7e5a630e4b36. Clean and minimal -- just what's needed.

src/aleph/db/accessors/balances.py
```diff
@@ -550,16 +551,51 @@ def _format_csv_row(*fields) -> str:
     return output.getvalue().rstrip("\n")
 
 
+def _apply_balance_deltas(session: DbSession, deltas: Mapping[str, int]) -> None:
+    """Apply per-address signed deltas to credit_balances.running_balance atomically.
+
+    running_balance is the raw signed sum of all credit_history.amount rows for the
+    address, NOT FIFO-aware. The lazy FIFO recompute on the `balance` column remains
+    the source of truth for the API read path until the expiration cron lands.
+    See docs/superpowers/plans/2026-05-08-credit-balance-eager-cache.md.
```
odesenfans:
```suggestion
    the source of truth for the API read path until the expiration cron lands.
```

foxpatch-aleph (approved):
Well-structured PR that adds eager-maintained running_balance to credit_balances. The chokepoint design (accumulating per-address deltas in each writer, then atomically applying them via _bulk_insert_credit_history) is clean and correct. The UPSERT handles all edge cases (new rows, existing rows, NULL pre-backfill values). Test coverage is thorough across unit and integration levels. No logic errors, security issues, or race conditions found.

odesenfans:
Superseded by a redesign: per-(address, expiration_date) bucket cache, sort-by-expiration policy, background expiration task, repair-driven backfill, no read-path writes. Opening a new PR.

PR #1132, opened 2026-05-13
Speed up get_unconfirmed_messages with denormalized first_confirmed_at

odesenfans:
## Summary
- Replace the `NOT EXISTS` anti-join against `message_confirmations` + `chain_txs` in `get_unconfirmed_messages` with a direct check on `messages.first_confirmed_at IS NULL`. The column is already trigger-maintained (migration 0049), so no schema change is needed.
- Drop the now-unused `chain` parameter and update the two callers (`chains/ethereum.py`, `chains/nuls2.py`) accordingly.

## Performance
Measured against a production-sized dataset (~22k unconfirmed signed messages):

Before (anti-join): ~10s
After (denorm column, using existing `ix_messages_first_confirmed`): ~300ms cold, ~50ms warm. No new index needed.

```
Limit  (cost=26259.44..26260.69 rows=500 width=1275) (actual time=48.696..48.796 rows=500 loops=1)
  ->  Sort  (Sort Key: reception_time, Sort Method: top-N heapsort  Memory: 738kB)
        ->  Bitmap Heap Scan on messages (Recheck Cond: first_confirmed_at IS NULL, Filter: signature IS NOT NULL)
              ->  Bitmap Index Scan on ix_messages_first_confirmed (Index Cond: first_confirmed_at IS NULL)
Execution Time: 48.856 ms
```

## Semantic note
The previous implementation filtered confirmations by chain (e.g. `chain=Chain.ETH` for the ETH packer). `first_confirmed_at` is set by a confirmation on any chain. In production only ETH confirmations are recorded, so the behavior is equivalent today. If multi-chain confirmations are reintroduced, the unconfirmed query will need to reconsider this.

## Test plan
- [ ] `hatch run testing:test tests/db/test_messages.py::test_get_unconfirmed_messages tests/db/test_messages.py::test_get_unconfirmed_messages_trusted_messages -v`
- [ ] Spot-check ETH and NULS2 packer paths still call `get_unconfirmed_messages` without `chain=`.

### DIFF
--- a/src/aleph/chains/ethereum.py
+++ b/src/aleph/chains/ethereum.py
@@ -411,7 +411,6 @@ class EthereumConnector(ChainWriter):
                                 session=session,
                                 limit=500,
                                 offset=offset,
-                                chain=Chain.ETH,
                             )
                         )
                         if not batch:
--- a/src/aleph/chains/nuls2.py
+++ b/src/aleph/chains/nuls2.py
@@ -208,7 +208,7 @@ class Nuls2Connector(ChainWriter):
                 while True:
                     batch = list(
                         get_unconfirmed_messages(
-                            session=session, limit=500, offset=offset, chain=Chain.NULS2
+                            session=session, limit=500, offset=offset
                         )
                     )
                     if not batch:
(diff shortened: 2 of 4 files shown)


foxpatch-aleph (commented):

foxpatch-aleph (approved):
Clean, well-motivated refactor replacing a slow anti-join with a direct IS NULL check on the trigger-maintained first_confirmed_at column. Removes the now-unnecessary chain parameter and handles both callers. The semantic change (any chain's confirmation now counts) is actually more correct than the old chain-filtered behavior.

**`src/aleph/db/accessors/messages.py`** (line 421): Excellent simplification. The old branching logic between chain=None and chain=<value> was a source of complexity for a feature (per-chain unconfirmed queries) that had no real use case.

PR #1137, opened 2026-05-14
fix: auto-reject PROGRAM messages with non-dict metadata

odesenfans:
## Summary

Some PROGRAM messages slipped past validation while `ExecutableContent.metadata` accepted lists. The current validator requires a dict, so reading those rows fails `parsed_content` and surfaces as 500s on `GET /api/v0/messages/<hash>` (ex: [42a4a8...3d96f3](https://api2.aleph.im/api/v0/messages/42a4a8567408240fbe089d77dccd068665fb5aa965d42ec19706e294f83d96f3) returns 500, while [the same hash on epyc](https://epyc.bearmetaldev.com/api/v0/messages/42a4a8567408240fbe089d77dccd068665fb5aa965d42ec19706e294f83d96f3) properly reports the message as rejected).

This change:

- Adds `mark_processed_message_as_rejected` in `aleph.repair`. It mirrors `mark_pending_message_as_rejected` but starts from a `MessageDb` row instead of a `PendingMessageDb`: cleans up VM rows for program/instance, upserts `rejected_messages`, flips `message_status` to REJECTED, and deletes the `messages` row. The trigger keeps `message_counts` consistent; FK cascades clean `message_confirmations` and `account_costs`.
- Adds `_reject_invalid_program_metadata` and wires it into `repair_node` so the API rejects affected PROGRAM messages on every startup. The query uses `jsonb_typeof(content->'metadata') = 'array'`; an empty result is a no-op.
- Ships `deployment/scripts/reject_processed_messages.py` for ad-hoc cleanups when a restart is not an option. Dry-run by default, `--commit` to persist; targets specific hashes via `--hash` / `--hashes-file`. Runs from inside the API container against the deployed config at `/var/pyaleph/config.yml`.

## Test plan

- [x] `venv/bin/python -m pytest tests/test_repair.py -v` — 5 tests, all pass (rejects list metadata, preserves dict/None metadata, ignores non-program types, no-op on empty DB).
- [x] `venv/bin/python -m pytest tests/db/test_messages.py tests/db/test_credit_balances.py` — adjacent suites still pass (63 total).
- [x] `venv/bin/ruff check` + `black` + `isort` clean on changed files.
- [ ] Manual: on a staging snapshot, confirm targeted hashes flip from PROCESSED to REJECTED and `GET /messages/<hash>` no longer 500s.

### DIFF
--- a/deployment/scripts/reject_processed_messages.py
+++ b/deployment/scripts/reject_processed_messages.py
@@ -1,0 +1,205 @@
+#!/usr/bin/env python3
+"""Mark processed messages as rejected.
+
+Use when a message was accepted under permissive validation rules that have
+since become stricter (ex: BaseExecutableContent.metadata now requires a dict
+and rejects lists, but some nodes accepted such messages historically). The
+API returns 500 on those messages because parsed_content access raises; moving
+them to the rejected state matches what nodes that rejected them in the first
+place expose to clients.
+
+The actual rejection logic lives in `aleph.repair.mark_processed_message_as_rejected`
+and is also wired into `repair_node`, which runs at startup. This script
+exists for ad-hoc cleanups when you have a known list of hashes and don't
+want to wait for the next restart.
+
+Runs as a dry-run by default. Pass --commit to actually persist changes.
+Hashes can be provided via repeated --hash flags or via --hashes-file (one
+hash per line, lines starting with # are skipped).
+"""
+
+import argparse
+import logging
+import sys
+from pathlib import Path
+from typing import Iterable, List
+
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent.parent / "src"))
+
+from sqlalchemy import select  # noqa: E402
+
+import aleph.config  # noqa: E402
+from aleph.db.connection import make_engine, make_session_factory  # noqa: E402
+from aleph.db.models.messages import MessageDb  # noqa: E402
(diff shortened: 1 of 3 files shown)


foxpatch-aleph (dismissed):
Clean, well-structured fix for a production bug where PROGRAM messages with list-typed metadata cause 500s via parsed_content. Implements a reusable rejection utility for processed messages, wires a repair function into startup, and ships a companion CLI script. Thorough test coverage and good code quality throughout.

**`src/aleph/repair.py`** (line 69): Consider using session.execute(delete_vm_updates(...)) instead of _ = list(...) to avoid loading results into memory and make the intent clearer. The list() is needed to force execution, but a comment explaining why would help maintainers.

foxpatch-aleph (approved):
A well-structured fix for PROGRAM messages with invalid list-typed metadata that cause 500 errors. The approach is correct: repair_node rejects them at startup, a standalone script handles ad-hoc cases, and both properly clean up VM rows and cascade to account_costs via FK. The race condition between the initial query and per-hash processing is properly handled with a re-check. Tests cover the main scenarios (list/dict/null metadata, non-PROGRAM types, empty DB). No bugs or security issues found.

**`deployment/scripts/reject_processed_messages.py`** (line 256): Minor: `changed` is incremented in both the `--commit` path and the dry-run path, so the summary count is not strictly 'changed' in the commit sense. Consider using two separate counters or mentioning 'processed' in the count label.

PR #1140, opened 2026-05-18
fix(jobs): track and terminate child processes via JobsRunner

odesenfans:
Part **3/8** of the clean-shutdown audit. Closes **Finding 1**.

## Problem

`start_jobs` previously created three `multiprocessing.Process` instances and dropped the handles on the floor. The parent had no way to `.terminate()` or `.join()` them. On shutdown the OS reaped them after the orchestrator grace timer expired.

## Fix

- New `JobsRunner` dataclass in `aleph.jobs` carrying both the spawned `Process` handles and the inline coroutines.
- `start_jobs(...) -> JobsRunner` (was `List[Coroutine]`).
- New `async def stop_jobs(runner, terminate_timeout=10.0)` sends SIGTERM to each alive child, joins them in parallel via `asyncio.gather` (worst case `terminate_timeout + 1.0` regardless of child count, well within k8s/Docker default grace periods), and SIGKILLs anyone that doesn't comply.
- `commands.main()` always initialises `runner = JobsRunner()` and registers `stack.push_async_callback(stop_jobs, runner)` unconditionally — `stop_jobs` early-returns on an empty runner.

## Tests

3 new tests in `tests/jobs/test_jobs_runner.py` covering cooperative SIGTERM, stubborn-child SIGKILL fallback (with elapsed-time assertion), and empty-runner no-op. Tests use a `multiprocessing.Event` for child-readiness synchronisation (no `sleep`-based races) and `mp.get_context(\"fork\")` for test speed.

---

Base: `fix/shutdown-main-process-signals`. Merge after the previous PR in this chain.

### DIFF
--- a/src/aleph/commands.py
+++ b/src/aleph/commands.py
@@ -29,7 +29,7 @@ from aleph.chains.connector import ChainConnector
 from aleph.cli.args import parse_args
 from aleph.db.connection import make_db_url, make_engine, make_session_factory
 from aleph.exceptions import InvalidConfigException, KeyNotFoundException
-from aleph.jobs import start_jobs
+from aleph.jobs import JobsRunner, start_jobs
 from aleph.jobs.cron.balance_job import BalanceCronJob
 from aleph.jobs.cron.credit_balance_job import CreditBalanceCronJob
 from aleph.jobs.cron.cron_job import CronJob, cron_job_task
@@ -187,14 +187,17 @@ async def main(args: List[str]) -> None:
 
         tasks: List[Coroutine] = []
 
+        runner = JobsRunner()
         if not args.no_jobs:
             LOGGER.debug("Creating jobs")
-            tasks += start_jobs(
+            runner = start_jobs(
                 config=config,
                 session_factory=session_factory,
                 ipfs_service=ipfs_service,
                 use_processes=True,
             )
+            tasks += runner.tasks
+        stack.push_async_callback(runner.stop)
 
         LOGGER.debug("Initializing p2p")
         p2p_client, p2p_tasks = await p2p.init_p2p(
--- a/src/aleph/jobs/__init__.py
+++ b/src/aleph/jobs/__init__.py
@@ -1,4 +1,6 @@
+import asyncio
 import logging
+from dataclasses import dataclass, field
 from multiprocessing import Process
 from typing import Coroutine, List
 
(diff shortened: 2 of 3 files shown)


foxpatch-aleph (commented):

foxpatch-aleph (approved):
Cleanly fixes orphaned child processes by introducing a JobsRunner dataclass that tracks Process handles, and a stop_jobs function that implements a two-phase termination (SIGTERM with timeout, fallback to SIGKILL). Tests cover all three paths: graceful exit, stubborn-child kill, and empty-runner no-op. The approach integrates well with the existing AsyncExitStack-based lifecycle.

**`tests/jobs/test_jobs_runner.py`** (line 34): Tests use mp.get_context("fork") for speed, but the main process uses set_start_method("spawn"). The tests exercise stop_jobs in isolation rather than the full start_jobs -> stop_jobs pipeline. Consider adding a test that uses "spawn" context for parity with production, or document why fork-only coverage is sufficient.

**`tests/jobs/test_jobs_runner.py`** (line 64): elapsed < 3.0 is generous for terminate_timeout=0.5 + 1.0s kill join. Consider tightening to 2.0 to catch performance regressions.

foxpatch-aleph (dismissed):
Clean PR that wraps background subprocess handles in a `JobsRunner` dataclass and adds a `stop_jobs` async function for orderly SIGTERM-then-SIGKILL shutdown. The logic is correct: `stop_jobs` handles the empty-runner case, uses `asyncio.gather` for parallel joins, and falls back to SIGKILL after timeout. Tests cover cooperative exit, stubborn-child killing (with elapsed-time assertion), and empty-runner safety. The changes are minimal and focused — no security issues, race conditions, or logic errors found.

foxpatch-aleph (dismissed):
Clean implementation of child process lifecycle management for JobsRunner. The two-phase termination (SIGTERM with timeout, then SIGKILL) with parallel joins correctly bounds wall-clock shutdown time regardless of child process count. Tests are race-condition-free using mp.Event synchronization. Code is clear and well-documented.

**`src/aleph/jobs/__init__.py`** (line 61): The docstring's claim that total wall time is bounded by "terminate_timeout + 1.0" should note this excludes the executor thread scheduling overhead, though in practice this is negligible.

**`src/aleph/commands.py`** (line 200): Consider a comment here explaining that stop_jobs is registered unconditionally (even with no_jobs) because it early-returns on an empty runner.

foxpatch-aleph (approved):
This is a well-structured fix that addresses the core problem (untracked subprocess handles) cleanly. The JobsRunner dataclass encapsulates both processes and tasks, and the stop() method implements a proper two-phase termination with SIGTERM→join→SIGKILL fallback. The code is correct, tests cover the three key scenarios (cooperative shutdown, stubborn child kill, empty runner no-op), and the design integrates naturally with the existing AsyncExitStack-based lifecycle.

**`src/aleph/jobs/__init__.py`** (line 22): Consider using `Sequence[Process]` instead of `List[Process]` if mutation after construction isn't needed — but `List` is fine given the `field(default_factory=list)` mutability pattern.

**`src/aleph/commands.py`** (line 190): Minor style nit: the `runner = JobsRunner()` on this line is immediately overwritten by `start_jobs()` when jobs are enabled. It only serves the `args.no_jobs` path. Could be restructured to avoid the double-assignment, but harmless.

PR #1138, opened 2026-05-18
fix(jobs): handle SIGTERM cleanly in subprocess entry points

odesenfans:
Part **1/8** of the clean-shutdown audit. Closes **Finding 8**.

## Problem

The three jobs subprocesses (`fetch_pending_messages_subprocess`, `pending_messages_subprocess`, `pending_txs_subprocess`) only catch `KeyboardInterrupt`. A parent-initiated `Process.terminate()` (SIGTERM) is unhandled, so they exit abruptly without closing RabbitMQ / SQLAlchemy / HTTP resources.

## Fix

- New module `aleph.toolkit.lifecycle` with two helpers (`install_signal_handlers`, `safe_async_cleanup`) used by this PR and subsequent ones in the series.
- Each subprocess wraps its root coroutine in a task, installs SIGTERM/SIGINT handlers that call `task.cancel()`, and catches `asyncio.CancelledError` so the existing async-context `finally` blocks run cleanly.

## Tests

3 new tests in `tests/toolkit/test_lifecycle.py` (signal handler invocation + `safe_async_cleanup` exception swallowing + completion).

---

Base: `main`. Subsequent PRs in this chain must be merged in order (1→2→…→8).

PR #1146, opened 2026-05-18
refactor(jobs): unify subprocess entry points on asyncio.run

odesenfans:
Follow-up to #1138 addressing the consistency nit raised in review.

## Problem

After #1138, the three jobs subprocesses installed SIGTERM handlers via two different idioms:

- `fetch_pending_messages_subprocess` used `asyncio.run(_runner())` with an inner coroutine because `asyncio.run` requires handler installation to happen on the running loop.
- `pending_messages_subprocess` and `pending_txs_subprocess` used `loop.create_task(...) + install_signal_handlers(loop, ...) + loop.run_until_complete(task)` inline because they had a pre-built loop from `prepare_loop()`.

Both work, but the asymmetry is needless noise for a future reader.

## Fix

All three subprocesses now share the same shape:

```python
config = prepare_config(config_values)
...
async def _runner():
    task = asyncio.create_task(<job_task>(config))
    install_signal_handlers(asyncio.get_running_loop(), task.cancel)
    try:
        await task
    except asyncio.CancelledError:
        LOGGER.info("<name> subprocess cancelled by signal")

try:
    asyncio.run(_runner())
except KeyboardInterrupt:
    ...
```

`prepare_loop()` is renamed to `prepare_config()` since it no longer creates the loop. This also removes the `asyncio.get_event_loop()` call which is deprecated in Python 3.12+ when no loop is running and emits a DeprecationWarning. `Tuple` is dropped from the `job_utils` typing imports along with the now-stale Motor docstring.

## Tests

No new tests. The 3 existing `tests/toolkit/test_lifecycle.py` tests still pass; behavior is unchanged across all three subprocesses.

### DIFF
--- a/src/aleph/jobs/fetch_pending_messages.py
+++ b/src/aleph/jobs/fetch_pending_messages.py
@@ -31,7 +31,7 @@ from aleph.toolkit.timestamp import utc_now
 from aleph.types.db_session import DbSessionFactory
 
 from ..toolkit.rabbitmq import make_mq_conn
-from .job_utils import MessageJob, make_pending_message_queue, prepare_loop
+from .job_utils import MessageJob, make_pending_message_queue, prepare_config
 
 LOGGER = getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


foxpatch-aleph (approved):
Clean refactoring that unifies three subprocess entry points under a single `asyncio.run(_runner())` pattern. The removal of `prepare_loop()` eliminates a deprecated `asyncio.get_event_loop()` call and the asymmetry between `fetch_pending_messages` (which already used `asyncio.run`) and the other two subprocesses. Signal handlers are installed at the same logical point (inside the running loop, before `await task`), so behavior is preserved. No logic, security, or testing concerns.

PR #1149, opened 2026-05-19
fix(api): enforce balance check on all authenticated uploads

odesenfans:
## Summary

`_verify_user_balance` had two independent bugs that together let underfunded uploads through the API and into the message pipeline:

1. **Size-threshold gate.** The check short-circuited whenever the file size sat at or below `max_unauthenticated_upload_file_size` (25 MiB by default). The gate was meant to keep anonymous uploads free; it accidentally applied to authenticated uploads too. For any STORE attached to a sub-25 MiB upload the API returned success without checking that the sender could pay.
2. **Hold-only balance math.** Even when the gate did fire, the check called `get_total_balance` (ALEPH tokens) and `get_total_cost_for_address` with the default `payment_type=PaymentType.hold`, but `get_total_and_detailed_costs` returned the message cost in whatever payment type the content declared. For a credit-paid STORE that compared ALEPH holdings against credit costs, which is incoherent both ways: users with credits but no ALEPH were falsely rejected, and users with ALEPH but no credits were falsely accepted (the message then died downstream).

This PR makes the API-level check unconditional whenever a STORE message accompanies the upload, and routes it through `validate_balance_for_payment`, the same helper `StoreMessageHandler.check_balance` already uses. That helper branches on the content's payment type, reads the right balance and cost columns, and enforces the 1-day minimum runtime requirement on the credit branch (credit storage cost is a per-second rate, mirrored from the pipeline's own model). The internal `InsufficientBalanceException` / `InsufficientCreditException` are translated to HTTP 402 at the controller boundary.

In `ipfs_add_file` the balance check also moves above `ipfs_service.add_bytes`, so an underfunded request leaves no pin on the IPFS daemon. The CID match check stays post-pin since it depends on the daemon's output.

The unauthenticated size cap remains as the entry gate for the anonymous path, but it no longer doubles as a "skip balance" condition. The API is now stricter than the message pipeline (which still has a legacy `PaymentType.hold + small file` carve-out); that asymmetry is acceptable while hold payments are phased out.

## Changes

- `storage.py`: drop the size-threshold gate in `_verify_user_balance`; rewrite the body to delegate to `validate_balance_for_payment` and translate internal exceptions to `HTTPPaymentRequired`. Drop the now-unused `max_unauthenticated_upload_file_size` plumbing in `_check_and_add_file` and `storage_add_file`.
- `ipfs.py`: move balance check to before `ipfs_service.add_bytes`.

## Test plan

- [x] `tests/api/test_ipfs.py`: strict pre-pin balance test asserts 402 + `add_bytes` not called + no file row + no grace period. `cid_mismatch` and `stat_timeout` now seed a balance row since they previously rode on the size-threshold skip. New tests cover credit-paid STORE both ways: zero credits with large ALEPH balance returns 402 with no pin; sufficient credits returns 200.
- [x] `tests/api/test_storage.py`: the `balance=0` parametrize case for `test_storage_add_file_with_message` now expects 402.
- [x] Full `tests/api/` suite: 177 passed, 6 skipped.
- [x] `ruff` + `black` clean.

### DIFF
--- a/src/aleph/web/controllers/ipfs.py
+++ b/src/aleph/web/controllers/ipfs.py
@@ -131,11 +131,11 @@ async def ipfs_add_file(request: web.Request):
                 max_size=max_unauthenticated_upload_file_size,
             )
 
-        # Validate the signed message BEFORE pinning so a bad signature
-        # cannot leave an orphan pin on the IPFS daemon. Note: by this
-        # point the file is already buffered to a temp file (multipart
-        # parts are consumed in arrival order); we gate the pin step,
-        # not the multipart read.
+        # Validate the signed message and balance BEFORE pinning so neither
+        # a bad signature nor an underfunded account leaves an orphan pin on
+        # the IPFS daemon. By this point the file is already buffered to a
+        # temp file (multipart parts are consumed in arrival order); we gate
+        # the pin step, not the multipart read.
         message = None
         message_content = None
         sync = False
@@ -173,7 +173,11 @@ async def ipfs_add_file(request: web.Request):
                     )
                 )
 
-        # Pin to IPFS — side effect: file is now on the local IPFS node.
+            message_content.estimated_size_mib = math.ceil(uploaded_file.size / MiB)
+            with session_factory() as session:
+                _verify_user_balance(session=session, content=message_content)
+
+        # Pin to IPFS, side effect: file is now on the local IPFS node.
         temp_file = await uploaded_file.open_temp_file()
         file_content = await temp_file.read()
         if isinstance(file_content, str):
@@ -181,7 +185,7 @@ async def ipfs_add_file(request: web.Request):
 
         cid = await ipfs_service.add_bytes(file_content)
 
-        # Post-pin: stat, CID match, balance check, persist.
+        # Post-pin: stat, CID match, persist.
         # Failures from this point on must leave the pin covered by the
         # 24 h grace period so the GC doesn't strand it.
         try:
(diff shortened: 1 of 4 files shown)


foxpatch-aleph (dismissed):
This PR fixes a security issue where authenticated uploads with STORE messages for files under 25 MiB bypassed the balance check, allowing credit-paid storage without sufficient balance. The fix correctly removes the size-threshold gate from _verify_user_balance, moves the IPFS balance check before ipfs_service.add_bytes to prevent orphan pins, and updates tests to validate the new behavior. The code changes are minimal, well-structured, and correctly scoped. Tests properly cover the fix with assertions on HTTP status codes, mock call counts, and database state.

**`src/aleph/web/controllers/ipfs.py`** (line 176): Balance check correctly placed before add_bytes (line 186). The CID match check (line 201) correctly stays after pinning since it depends on the daemon's output. The except Exception handler (line 221) correctly only covers post-pin failures — the pre-pin balance check properly short-circuits before reaching that handler.

**`src/aleph/web/controllers/storage.py`** (line 153): Function signature changed correctly: removed max_unauthenticated_upload_file_size param that was only used in the now-deleted size-threshold gate. The function is now truly unconditional when called.

**`tests/api/test_ipfs.py`** (line 355): Key test: replaces the old test that validated the buggy behavior. Now asserts 402 + add_bytes.not_called + no file row + no grace period with a tiny (34 byte) file, confirming size no longer gates the balance check.

**`tests/api/test_ipfs.py`** (line 305): Appropriate seeding of balance row for cid_mismatch and stat_timeout tests — these previously rode on the size-threshold skip and now need sufficient balance to reach their actual test targets.

**`tests/api/test_storage.py`** (line 299): Correctly changed balance=0 parametrize entry from 200 to 402 — mirrors the storage-side behavioral change.

foxpatch-aleph (approved):
This PR fixes a security bug where authenticated uploads with small files (<25 MiB) bypassed the balance check entirely, and where credit-paid STORE messages were checked against ALEPH token balances (incoherent). The fix removes the size-threshold gate and routes the check through validate_balance_for_payment, which correctly handles both hold and credit payment types. The IPFS controller also moves the balance check before the IPFS pin to prevent orphan pins. Tests comprehensively cover the fix: pre-pin 402 rejection for both ALEPH and credit payments, correct CID mismatch handling, and grace period behavior. The code is clean, the delegation to shared validation logic is the right approach, and all tests pass.

PR #1145, opened 2026-05-18
fix(api): close NodeCache and IpfsService on shutdown

odesenfans:
Part **8/8** of the clean-shutdown audit. Closes **Finding 6**.

## Problem

`api_entrypoint.py` creates `node_cache` and `ipfs_service` eagerly but never closes them. Two `# TODO` comments already document this gap.

## Fix

- Append `safe_async_cleanup(\"ipfs service\", ipfs_service.close())` and `safe_async_cleanup(\"node cache\", node_cache.close())` to `_on_cleanup`, between the `p2p client` close and `close_sessions`.
- Remove the two TODO comments.

After this PR (the last in the series), shutdown of the API process closes: broadcasters → mq_channel → p2p_client (which closes the shared mq_conn) → ipfs_service → node_cache → P2P HTTP sessions → engine.dispose(). The main process closes: p2p_client → P2P HTTP sessions → stop_jobs (children) → ipfs_service → node_cache → mq_channel → mq_conn → engine.dispose() (LIFO via AsyncExitStack).

## Tests

No new tests (wiring change). The series is end-to-end validated via the manual smoke test in `docs/superpowers/plans/2026-05-18-clean-shutdown.md` Appendix A.

---

Base: `fix/shutdown-close-p2p-client`. **Final PR in the series.**

PR #1139, opened 2026-05-18
fix(commands): handle SIGTERM via AsyncExitStack scaffolding

odesenfans:
Part **2/8** of the clean-shutdown audit. Closes **Finding 2**.

## Problem

`asyncio.run(main(...))` has no SIGTERM handler. SIGTERM (sent by `docker stop`, `systemctl stop`, kubernetes pod shutdown) does NOT raise `KeyboardInterrupt`, so the event loop keeps running until the orchestrator's grace period expires and the process is SIGKILLed.

## Fix

- New `_signal_aware_main(args)` wraps `main(args)` in an `asyncio.create_task`, registers SIGTERM/SIGINT handlers that call `task.cancel()`, and is awaited from `run()` via `asyncio.run(_signal_aware_main(...))`.
- `main()` body refactored to use `contextlib.AsyncExitStack`. `node_cache` and `ipfs_service` are entered via `stack.enter_async_context(...)`. `set_start_method(\"spawn\")` is moved above the stack and guarded with `get_start_method(allow_none=True) != \"spawn\"` so calling `main()` twice in a test harness doesn't `RuntimeError`.
- No `stack.push_async_callback(...)` calls yet — those are added incrementally by PRs 3-7.

After this PR, SIGTERM/SIGINT cancel the gather; existing `node_cache` / `ipfs_service` async-context exits run. No new resources closed yet beyond what was already in scope.

## Tests

No new tests (structural refactor; integration covered by the manual smoke test in the plan).

---

Base: `fix/shutdown-subprocess-signals` (#1138). Merge after #1138.

PR #1154, opened 2026-05-19
fix: silence shutdown noise from MqWatcher and Web3 provider

odesenfans:
## Summary

Two small follow-ups to the earlier shutdown-cleanup PRs (#1139 - #1145) that eliminate two remaining classes of noisy log output observed when Ctrl-C'ing pyaleph during a fresh sync.

### 1. MqWatcher prefetched-message warnings

Every Ctrl-C produced a flurry of:

```
[WARNING] aio_pika.queue: Message IncomingMessage:{...} lost for consumer with no_ack <RobustQueueIterator: ...>
```

`MqWatcher._check_for_message` opened the queue iterator with `no_ack=True` even though it only uses messages as level-triggered signals (the actual data lives in `pending_txs` / `pending_messages`). aio_pika unconditionally logs a warning per prefetched message still in the iterator buffer at close time when `no_ack=True`.

Switched to `no_ack=False` with explicit `await message.ack()`. Buffered messages are now nack'd with `requeue=True` on shutdown, so nothing is lost and the warning disappears.

### 2. Unclosed aiohttp session from the ETH Web3 provider

After SIGTERM:

```
[ERROR] asyncio: Unclosed client session
[ERROR] asyncio: Unclosed connector
```

`web3.AsyncHTTPProvider` caches an `aiohttp.ClientSession` per endpoint in its `HTTPSessionManager`. The provider exposes `disconnect()` to close it, but the only call site was inside `EthereumConnector.packer`'s `finally`. Reader-only nodes (no `packing_node`) never reach it, so the session leaks at shutdown.

- `EthereumConnector` is now an async context manager whose `__aexit__` calls `provider.disconnect()`.
- `ChainConnector` is also an async context manager. It owns an internal `AsyncExitStack` and enters every chain connector that supports the protocol, dedup'd by `id()` since `EthereumConnector` is registered as both reader and writer.
- `commands.py` registers `chain_connector` with the main `AsyncExitStack`, alongside the other shutdown callbacks added in the previous PRs.
- The duplicate `provider.disconnect()` in `packer`'s `finally` is removed (cleanup now lives on the context manager).

## Test plan

- [x] `hatch run testing:test tests/chains tests/message_processing` (104 passed, 3 skipped)
- [x] black / isort / ruff clean on the four modified files
- [ ] Local smoke test: `just start-dev-env && just run`, wait for ETH messages to arrive, Ctrl-C, confirm the two warning families are gone

### DIFF
--- a/src/aleph/chains/connector.py
+++ b/src/aleph/chains/connector.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+from contextlib import AsyncExitStack
 from typing import Dict, Self, Union
 
 from aleph_message.models import Chain
@@ -39,6 +40,23 @@ class ChainConnector:
 
         self.readers = {}
         self.writers = {}
+        self._exit_stack = AsyncExitStack()
+
+    async def __aenter__(self) -> Self:
+        await self._exit_stack.__aenter__()
+        # A connector can be both a reader and a writer; dedupe by identity so
+        # __aexit__ runs at most once per instance.
+        seen: set[int] = set()
+        for connector in (*self.readers.values(), *self.writers.values()):
+            if id(connector) in seen:
+                continue
+            seen.add(id(connector))
+            if hasattr(connector, "__aexit__"):
+                await self._exit_stack.enter_async_context(connector)
+        return self
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb) -> None:
+        await self._exit_stack.__aexit__(exc_type, exc_val, exc_tb)
 
     @classmethod
     async def new(
--- a/src/aleph/chains/connector.py
+++ b/src/aleph/chains/connector.py
@@ -1,6 +1,6 @@
 import asyncio
 import logging
-from contextlib import AsyncExitStack
+from contextlib import AbstractAsyncContextManager, AsyncExitStack
 from typing import Dict, Self, Union
 
 from aleph_message.models import Chain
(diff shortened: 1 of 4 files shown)


foxpatch-aleph (dismissed):
Clean follow-up PR that eliminates two sources of shutdown noise: MqWatcher warning from aio_pika on prefetched unacked messages, and unclosed aiohttp session from the Web3 provider. Both fixes are correct and minimal. The ChainConnector async context manager properly deduplicates by identity and skips connectors without cleanup needs. No logic errors or regressions found.

**`src/aleph/chains/connector.py`** (line 50): The `id()` dedup is safe here since connectors are long-lived singletons, but worth noting that this relies on object identity stability. If a connector were ever recreated with a new identity but the same logical instance, the dedup could break. Consider using the `connector` object itself as a dict key instead of `id(connector)` for slightly more robustness, though this is a nit.

foxpatch-aleph (approved):
Well-targeted PR that fixes two sources of noisy shutdown warnings. The MqWatcher change (no_ack=False with explicit ack) correctly avoids aio_pika's 'message lost' warning while preserving message delivery via requeue. The EthereumConnector/ChainConnector async context manager refactor ensures the aiohttp ClientSession is properly closed on shutdown even for reader-only nodes, removing the 'Unclosed client session' error. The deduplication by id() in ChainConnector.__aenter__ correctly handles connectors registered as both reader and writer. The removal of the redundant provider.disconnect() from packer's finally block is a good cleanup since the context manager handles it now. All changes are well-reasoned and correctly implemented.

PR #1155, opened 2026-05-20
typing: narrow JSONB content fields and credit-handler signatures

odesenfans:
Item (2) of the type-hint audit. Two related tightening passes that remove `Any` from places where the real type is known.

## JSONB content columns

`MessageDb.content`, `AggregateElementDb.content`, and `AggregateDb.content` were declared `Mapped[Any]` while the underlying JSONB columns store dict shapes at the top level in every Aleph message variant. Narrow them to `Mapped[Dict[str, Any]]`. Callers that read `.content` now see a dict statically, eliminating an entire category of accidental scalar-access bugs.

## Credit-handler signatures

`update_credit_balances_distribution`, `update_credit_balances_expense`, and `update_credit_balances_transfer` each took `content: Mapping[str, Any]` and immediately ran `CreditDistributionContent.model_validate(content)` (and friends). The signature claimed an untyped dict, but the body required a specific Pydantic shape and raised `InvalidMessageFormat` on mismatch. The signature was lying.

Move the parsing into the `PostMessageHandler.process_post` dispatcher via a small `_parse_credit_content` helper (TypeVar-generic, raises `InvalidMessageFormat` with a kind-tagged message). The three handler signatures now take the parsed model directly. This:

- removes three try/except blocks of duplicate error handling
- makes the handlers self-documenting via their parameter type
- centralizes parsing in the dispatcher where the dispatch decision is made

`update_balances` (the non-credit balance updater) does raw dict access and never parses into a Pydantic model, so it's left as `Mapping[str, Any]`.

## Out of scope

The original audit also flagged `chain_data_service.py`'s `list[dict[str, Any]]` returns. On closer inspection, only the smart-contract protocol path normalizes the dict shape; the on-chain and off-chain protocol paths pass arbitrary JSON through. Introducing a `TxMessageDict` TypedDict honestly there requires normalizing those legacy paths too, which is a larger refactor. Deferred to a follow-up.

## Test plan

- [x] `venv/bin/python -m pytest tests/message_processing/test_process_posts.py tests/db/test_credit_balances.py tests/balances/test_balances.py -q` — 60 passed
- [x] `venv/bin/isort` + `venv/bin/black` clean on changed files
- [x] Smoke import: `from aleph.handlers.content.post import update_credit_balances_distribution, _parse_credit_content` resolves; signature is `(session, content: CreditDistributionContent, message_hash, message_timestamp) -> None`

### DIFF
--- a/src/aleph/db/models/aggregates.py
+++ b/src/aleph/db/models/aggregates.py
@@ -1,5 +1,5 @@
 import datetime as dt
-from typing import Any
+from typing import Any, Dict
 
 from sqlalchemy import TIMESTAMP, Boolean, ForeignKey, Index, String
 from sqlalchemy.dialects.postgresql import JSONB
@@ -22,7 +22,7 @@ class AggregateElementDb(Base):
     item_hash: Mapped[str] = mapped_column(String, primary_key=True)
     key: Mapped[str] = mapped_column(String, nullable=False)
     owner: Mapped[str] = mapped_column(String, nullable=False)
-    content: Mapped[Any] = mapped_column(JSONB, nullable=False)
+    content: Mapped[Dict[str, Any]] = mapped_column(JSONB, nullable=False)
     creation_datetime: Mapped[dt.datetime] = mapped_column(
         TIMESTAMP(timezone=True), nullable=False
     )
@@ -44,7 +44,7 @@ class AggregateDb(Base):
 
     key: Mapped[str] = mapped_column(String, primary_key=True)
     owner: Mapped[str] = mapped_column(String, primary_key=True)
-    content: Mapped[Any] = mapped_column(JSONB, nullable=False)
+    content: Mapped[Dict[str, Any]] = mapped_column(JSONB, nullable=False)
     creation_datetime: Mapped[dt.datetime] = mapped_column(
         TIMESTAMP(timezone=True), nullable=False
     )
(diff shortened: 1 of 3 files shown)


foxpatch-aleph (dismissed):
Clean, well-scoped typing improvement. The JSONB column type narrowing from `Mapped[Any]` to `Mapped[Dict[str, Any]]` is correct since these columns always store dict-shaped data. The credit handler refactoring removes three identical try/except blocks by centralizing parsing in a `_parse_credit_content` helper, making signatures self-documenting. The PR correctly defers the `chain_data_service.py` typing as out of scope. No correctness or security concerns.

**`src/aleph/handlers/content/post.py`** (line 85): Consider constraining `kind` with `Literal["distribution", "expense", "transfer"]` for tighter typing, though it's internal-only code so this is a minor style suggestion.

odesenfans:
Applied in 77e5219d. `kind: Literal["distribution", "expense", "transfer"]`.

foxpatch-aleph (approved):
This PR correctly narrows Mapped[Any] to Mapped[Dict[str, Any]] on three JSONB content columns (MessageDb, AggregateElementDb, AggregateDb), eliminating an entire category of accidental scalar-access bugs. It also centralizes three nearly-identical try/except ValidationError blocks into a generic _parse_credit_content helper, removing duplicate error handling from credit-balance handler functions and making their signatures self-documenting via typed Pydantic parameters. The tests pass, the refactoring is behavior-preserving, and there are no regressions or security issues.

PR #1156, opened 2026-05-21
fix(forget): authorize FORGET via security-aggregate delegation

odesenfans:
## Summary

- Replace the strict `target.sender == message.sender` check in `ForgetMessageHandler.check_permissions` with a per-target call to `is_sender_authorized_for_owner`, anchored on `target.parsed_content.address`. An owner (or any delegate authorized via the `security` aggregate with `types: [FORGET]`) can now forget content that was created on their behalf by another delegate.
- Tighten the inverse case: a former delegate whose `security`-aggregate authorization has been revoked can no longer forget messages they previously signed under that delegation. The old code consulted only the signer, never the aggregate.
- Promote the existing private `_check_delegated_authorization` helper in `aleph.permissions` to a public `is_sender_authorized_for_owner` with an updated docstring spelling out which message to pass when scoping the type / channel / chain filters.
- No DB schema change, no migration, no API surface change. The all-or-nothing guarantee for multi-target FORGETs is structural: `check_permissions` runs to completion before `process` is invoked by the message pipeline.

Design spec and implementation plan live under `docs/superpowers/specs/2026-05-20-forget-permission-delegation-design.md` and `docs/superpowers/plans/2026-05-20-forget-permission-delegation.md`.

## Behavior changes

- **Closed:** `O` (or any delegate of `O` with `types: [FORGET]`) can forget any message whose `content.address = O`, regardless of who signed it.
- **Tightened:** A former delegate `D1` can no longer forget messages they signed with `content.address = O` after `O` has revoked their delegation. The right to manage that content belongs to its declared owner.
- **Unchanged:** A sender forgetting messages they signed with `content.address = themselves` continues to work (the general rule applied to the self-owner case).

## Test plan

- [x] `tests/permissions/test_forget_permissions.py` (new): 12 tests covering self-ownership regression, owner-forgets-delegate, second-delegate cleanup, the original combined scenario, forget-after-revocation, cross-owner multi-target, aggregate-key expansion, stranger denial, revoked-delegate denial (the intended tightening), wrong-scope denial, no-claim-over-self-signed denial, and all-or-nothing target validation.
- [x] `tests/message_processing/test_process_forgets.py`: 5 existing end-to-end tests still green (no regression).
- [x] `tests/permissions/test_check_sender_authorization.py`: 10 existing tests still green (the rename to `is_sender_authorized_for_owner` did not change behavior).
- [x] Broader pyaleph test suite (`tests/` minus `tests/integration`): 746 passed, 10 skipped, 0 failed.


foxpatch-aleph (dismissed):
This PR correctly replaces the strict sender-equality check in ForgetMessageHandler with a per-target authorization check using the existing security-aggregate delegation system. The rename of _check_delegated_authorization to is_sender_authorized_for_owner is clean, the docstring is clear about which message to pass, and the 12 new tests thoroughly cover the positive, negative, and edge cases. The code is correct, well-tested, and addresses a genuine gap in the permission system.

**`tests/permissions/test_forget_permissions.py`** (line 396): The docstring for test_forget_by_aggregate_key goes into helpful detail about the ItemHash typing constraint, but the distinction between the aggregate lookup key (line 414) and the underlying message's item_hash (line 415) is easy to miss. Consider a brief inline comment near the assignment of aggregate_key to clarify that it serves as the AggregateElementDb lookup key, not as a message hash.

foxpatch-aleph (dismissed):
This PR correctly replaces the strict sender-equality check in ForgetMessageHandler.check_permissions with a delegation-aware authorization check via is_sender_authorized_for_owner. The owner of content (or their authorized FORGET delegate) can now forget that content regardless of who signed it. The change also tightens security by preventing revoked delegates from forgetting content they previously signed. The code is clean, well-documented, and thoroughly tested with 12 new tests covering all edge cases including self-ownership, multi-target, cross-owner, aggregate-key, and revocation scenarios.

**`tests/permissions/test_forget_permissions.py`** (line 381): Minor: the comment "base check passes trivially (sender == content.address)" is slightly misleading as the per-target authorization doesn't use content.address — it uses target_message.parsed_content.address. The comment refers to the base-level ContentHandler.check_permissions check, which is correct but could be clarified.

foxpatch-aleph (commented):
Well-designed PR that replaces strict sender-equality with delegation-based per-target authorization for FORGET messages. The 'create-implies-delete' rule is sound and consistently applied. Tests are thorough. The only substantive issue is a docstring that contradicts the actual call site.

**`src/aleph/permissions.py`** (line 12): The docstring states 'message is the operation being requested (e.g. the FORGET being submitted), NOT the target it acts on', but in forget.py:147 this function is called with the target message (the message being forgotten), not the FORGET message itself. The call site behavior is correct — we want to check delegation scoped to the target's type/channel/chain — but the docstring should document both valid usage patterns rather than prescribing only one.

foxpatch-aleph (approved):
This PR correctly replaces the strict sender-equality check in FORGET permissioning with a per-target authorization check via `is_sender_authorized_for_owner`. The new model binds forget authorization to the target's owner (`content.address`) rather than the target's signer, enabling owners and their delegates to manage content irrespective of which delegate originally signed it. A desirable security tightening is that revoked delegates can no longer forget messages they signed under a prior delegation. The function rename from `_check_delegated_authorization` to `is_sender_authorized_for_owner` is clean, the docstring clearly documents both calling patterns, and 12 new unit tests cover all relevant positive, negative, and edge-case scenarios. Existing e2e tests remain green.

PR #1157, opened 2026-05-21
fix(message-handler): reject pending messages with missing signature early

odesenfans:
## Summary

When a P2P pending message arrives with a null `signature` (and `check_message=True`, which is the default), `MessagePublisher.add_pending_message` now rejects it before reaching the DB. Previously the row was caught by the `signature_not_null_if_check_message` CHECK constraint at INSERT time, which:

- raised `IntegrityError` and was misreported as "Duplicate pending message detected" in logs (the `except IntegrityError` block in `message_handler.py` only narrates the unique-conflict case),
- left no `RejectedMessageDb` record, so the sender got no signal,
- wasted the upstream `load_fetched_content` work (inline content deserialization, content-handler `is_related_content_fetched` check) before the failed insert.

The new path calls `reject_new_pending_message` with `InvalidSignature("Missing signature")` so the message is recorded as `REJECTED` with `ErrorCode.INVALID_SIGNATURE`. On-chain ingestion (`check_message=False`) is unaffected and continues to accept a null signature, since the on-chain tx itself is the proof of origin.

## Test plan

- [x] `tests/message_processing/test_process_pending_messages.py::test_add_pending_message_rejects_null_signature` (new): fails on `main`, passes with the fix; checks return value, absence of pending row, REJECTED status, and `INVALID_SIGNATURE` error code.
- [x] `test_duplicated_pending_message` still passes.
- [x] `black`, `isort`, `ruff` clean on changed files.

### DIFF
--- a/src/aleph/handlers/message_handler.py
+++ b/src/aleph/handlers/message_handler.py
@@ -191,6 +191,20 @@ class MessagePublisher(BaseMessageHandler):
                 session.commit()
                 return None
 
+            if check_message and not message.signature:
+                error = InvalidSignature("Missing signature")
+                LOGGER.warning(
+                    "Rejecting message %s: missing signature", message.item_hash
+                )
+                reject_new_pending_message(
+                    session=session,
+                    pending_message=message_dict,
+                    exception=error,
+                    tx_hash=tx_hash,
+                )
+                session.commit()
+                return None
+
             pending_message = PendingMessageDb.from_obj(
                 message,
                 reception_time=reception_time,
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
A clean, well-structured fix that rejects pending P2P messages with null signatures early, before wasting work on load_fetched_content and before hitting the DB constraint that previously produced a misleading IntegrityError log message. The code follows existing patterns in the same method, the check_message guard correctly preserves on-chain ingestion behavior, and the test comprehensively covers all relevant assertions.

PR #1162, opened 2026-05-24
feat(api): filter account files by hash

odesenfans:
## Summary

- Add an optional `file_hash` query param to `GET /api/v0/addresses/{address}/files` so callers can look up a single file owned by an address without paginating through every pin.
- The response shape is unchanged; the result is just narrowed to the matching pin. Each item already includes `item_hash` (the STORE message hash), which is useful for follow-up actions like forgetting the message.
- Returns `404` when no pin matches (consistent with the existing behavior when an address has no pins).

## Test plan

- [x] New `tests/api/test_account_files.py` covers: match returns the right `item_hash`, unknown hash returns 404, hash pinned by a different owner returns 404.
- [x] `pytest tests/api/test_account_files.py tests/db/test_files.py tests/api/test_cursor_pagination.py` passes (17/17) against a local postgres.
- [x] `black`, `isort`, `ruff`, `mypy` clean on touched files.

### DIFF
--- a/src/aleph/db/accessors/files.py
+++ b/src/aleph/db/accessors/files.py
@@ -237,6 +237,7 @@ def get_address_files_for_api(
     after_time: Optional[dt.datetime] = None,
     after_hash: Optional[str] = None,
     cursor_mode: bool = False,
+    file_hash: Optional[str] = None,
 ) -> Iterable[Row]:
     select_stmt = (
         select(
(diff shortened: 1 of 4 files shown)


foxpatch-aleph (approved):
A clean, minimal feature that adds an optional `file_hash` query parameter to filter account files. The implementation is correct across all three layers (schema, accessor, controller), includes thorough tests covering match, miss, and cross-owner scenarios, and follows existing code patterns closely. The only notable observation is that `total_size` in the response reflects all files for the address rather than only the filtered file, but this is consistent with pre-existing behavior and is not a bug.

**`src/aleph/web/controllers/accounts.py`** (line 691): When `file_hash` is specified, `total_size` from `get_address_files_stats` still reports the total size of all files for the address, not just the filtered file. For example, if the address has 10 GB of files and you filter by a single file hash, `total_size` will still show 10 GB while only one file is returned. Consider passing `file_hash` to `get_address_files_stats` (or documenting this behavior) so the response is internally consistent.

**`src/aleph/web/controllers/accounts.py`** (line 720): Same `total_size` mismatch as the cursor path above — `get_address_files_stats` is called without the `file_hash` filter.

PR #1164, opened 2026-05-25
Add STORE file-fetch metrics for IPFS herd measurement

aliel:
Instrument fetch_related_content to measure download time and failure
  rate during the IPFS thundering herd (90 CCNs pulling the same CID at
  once).

  - Log structured `ipfs_fetch` lines (hash, type, path, size, duration, outcome) on every pin/http fetch — source for p95/max via logs.
  - Add Redis counters surfaced at /metrics: pyaleph_store_fetch_total, pyaleph_store_fetch_failed_total, and pyaleph_store_fetch_duration_ms_sum (sum of successful durations; mean = sum / (total - failed)).
  - Add NodeCache.incrby for the duration sum.

### DIFF
--- a/src/aleph/handlers/content/store.py
+++ b/src/aleph/handlers/content/store.py
@@ -201,6 +201,10 @@ class StoreMessageHandler(ContentHandler):
             if ipfs_enabled and _should_pin_on_ipfs(
                 file_stats=file_stats, min_file_size_for_pinning=1024 * 1024
             ):
+                # Counted before the fetch so every attempt is represented. A crash
+                # between here and the success/failure path leaves the total without a
+                # matching duration or failure entry, marginally skewing the mean — an
+                # acceptable tradeoff for approximate monitoring counters.
                 await self.storage_service.node_cache.incr(STORE_FETCH_TOTAL_KEY)
                 try:
                     with Timer() as timer:
@@ -208,9 +212,10 @@ class StoreMessageHandler(ContentHandler):
                 except Exception:
                     await self.storage_service.node_cache.incr(STORE_FETCH_FAILED_KEY)
                     LOGGER.warning(
-                        "ipfs_fetch hash=%s type=ipfs path=pin size=%s outcome=fail",
+                        "ipfs_fetch hash=%s type=ipfs path=pin size=%s duration=%.2f outcome=fail",
                         file_hash,
                         file_stats.size,
+                        timer.elapsed(),
                     )
                     raise
                 await self.storage_service.node_cache.incrby(
@@ -246,9 +251,10 @@ class StoreMessageHandler(ContentHandler):
         except AlephStorageException:
             await self.storage_service.node_cache.incr(STORE_FETCH_FAILED_KEY)
             LOGGER.warning(
-                "ipfs_fetch hash=%s type=%s path=http outcome=unavailable",
+                "ipfs_fetch hash=%s type=%s path=http duration=%.2f outcome=unavailable",
                 file_hash,
                 item_type,
+                timer.elapsed(),
             )
             raise FileUnavailable("Could not retrieve file from storage at this time")
 
--- a/src/aleph/handlers/content/store.py
+++ b/src/aleph/handlers/content/store.py
@@ -214,7 +214,7 @@ class StoreMessageHandler(ContentHandler):
                     )
                     raise
                 await self.storage_service.node_cache.incrby(
-                    STORE_FETCH_DURATION_MS_SUM_KEY, int(timer.elapsed() * 1000)
+                    STORE_FETCH_DURATION_MS_SUM_KEY, round(timer.elapsed() * 1000)
                 )
                 LOGGER.info(
                     "ipfs_fetch hash=%s type=ipfs path=pin size=%s duration=%.2f outcome=ok",
@@ -253,7 +253,7 @@ class StoreMessageHandler(ContentHandler):
             raise FileUnavailable("Could not retrieve file from storage at this time")
 
         await self.storage_service.node_cache.incrby(
-            STORE_FETCH_DURATION_MS_SUM_KEY, int(timer.elapsed() * 1000)
+            STORE_FETCH_DURATION_MS_SUM_KEY, round(timer.elapsed() * 1000)
         )
         LOGGER.info(
             "ipfs_fetch hash=%s type=%s path=http size=%s duration=%.2f outcome=ok",
--- a/src/aleph/handlers/content/store.py
+++ b/src/aleph/handlers/content/store.py
@@ -42,6 +42,7 @@ from aleph.services.ipfs import IpfsService
 from aleph.storage import StorageService
 from aleph.toolkit.constants import MiB
 from aleph.toolkit.costs import are_store_and_program_free, is_credit_only_required
+from aleph.toolkit.timer import Timer
 from aleph.toolkit.timestamp import timestamp_to_datetime, utc_now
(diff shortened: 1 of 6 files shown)


foxpatch-aleph (changes requested):
Solid instrumentation PR that follows existing code patterns. The logic is correct — total/failed counters and duration sum are properly tracked for both IPFS pin and HTTP fetch paths. The main issue is that the existing test_metrics test needs updating to include the three new metric fields in the expected Prometheus output, otherwise it will fail. Minor suggestion: consider rounding rather than truncating the millisecond duration conversion.

**`tests/web/controllers/test_metrics.py`** (line 64): The expected output string in test_metrics must include the three new fields: pyaleph_store_fetch_total 0, pyaleph_store_fetch_failed_total 0, and pyaleph_store_fetch_duration_ms_sum 0. Since asdict() outputs all fields including defaults, the current assertion will fail.

**`src/aleph/handlers/content/store.py`** (line 208): Consider catching a more specific exception from pin_add (e.g., aioipfs.APIError) rather than the broad except Exception. This avoids masking unexpected errors like asyncio.CancelledError.

**`src/aleph/handlers/content/store.py`** (line 217): Consider using round() instead of int() for the millisecond conversion: int(round(timer.elapsed() * 1000)). Truncation via int() loses up to 1ms per fetch, which could slightly skew the sum metric for many fast operations.

foxpatch-aleph (commented):
The PR correctly instruments STORE file-fetch operations with Redis-backed metrics and Prometheus exposure. The instrumentation points are well-chosen, the code follows existing patterns, and the metrics keys are properly surfaced at /metrics. The main concern is the lack of test coverage for the new instrumentation — existing tests will pass without verifying that incr/incrby are called. No correctness bugs were found, though the total counter's placement before the operation means crash-interrupted fetches will slightly skew the mean calculation (acceptable for monitoring). Consider adding assertions to existing tests to verify the new Redis calls.

**`src/aleph/handlers/content/store.py`** (line 204): Incrementing STORE_FETCH_TOTAL_KEY before the operation means that if a crash occurs between incr() and the success/failure path, the total will be incremented without a corresponding duration or failure counter. This is acceptable for approximate monitoring metrics, but worth documenting.

**`src/aleph/handlers/content/store.py`** (line 216): Consider adding duration to the failure log line on line 211 (akin to the success log on line 219), so fast-vs-slow failures can be distinguished in log analysis.

**`src/aleph/web/controllers/metrics.py`** (line 41): These constants duplicate those in store.py. A shared module would prevent drift if keys are renamed, but the circular import concern is noted. Consider a dedicated metrics_keys.py module or importing from the source where possible.

**`tests/storage/test_store_message.py`** (line 98): Consider adding assertions to verify that node_cache.incr() and node_cache.incrby() are called with the expected keys, to prevent regressions in the metrics instrumentation.

foxpatch-aleph (dismissed):
This PR adds Prometheus-style metrics and structured logging to measure STORE file-fetch performance, specifically targeted at diagnosing the IPFS thundering herd issue. The implementation is clean, well-commented about its intentional tradeoffs (e.g., the incr-before-try skew), and follows existing patterns in the codebase (Redis key constants, metrics wiring). Tests cover the success and failure paths for both pin and HTTP fetch with the correct metric assertions. Minor nits: log format inconsistency between size field presence in pin-failure vs http-failure log lines, and no test for HTTP fetch failure specifically, but neither is a blocker.

**`src/aleph/handlers/content/store.py`** (line 215): The pin-failure log line includes `size=%s` but the http-failure log at line 254-258 omits size. Either both should include a size field with the best available value (e.g., `size=0` or `size=N/A` when unknown) for consistency, or the structured log format should be documented somewhere.

**`src/aleph/handlers/content/store.py`** (line 215): Minor: `size=%s` logs an int via `%s` — using `%d` would be more idiomatic for an integer field.

foxpatch-aleph (commented):
A clean, well-structured PR that adds IPFS fetch metrics by instrumenting the two fetch paths in `fetch_related_content` (IPFS pin and HTTP). The code is correct: `Timer.__exit__` always sets `end_time` even on exception, so `timer.elapsed()` is safe in `except` blocks. The `incr-before-fetch` / `incr(fail)`-or-`incrby(duration)` pattern is consistently applied. Tests cover the IPFS pin success and failure paths and validate Prometheus output. The main gap is the lack of integration-level test coverage for the `ItemType.storage` HTTP fetch path — only the isolated `_store_fetch_keys` unit test validates storage key selection. A minor maintenance concern is the duplication of Redis key strings between `store.py` and `metrics.py`, though this is intentional to avoid circular imports.

**`tests/storage/test_store_message.py`** (line 260): Consider adding a test for the `ItemType.storage` HTTP fetch path to verify that `STORE_FETCH_STORAGE_*` counters are used. Currently only the isolated `_store_fetch_keys` unit test validates storage key selection, not the integration flow.

**`src/aleph/web/controllers/metrics.py`** (line 43): Redis key strings are duplicated from `store.py`. The circular-import concern is valid but consider extracting shared constants into a third module (e.g. `aleph/toolkit/metrics.py`) to avoid drift between the two files.

foxpatch-aleph (dismissed):
A well-structured change adding Redis-backed STORE file-fetch metrics. The implementation correctly distinguishes IPFS pin operations from HTTP storage fetches, handles success and failure paths with appropriate counter increments, and surfaces the data at /metrics. The tests cover the key scenarios (successful pin, successful HTTP fetch, pin failure). The code follows existing patterns in the codebase (duplicated Redis key constants to avoid circular imports, _read_int_key helper) and is well-commented for non-obvious tradeoffs like the crash-skew risk on counter ordering.

**`src/aleph/handlers/content/store.py`** (line 273): The HTTP failure log at line 274 doesn't include a `size=` field, unlike the IPFS pin failure log at line 235. Consider adding it for consistency, e.g. `size=0` or omitting it if not available.

foxpatch-aleph (approved):
Clean PR that instruments fetch_related_content with Redis-backed metrics counters for the IPFS thundering herd scenario. The architecture is well-considered: shared metric key constants in a leaf toolkit module prevent circular imports, the Timer context manager is properly reused, and both IPFS pin and HTTP fetch paths are instrumented separately by item type. Tests cover success/failure paths for both IPFS and storage types, as well as key isolation. The documented tradeoff (total counter incremented before the operation, so crashes can orphan a total without matching failure/duration) is acceptable for approximate monitoring. No correctness or security issues found.

**`src/aleph/handlers/content/store.py`** (line 202): nit: `total_key` is incremented before entering the `try` block for the pin path. If `_get_file_stats_from_ipfs` on line 190 raises (UnknownHashError, FileUnavailable), the total was already counted but no failure/duration is recorded. The comment on lines 198-201 acknowledges this tradeoff — consider also noting that a stat failure orphans the counter too.

**`src/aleph/handlers/content/store.py`** (line 245): observation: Only `AlephStorageException` is caught here; other exceptions (RuntimeError, asyncio.TimeoutError) will leave the total counter incremented without a matching failure count. This is consistent with the documented tradeoff but differs from the ipfs pin path which catches bare `Exception`. Consider whether this should be consistent.

**`tests/storage/test_store_message.py`** (line 308): style: The explicit `session.commit()` here is inconsistent with other test patterns that rely on the session factory context manager (e.g., opposite-line 126-130). Consider removing for consistency (the factory auto-commits). Non-blocking.

**`tests/storage/test_store_message.py`** (line 260): nice test: Broad `RuntimeError` side_effect confirms the `except Exception` catch works for non-API errors. Good defensive coverage.

PR #1165, opened 2026-05-26
feat(messages): add contentFormat query parameter to messages list endpoint

odesenfans:
## Summary

Adds a `contentFormat` query parameter to `GET /api/v0/messages.json` with three mutually exclusive values:

- `full` (default): the complete message content (unchanged behaviour).
- `headers`: a reduced, per-type metadata subset of `content`, built entirely from the denormalized columns (`owner`, `content_type`, `content_ref`, `content_key`, `content_item_hash`). The content JSONB is deferred and never read, so `headers` is as cheap on the server as `none` while still returning useful metadata.
- `none`: omit `content` entirely (the behaviour of the now-deprecated `excludeContent=true`).

Motivation: list responses for heavy messages routinely reach ~10 MB. `excludeContent` drops `content` wholesale, which removes the lightweight metadata an explorer needs (message subtype, target ref, aggregate key, stored item hash), forcing clients to fetch full content anyway. `headers` mode keeps just that metadata and strips the heavy remainder.

### Reduced `content` per type (`headers`)

| Message type | `content` keys returned |
| --- | --- |
| all | `address` |
| POST | + `type`, `ref` |
| AGGREGATE | + `key` |
| STORE | + `item_hash`, `ref` |
| PROGRAM / INSTANCE / FORGET | `address` only |

Fields whose backing column is NULL are omitted. `content.time` is intentionally omitted; the top-level `time` already covers it.

### Compatibility and scope

- `excludeContent` is kept as a deprecated alias. A pydantic validator resolves precedence: an explicit `contentFormat` wins; otherwise `excludeContent=true` resolves to `none`, else `full`. Existing `excludeContent` behaviour and tests are unchanged.
- Both `none` and `headers` defer the content JSONB at all list query sites (cursor mode and legacy page mode).
- Websocket path supports `full`/`none` only; `headers` degrades to `none` there (history and live broadcast).
- Out of scope: the single-message endpoint `GET /messages/{hash}` (separate strict-pydantic serialization path) and a third websocket variant.
- No migration: the required columns already exist.

## Test Plan

- [ ] `hatch run testing:test tests/api/test_content_format.py tests/api/test_list_messages.py` (50 tests pass locally)
- [ ] Per-type reduced content verified for POST/AGGREGATE/STORE/FORGET against fixtures
- [ ] Precedence: `contentFormat` overrides `excludeContent`; `excludeContent=true` resolves to `none`
- [ ] `contentFormat=headers` exercised through the cursor-pagination branch
- [ ] Websocket history strips content for both `none` and `headers`
- [ ] Invalid `contentFormat` returns 422
- [ ] Spot-check the response size reduction on a heavy real-world query (`full` vs `headers`)


foxpatch-aleph (dismissed):
Clean implementation of contentFormat query parameter for message list endpoint. The feature is well-motivated, backward-compatible with the deprecated excludeContent flag, and leverages denormalized columns to build reduced content without reading the JSONB. Test coverage is thorough across unit and integration tiers. Minor issues: _WsClient uses a boolean instead of ContentFormat enum (making the headers→none degradation implicit in the live WS broadcast path), and a redundant `or ContentFormat.FULL` fallback on line 529 that is dead code due to the model validator. Neither is a blocker — the feature works correctly.

**`src/aleph/web/controllers/messages.py`** (line 82): Consider changing `_WsClient.exclude_content: bool` to `content_format: ContentFormat` for consistency. The live broadcast path in `_on_message`/`_send_to_client` degrades headers→none implicitly via bool conversion, while `_send_history_to_ws` does so explicitly on line 626. Using ContentFormat everywhere would make the degradation logic consistent and centralized.

**`src/aleph/web/controllers/messages.py`** (line 529): The `or ContentFormat.FULL` fallback is dead code — `resolve_content_format` validator guarantees `content_format` is never None. Same issue on line 620.

foxpatch-aleph (approved):
Clean, well-tested feature that adds a contentFormat query parameter (full/headers/none) to the messages list endpoint. Properly deprecates excludeContent, defers the JSONB content column when possible, and provides thorough test coverage. Minor dead-code observation in the form of redundant fallback checks for content_format that is guaranteed non-None by the model validator.

**`src/aleph/web/controllers/messages.py`** (line 529): Redundant fallback: `query_params.content_format` is guaranteed non-None after the `resolve_content_format` model validator runs (see messages_query_params.py:146). The `or ContentFormat.FULL` here and at line 620 will never be reached.

PR #1166, opened 2026-05-27
config: bump max file upload size to 1 GiB

odesenfans:

- [ ] Are my files well typed
- [ ] Are there enough tests

## Changes

## How to test

Explain how to test your PR.

## Print screen / video

## Notes


foxpatch-aleph (dismissed):
Simple and correct config bump from 100 MiB to 1 GiB. The GiB constant is already defined and used elsewhere. No downstream side effects — authenticated uploads increase, unauthenticated uploads remain capped at 25 MiB.

foxpatch-aleph (dismissed):
Bumps the default maximum upload file size from 100 MiB to 1 GiB across all relevant config paths (storage, IPFS, aiohttp server-level). The associated test is correctly refactored to use a mock config override rather than hardcoding values tied to the old default, making it faster and independent of the production constant. The change is minimal, correct, and well-tested.

**`tests/api/test_ipfs.py`** (line 284): The test now correctly decouples from the production default by overriding the config value. Consider also adding a similar decoupled cap test for the storage endpoint (`/api/v0/storage/add_file`) since it also consumes `DEFAULT_MAX_FILE_SIZE`, but this is a pre-existing gap outside the scope of this PR.

foxpatch-aleph (approved):
Clean, focused PR that raises the authenticated IPFS upload cap from 100 MiB to 1 GiB while keeping other size limits (storage.max_file_size at 100 MiB, unauthenticated at 25 MiB, CAR at 4 GiB) unchanged. The new constant name clearly distinguishes the IPFS upload endpoint from the storage endpoint. The test is correctly refactored to be independent of the production default by pinning a small cap via mock_config. No correctness, security, or testing issues found.

PR #1169, opened 2026-05-28
fix(store): widen IPFS size lookup timeout in pre_check_balance

odesenfans:
## Summary

- `pre_check_balance` called `get_ipfs_size(item_hash)` with the helper's default `timeout=1, tries=1`. On any busy local kubo (which is the normal state right after a CAR upload), the single `dag.get` round-trip can blow past 1s, raising `FileUnavailable`. Pending-message retries each get the same 1s budget, so they keep failing and the message is rejected with error code 4 (`FILE_UNAVAILABLE`) even when the content is fully pinned locally.
- Pass `config.ipfs.stat_timeout` and `tries=3` from the call site so this deadline matches the rest of the IPFS read path.
- Reshape `FileNotFoundException.__init__` to take `(file_hash, details)`. All six raise sites in `store`, `service/ipfs/service`, and `message_handler` were stuffing prose into the `file_hash` slot, so rejection records and logs looked like `File not found: Could not retrieve IPFS content at this time` with no actual CID. They now carry the CID and the diagnostic separately.

## Concrete symptom this fixes

A user uploads a website CAR via `/api/v0/ipfs/add_car`. The endpoint succeeds (CAR imported, root recursively pinned, all blocks present on the local daemon, `files.stat` returns valid `CumulativeSize`). The accompanying STORE message is later rejected with `error_code=4`, `details="File not found: Could not retrieve IPFS content at this time"`. The rejection comes from `pre_check_balance`'s 1-second `dag.get` budget, not from anything actually being unavailable.

## Test plan

- [x] `pytest tests/services/test_ipfs_service.py tests/storage/test_store_message.py -v` (17 passed)
- [ ] On the affected node, deploy and re-upload the same CAR; confirm the STORE message reaches `processed`
- [ ] Tail rejection records and confirm `details` now carries the CID instead of prose

### DIFF
--- a/src/aleph/handlers/content/store.py
+++ b/src/aleph/handlers/content/store.py
@@ -92,7 +92,9 @@ async def _get_file_stats_from_ipfs(
         except aioipfs.InvalidCIDError as e:
             raise UnknownHashError(f"Invalid IPFS hash from API: '{cid}'") from e
         if stats is None:
-            raise FileUnavailable("Could not retrieve IPFS file stats at this time")
+            raise FileUnavailable(
+                str(cid), "could not retrieve IPFS file stats at this time"
+            )
 
         if stats["Type"] == "file":
             is_folder = False
@@ -114,7 +116,8 @@ async def _get_file_stats_from_ipfs(
             getattr(error, "message", None),
         )
         raise FileUnavailable(
-            f"Timeout ({stat_timeout}s) while retrieving stats of hash {cid}: {getattr(error, 'message', None)}"
+            str(cid),
+            f"timeout {stat_timeout}s retrieving IPFS file stats: {getattr(error, 'message', None)}",
         )
 
     except aioipfs.APIError as error:
@@ -268,7 +271,9 @@ class StoreMessageHandler(ContentHandler):
                 item_type,
                 timer.elapsed(),
             )
-            raise FileUnavailable("Could not retrieve file from storage at this time")
+            raise FileUnavailable(
+                file_hash, "could not retrieve file from storage at this time"
+            )
 
         await self.storage_service.node_cache.incrby(
             duration_key, round(timer.elapsed() * 1000)
@@ -318,7 +323,9 @@ class StoreMessageHandler(ContentHandler):
                 ipfs_byte_size: int | None = stored_file.size
             else:
                 ipfs_byte_size = await self.storage_service.ipfs_service.get_ipfs_size(
-                    content.item_hash
+                    content.item_hash,
+                    timeout=config.ipfs.stat_timeout.value,
+                    tries=3,
                 )
             if ipfs_byte_size:
                 storage_mib = Decimal(ipfs_byte_size / MiB)
--- a/src/aleph/handlers/message_handler.py
+++ b/src/aleph/handlers/message_handler.py
@@ -108,7 +108,9 @@ class BaseMessageHandler:
         except (ContentCurrentlyUnavailable, Exception) as e:
             if not isinstance(e, ContentCurrentlyUnavailable):
                 LOGGER.exception("Can't get content of object %s" % item_hash)
-            raise MessageContentUnavailable(f"Could not fetch content for {item_hash}")
+            raise MessageContentUnavailable(
+                item_hash, "could not fetch message content"
+            )
 
         try:
             validated_message = MessageDb.from_pending_message(
--- a/src/aleph/services/ipfs/service.py
+++ b/src/aleph/services/ipfs/service.py
@@ -205,6 +205,7 @@ class IpfsService:
                     raise FileUnavailable(
                         hash, "could not retrieve IPFS content at this time"
                     )
+                continue
             except (
                 concurrent.futures.CancelledError,
                 aiohttp.client_exceptions.ClientConnectorError,
--- a/src/aleph/services/ipfs/service.py
+++ b/src/aleph/services/ipfs/service.py
@@ -199,9 +199,11 @@ class IpfsService:
                 await asyncio.sleep(0.5)
                 continue
             except asyncio.TimeoutError:
-                raise FileUnavailable(
-                    hash, "could not retrieve IPFS content at this time"
-                )
+                if try_count >= tries:
+                    raise FileUnavailable(
+                        hash, "could not retrieve IPFS content at this time"
+                    )
+                await asyncio.sleep(0.5)
             except (
                 concurrent.futures.CancelledError,
                 aiohttp.client_exceptions.ClientConnectorError,
--- a/src/aleph/services/ipfs/service.py
+++ b/src/aleph/services/ipfs/service.py
@@ -199,11 +199,12 @@ class IpfsService:
                 await asyncio.sleep(0.5)
                 continue
             except asyncio.TimeoutError:
+                # A timeout already consumed the full `timeout` budget, so it is
+                # its own backoff: retry immediately, or give up once exhausted.
                 if try_count >= tries:
                     raise FileUnavailable(
                         hash, "could not retrieve IPFS content at this time"
                     )
-                await asyncio.sleep(0.5)
             except (
                 concurrent.futures.CancelledError,
                 aiohttp.client_exceptions.ClientConnectorError,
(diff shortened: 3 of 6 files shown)


foxpatch-aleph (dismissed):
Correctly fixes a production issue where pre_check_balance used a 1s/1try timeout for IPFS size lookups, causing false FILE_UNAVAILABLE rejections on busy kubo nodes. Also improves diagnostic quality by ensuring FileNotFoundException.error messages carry the actual CID instead of prose. The changes are well-scoped, consistent with existing patterns, and all existing tests pass. No bugs or security issues found.

**`tests/message_processing/test_process_stores.py`** (line 462): Existing tests use `assert_called_once_with(hash)` which doesn't verify the new `timeout` and `tries` kwargs. Consider hardening to `assert_called_once_with(hash, timeout=mock_config.ipfs.stat_timeout.value, tries=3)` to explicitly validate the new parameters are passed.

**`tests/message_processing/test_process_stores.py`** (line 517): Same as line 462 — the assertion doesn't validate the new `timeout` and `tries` keyword arguments.

**`tests/message_processing/test_process_stores.py`** (line 581): Same as line 462 — the assertion doesn't validate the new `timeout` and `tries` keyword arguments.

**`tests/message_processing/test_process_stores.py`** (line 709): Same as line 462 — the assertion doesn't validate the new `timeout` and `tries` keyword arguments.

**`tests/message_processing/test_process_stores.py`** (line 807): Same as line 462 — the assertion doesn't validate the new `timeout` and `tries` keyword arguments.

foxpatch-aleph (changes requested):
The PR correctly addresses the core issue of widening the IPFS size lookup timeout in pre_check_balance, and the FileNotFoundException refactor improves observability by preserving the CID in error records. However, there is a logic bug: asyncio.TimeoutError in get_ipfs_size raises FileUnavailable immediately without retrying, even though the caller now passes tries=3. This means timeout errors won't actually benefit from retries, contradicting the intent of the change. Additionally, the FileNotFoundException constructor uses a parent __init__ that wraps string messages in a list, which is a pre-existing oddity worth noting. Tests should also verify the timeout/tries arguments are passed to get_ipfs_size.

**`src/aleph/services/ipfs/service.py`** (line 201): asyncio.TimeoutError raises FileUnavailable immediately without retrying, but the caller now passes tries=3. This means timeout errors won't benefit from the retry logic — the first timeout aborts the whole operation. The TimeoutError handler should either retry (like APIError at line 197) or the tries parameter passed from pre_check_balance should be revisited.

**`src/aleph/handlers/content/store.py`** (line 299): The new timeout/tries parameters are correct, but there are no tests verifying that pre_check_balance passes config.ipfs.stat_timeout and tries=3 to get_ipfs_size. The existing mock-based tests should assert the call arguments.

**`src/aleph/types/message_status.py`** (line 175): The parent MessageProcessingException.__init__ wraps a single string in a list, so self.args[0] becomes ["File not found: <hash>"]. The file_hash and details attributes are correctly set on the instance, but the underlying storage is wrapped in a list. This is pre-existing behavior unrelated to this PR, but worth being aware of if code reads exception.args[0] directly.

odesenfans:
Addressed in 8b6050e7.

**1. TimeoutError doesn't retry (`service.py:201`)** — fixed. Verified the control flow: `APIError` set `result=None`/continued (counted as a try), `TimeoutError` raised immediately, `Cancelled/ConnectorError` decremented `try_count`/continued. So `tries=3` was dead weight against timeouts. Changed the handler to retry until `try_count >= tries` then raise, matching the intent of the caller. Not mirroring the APIError path because that one silently returns `None` on exhaustion, which in `pre_check_balance` would proceed with `message_cost = Decimal(0)` instead of triggering a retry — wrong outcome for a timed-out cost lookup.

**2. Tests don't assert the new kwargs (`store.py:299`)** — fixed. Running `tests/message_processing/test_process_stores.py::*pre_check_balance*` against the rebased branch in fact surfaced 5 failures: the existing `assert_called_once_with(hash)` assertions were checking positional-only and silently masking the new call shape, which would have left a CI hole. Updated all 5 to `assert_called_once_with(hash, timeout=30, tries=3)`. Also added `test_get_ipfs_size_timeout_error_multiple_tries` exercising the retry-on-timeout path.

**3. `MessageProcessingException.__init__` wraps in a list (`message_status.py:175`)** — acknowledged, not in this PR. Verified at `message_status.py:71-79`: the constructor explicitly does `[errors]` for str input, and the parent's `details()` method at `:85-111` iterates `self.args[0]` as a sequence, so the wrapping is load-bearing for that consumer. My PR adds `self.file_hash` and `self.details` as direct attributes specifically so callers that want the clean values don't have to go through `.args`. Reshaping the parent contract is a wider refactor and out of scope here.

Test run on the new commit: `pytest tests/services/test_ipfs_service.py tests/storage/test_store_message.py tests/message_processing/test_process_stores.py::*pre_check_balance*` → 26 passed.

foxpatch-aleph (commented):

foxpatch-aleph (dismissed):
This PR fixes a production bug where `pre_check_balance` would reject STORE messages with error code 4 because it called `get_ipfs_size` with the default 1s timeout and single try, which is insufficient when kubo is busy after a CAR upload. The fix passes `config.ipfs.stat_timeout` (typically 30s) and `tries=3`, matching the rest of the IPFS read path. Additionally, the `FileNotFoundException` constructor is reshaped to accept `(file_hash, details)` so rejection records carry the actual CID instead of a prose message. All six raise sites are correctly updated. Tests cover the new retry behavior and updated call-site arguments.

**`src/aleph/services/ipfs/service.py`** (line 201): The `aioipfs.APIError` handler on line 198 also retries via the while loop, but doesn't have its own dedicated raise-on-exhaustion check. It relies on the `while (result is None) and (try_count < tries)` condition, which correctly limits retries. This is pre-existing behavior and unchanged by this PR.

foxpatch-aleph (dismissed):
This PR correctly widens the IPFS size lookup timeout in `pre_check_balance` from the aggressive default of `timeout=1, tries=1` to `timeout=30, tries=3`, matching the rest of the IPFS read path. The `FileNotFoundException` refactor properly separates the CID from the diagnostic details across all six raise sites, which will improve rejection records and logging. All changes are logically sound and the new test (`test_get_ipfs_size_timeout_error_multiple_tries`) correctly validates the retry behavior.

**`src/aleph/services/ipfs/service.py`** (line 201): Nit: after a non-fatal timeout (`try_count < tries`), there's no explicit `continue` to short-circuit to the next loop iteration, unlike the `APIError` handler above. This works correctly (falls through to the `while` condition), but adding a `continue` would make the pattern consistent across all exception handlers.

**`tests/services/test_ipfs_service.py`** (line 214): Pre-existing issue (not introduced by this PR): `assert result is None` and `assert_called_once_with` on lines 215-216 are unreachable dead code inside the `pytest.raises(FileUnavailable)` block. Since the exception is always raised on timeout with `tries=1`, these assertions never execute. Consider cleaning them up.

foxpatch-aleph (dismissed):
Correctly fixes a real bug where get_ipfs_size used a 1s/1-try default from pre_check_balance, causing false FileUnavailable rejections on busy kubo. The fix passes config.ipfs.stat_timeout (30s) and tries=3. The timeout retry logic in get_ipfs_size is sound — timeouts retry immediately (the budget is already consumed) and only raise after exhausting tries. The FileNotFoundException refactoring correctly captures the CID in self.file_hash for logging. Tests are properly updated, including a new test for timeout retry exhaustion.

**`src/aleph/services/ipfs/service.py`** (line 135): If `tries=0` is passed, the while loop `(try_count < tries)` never executes, and the function returns `None`. Consider documenting this edge case or enforcing `tries >= 1`.

foxpatch-aleph (approved):
Fixes a real production issue where pre_check_balance used a 1s/1-try timeout for IPFS size lookups, causing false rejections on busy local Kubo nodes. The fix passes the configured stat_timeout (30s) with 3 retries, matching the rest of the IPFS read path. The secondary change to FileNotFoundException improves error diagnostics by including the actual CID instead of prose messages. Code is correct, tests are updated, and the pattern is consistent with existing abstractions.

**`src/aleph/services/ipfs/service.py`** (line 199): The timeout retry path does not sleep between attempts (unlike the APIError path which sleeps 0.5s). This is intentional per the comment — the timeout already consumed the full budget — but worth noting that 3 consecutive timeouts at 30s each means a worst-case ~90s blocking call.

PR #1170, opened 2026-05-28
feat(store): skip IPFS lookup in pre_check_balance when file is local

odesenfans:
## Summary

- `pre_check_balance` calls `get_ipfs_size(item_hash)` to compute the storage cost, even for content this node just uploaded itself and already recorded in its own `files` table.
- That round-trip is what is rejecting freshly-uploaded CARs: under any kubo busyness the `dag.get` can fail and the message ends up rejected with `FILE_UNAVAILABLE` even though the file is fully pinned locally.
- Query `get_file(session, item_hash)` first. On hit, use `stored_file.size` and skip the IPFS call entirely. On miss, fall back to `get_ipfs_size` as before.

Covers both `add_car` directories and `add_file` singletons (both write to `files` before broadcasting the STORE), and also helps when a STORE arrives from the network for content the node happens to already host.

## Relationship to #1169

#1169 widens the `get_ipfs_size` deadline (the IPFS-call path). This PR removes the IPFS call entirely on the common hot path (locally-known content). They are independent fixes for the same production symptom. Either alone fixes the freshly-uploaded-CAR case; together they also harden the remote-content path.

## Test plan

- [x] `pytest tests/storage/test_store_message.py -v` (4 passed)
- [ ] On the affected node, deploy and re-upload the same CAR; confirm the STORE message reaches `processed` without any `dag.get` traffic in the kubo logs for that CID

### DIFF
--- a/src/aleph/handlers/content/store.py
+++ b/src/aleph/handlers/content/store.py
@@ -18,6 +18,7 @@ from aleph_message.models import ItemHash, ItemType, PaymentType, StoreContent
 from aleph.config import get_config
 from aleph.db.accessors.files import (
     delete_file_pin,
+    get_file,
     get_file_tag,
     get_message_file_pin,
     insert_grace_period_file_pin,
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (approved):
A straightforward and well-motivated fix that avoids an unnecessary IPFS round-trip in the pre_check_balance path. When the file is already recorded in the local files table (e.g. from a prior add_car or add_file upload), the stored size is used directly, eliminating the risk of dag.get failures on busy kubo daemons. The code is correct, minimal, and clearly commented. No test coverage was added for the new code path, but the change is simple enough to reason about without it.

**`src/aleph/handlers/content/store.py`** (line 253): Nit: the new code uses `int | None` while `get_ipfs_size` returns `Optional[int]`. Both are valid in Python 3.12+; consider using a consistent style across the module.

PR #1171, opened 2026-05-28
Add randomized delay before IPFS fetch

aliel:
Sleep a random duration drawn uniformly from
  [0, ipfs.fetch_jitter_seconds] at the start of fetch_related_content
  when item_type is ipfs, spreading concurrent fetches over a window.

  - Add aleph.ipfs.fetch_jitter_seconds config field, default 5s.
  - New _apply_fetch_jitter helper; emits an ipfs_fetch_jitter log line with hash and delay. Storage-type fetches are unaffected.
  - Pin fetch_jitter_seconds to 0 in the test config so existing tests do not sleep.
  - Unit tests for the zero-window skip and within-window draw.

### DIFF
--- a/src/aleph/config.py
+++ b/src/aleph/config.py
@@ -268,6 +268,15 @@ def get_defaults():
             },
             # Timeout for file stat requests (seconds)
             "stat_timeout": 30,
+            # Randomized delay (in seconds, drawn uniformly from [0, value]) before
+            # an IPFS file fetch starts, to spread the thundering herd of CCNs
+            # pulling a newly-announced CID from the origin. A small non-zero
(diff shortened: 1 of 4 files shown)


foxpatch-aleph (approved):
A clean, well-motivated change that adds randomized jitter before IPFS fetches to spread thundering-herd load across CCNs. The implementation is minimal and correct: a small helper function draws from `random.uniform(0, config.fetch_jitter_seconds)` and sleeps, called at the right point in `fetch_related_content` for IPFS items only. Tests cover both the zero-window no-op and the within-range delay cases, and the test config pins the value to 0 so existing tests are unaffected. The config comment and docstring are well-written and provide clear rationale.

PR #1173, opened 2026-05-29
fix(ws): serialize MQ consumer start to stop duplicate message fanout

odesenfans:
## Problem

The explorer frontend reports that processed messages appear **multiple times** over the WebSocket stream.

`MessageBroadcaster` (introduced in #1062) runs a single shared RabbitMQ consumer per gunicorn worker that fans processed messages out to all connected clients. The consumer-start logic in `add()` is not concurrency-safe:

```python
self._clients.add(client)
await self._node_cache.incr(...)     # await -> yields the event loop
if self._consumer_tag is None:        # both racers still see None
    await self._start_consumer()      # both start a consumer
```

`_start_consumer()` only assigns `self._consumer_tag` on its **last line**, after `channel()`, `declare_queue()` and `consume()` all await. When two clients connect close together (routine when the explorer hits one worker), both pass the `is None` check and each create their own `exclusive` queue bound to `processed.*`. The topic exchange then delivers a copy of every message to **each** queue, so `_on_message` fires once per consumer and every client receives each message more than once.

It compounds: more concurrent first-connections means higher multiplicity, and the losing racer's `_channel`/`_queue`/`_consumer_tag` get overwritten so cleanup only tears down the last consumer. The orphaned channel keeps a live consumer (so `x-expires` never triggers) and keeps duplicating until its connection drops.

The publish side was ruled out: `processed.*` matches only `processed.<hash>` (new messages); confirmations publish as `confirmed.<hash>`, so each message is published to the WS exchange exactly once. The duplication is purely consumer-side.

## Fix

Guard the consumer lifecycle with an `asyncio.Lock` so the check-and-start in `add()`, the teardown in `remove()`/`shutdown()`, and `_restart_consumer()` cannot interleave. The shared consumer is now started exactly once regardless of how many clients connect concurrently.

## Tests

- `test_concurrent_add_starts_single_consumer` reproduces the race (fails before the fix with `2 == 1`) and passes after.
- `test_add_then_remove_starts_and_stops_consumer_once` guards the normal lifecycle against deadlock.

All 20 `tests/web/controllers/` tests pass; black/isort/ruff clean.

### DIFF
--- a/src/aleph/web/controllers/messages.py
+++ b/src/aleph/web/controllers/messages.py
@@ -107,6 +107,11 @@ class MessageBroadcaster:
         self._queue: Optional[aio_pika.abc.AbstractQueue] = None
         self._consumer_tag: Optional[aio_pika.abc.ConsumerTag] = None
         self._health_task: Optional[asyncio.Task] = None
+        # Serializes the consumer lifecycle (start/stop/restart). Without it,
+        # two clients connecting concurrently could both observe
+        # _consumer_tag is None and each start a consumer, binding a second
+        # queue to processed.* so every message gets fanned out more than once.
+        self._consumer_lock = asyncio.Lock()
 
         # Connection limit (same on every worker — from config).
         self.max_connections: int = config.websocket.max_message_connections.value
@@ -166,11 +171,12 @@ class MessageBroadcaster:
         """Restart the consumer after a channel failure."""
         await self._node_cache.incr(WS_BROADCASTER_CONSUMER_RESTARTS_KEY)
         LOGGER.warning("MessageBroadcaster: restarting consumer")
-        await self._stop_consumer()
-        try:
-            await self._start_consumer()
-        except Exception:
-            LOGGER.exception("MessageBroadcaster: failed to restart consumer")
+        async with self._consumer_lock:
+            await self._stop_consumer()
+            try:
+                await self._start_consumer()
+            except Exception:
+                LOGGER.exception("MessageBroadcaster: failed to restart consumer")
 
     async def _health_check_loop(self):
         """Periodic health check that restarts the consumer if the channel dies."""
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (dismissed):
Correctly fixes a concurrency bug in MessageBroadcaster where concurrent add() calls could each start a separate MQ consumer, causing duplicate message delivery. The fix uses an asyncio.Lock to serialize consumer lifecycle operations across add(), remove(), shutdown(), and _restart_consumer(). Tests reproduce the race condition and verify the fix. No deadlock risks, no logic errors, clean minimal diff.

**`tests/web/controllers/test_message_broadcaster.py`** (line 63): Consider also adding a test for the interleaving of remove() + concurrent add() — ensure that when the last client disconnects (triggering stop_consumer) while a new client simultaneously connects, the new add() correctly restarts the consumer. The current tests cover the main race (two concurrent adds) and the happy path, which is sufficient for approval.

foxpatch-aleph (dismissed):
Correct and well-tested fix for a genuine concurrency bug. The asyncio.Lock guards the consumer lifecycle check-and-start/stop/restart so that the shared MQ consumer is created exactly once regardless of concurrent client connections. The test reproduces the race condition and passes only with the fix. The lock boundaries are correctly placed — notably, remove() cancels the health task before acquiring the lock, avoiding a potential deadlock with _restart_consumer().

**`src/aleph/web/controllers/messages.py`** (line 230): Minor: reading self._clients outside the lock after _stop_consumer() released it. In practice safe since shutdown runs during server teardown with no new connections, but a self._clients copy under the lock would be more consistent.

**`tests/web/controllers/test_message_broadcaster.py`** (line 57): Consider also testing the _restart_consumer path through the health check to ensure the lock doesn't deadlock when add()/remove() run concurrently with a channel failure.

foxpatch-aleph (approved):
This PR correctly addresses a TOCTOU race condition in MessageBroadcaster where concurrent client connections could start multiple MQ consumers, each binding a separate queue to processed.* and causing duplicate message delivery. The fix uses an asyncio.Lock to serialize consumer lifecycle operations (start/stop/restart). The lock placement is minimal and correct across all four code paths that interact with the consumer. Tests reproduce the race condition and verify it's fixed. Minor missing coverage (concurrent remove/shutdown races) is non-blocking since the lock protects those uniformly.

**`tests/web/controllers/test_message_broadcaster.py`** (line 16): The _noop yields via asyncio.sleep(0) to simulate the real node_cache.incr's await point. This is what makes the test correctly reproduce the race — a good pattern worth preserving.

PR #1174, opened 2026-05-29
fix(api): validate page path param in messages list endpoint

RezaRahemtola:
## Summary

The `GET /api/v0/messages/page/{page}.json` endpoint cast the path parameter with a bare `int(url_page_param)`, **bypassing** the pydantic validation that the `?page=` query-param path goes through. Two attacker-reachable inputs both surfaced as **500s** (there is no error-handling middleware):

- **Non-numeric** (`/messages/page/abc.json`) → `int('abc')` raises `ValueError` → uncaught → 500.
- **Zero / negative** (`/messages/page/-5.json`) → parses fine, bypasses the `>= 1` floor → negative SQL `OFFSET` in the count query (`db/accessors/messages.py:414`) → 500.

Low severity (unauthenticated read endpoint, parameterized queries — no data leak or injection), but it returns server errors instead of client errors and adds log/Sentry noise.

## Fix

Reuse the existing `get_path_page()` helper (`web/controllers/utils.py`) — already used by the posts controllers — which raises `400` on non-numeric and `422` on `page < 1`.

## Tests

Added regression cases to `test_pagination` exercising the `/page/{page}.json` endpoint with `abc` (→400) and `0`/`-3` (→422). The existing `page=0/-3` cases only covered the query-param path, which was already validated.

## Verification

`ruff`, `black`, `isort`, and `py_compile` pass on the changed files. The full test suite requires the project's hatch + Postgres/redis/anvil test stack, which wasn't available in my environment — please confirm CI green.

### DIFF
--- a/src/aleph/web/controllers/messages.py
+++ b/src/aleph/web/controllers/messages.py
@@ -58,6 +58,7 @@ from aleph.web.controllers.app_state_getters import (
 )
 from aleph.web.controllers.utils import (
     get_item_hash_from_request,
+    get_path_page,
     mq_make_aleph_message_topic_queue,
     validate_cursor_pagination,
 )
@@ -521,8 +522,8 @@ async def view_messages_list(request: web.Request) -> web.Response:
 
     # If called from the messages/page/{page}.json endpoint, override the page
     # parameters with the URL one
-    if url_page_param := request.match_info.get("page"):
-        query_params.page = int(url_page_param)
+    if (url_page := get_path_page(request)) is not None:
+        query_params.page = url_page
 
     find_filters = query_params.model_dump(exclude_none=True)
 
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
Clean, minimal fix for two server-error paths on the /messages/page/{page}.json endpoint. Reuses an existing helper (get_path_page) that's already proven in the posts controllers. Tests cover the regression cases for non-numeric and out-of-range page values. The only nit is a pre-existing misleading error message in get_path_page itself ("greater than 1" should say "at least 1"), but that's outside this PR's scope.

**`src/aleph/web/controllers/utils.py`** (line 74): Pre-existing: error message says "greater than 1" but the check is `page < 1`. Page 1 passes validation but the message implies it wouldn't. Consider changing to "Page number must be at least 1." — out of scope for this PR.

foxpatch-aleph (approved):
Correctly reuses the existing get_path_page() helper to validate path parameters on the messages/page/{page}.json endpoint, converting uncaught 500 errors into appropriate 400 (non-numeric) and 422 (page<1) client errors. The change is minimal, follows established patterns from the posts controllers, and includes matching regression tests.

PR #1179, opened 2026-06-04
ipfs: Route all IPFS storage operations through pinning_client

aliel:
When config.ipfs.pinning.* is configured, the separate pinning service
  now serves every storage-role call: pin/unpin, add_bytes/add_json,
  cat, dag.get, block.stat, files.stat, and the streamed cat/get
  iterators. The main daemon (ipfs_client) is kept for P2P operations
  only: swarm.connect, id, pubsub sub/pub.

  Single-daemon deployments are unaffected — when no separate pinning
  service is configured, IpfsService constructor falls back to
  pinning_client = ipfs_client, so storage calls land on the same client
  as before.

  - service.py: get_ipfs_size, get_ipfs_content, get_ipfs_content_iterator, get_ipfs_directory_iterator now use pinning_client.
  - garbage_collector.py: pin.rm uses pinning_client.
  - store.py: _get_file_stats_from_ipfs uses pinning_client.
  - test_process_stores.py: stat mock updated to pinning_client.files.stat (test_storage.py already mocked both clients).

### DIFF
--- a/src/aleph/handlers/content/store.py
+++ b/src/aleph/handlers/content/store.py
@@ -80,7 +80,9 @@ class IpfsFileStats:
 async def _get_file_stats_from_ipfs(
     cid: ItemHash, ipfs_service: IpfsService, stat_timeout: int
 ) -> IpfsFileStats:
-    ipfs_client = ipfs_service.ipfs_client
+    # Stat is a storage operation: when a separate pinning service is
+    # configured, route it there. Falls back to the main daemon otherwise.
+    ipfs_client = ipfs_service.pinning_client
 
     try:
         try:
--- a/src/aleph/handlers/content/store.py
+++ b/src/aleph/handlers/content/store.py
@@ -80,9 +80,10 @@ class IpfsFileStats:
 async def _get_file_stats_from_ipfs(
     cid: ItemHash, ipfs_service: IpfsService, stat_timeout: int
 ) -> IpfsFileStats:
-    # Stat is a storage operation: when a separate pinning service is
-    # configured, route it there. Falls back to the main daemon otherwise.
-    ipfs_client = ipfs_service.pinning_client
+    # Stat is a storage operation: route through the storage client (which
+    # the service maps to the pinning service when configured, otherwise the
+    # main daemon).
+    ipfs_client = ipfs_service.pinning_client()
 
     try:
         try:
--- a/src/aleph/services/ipfs/service.py
+++ b/src/aleph/services/ipfs/service.py
@@ -83,19 +83,32 @@ class IpfsService:
     ):
         # P2P/network role only: swarm, identify, pubsub, peering.
         self.ipfs_client = ipfs_client
-        # Storage role: pinning, content reads, file stats. Falls back to the
-        # main daemon when no separate pinning service is configured, so
-        # single-daemon deployments behave the same as before.
-        self.pinning_client = pinning_client or ipfs_client
+        # Storage role: pinning, content reads, file stats. When the operator
+        # configures `ipfs.pinning.*` to a separate daemon, that daemon backs
+        # this client. Otherwise it falls back to the main daemon, so
+        # single-daemon deployments behave the same as before. Access via the
+        # ``pinning_client()`` method, not the attribute, so the routing
+        # decision lives in one place if it ever needs to grow.
+        self._pinning_client = pinning_client or ipfs_client
+
+    def pinning_client(self) -> aioipfs.AsyncIPFS:
+        """Return the aioipfs client that serves IPFS storage operations.
+
+        Centralizes the choice between the main daemon and the configured
+        pinning service so callers do not bind to the underlying attribute.
+        """
+        return self._pinning_client
 
     @classmethod
     def new(cls, config: Config) -> Self:
         # Create P2P client (for content retrieval, pubsub)
         p2p_client = make_ipfs_p2p_client(config)
 
-        # Create separate pinning client if configured differently
+        # Build a separate pinning client when an external pinning service is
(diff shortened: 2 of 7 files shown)


foxpatch-aleph (commented):

foxpatch-aleph (dismissed):
This PR correctly routes all storage-role IPFS operations through the pinning_client attribute of IpfsService. The fallback to the main daemon when no separate pinning service is configured is handled in the constructor, making this a safe, backward-compatible change. The code is clean, well-commented, and the existing tests pass because they all construct IpfsService without a separate pinning_client, which triggers the fallback path. The main gap is the absence of tests for the separate-pinning-daemon code path, but this is acceptable given the change's simplicity and the existing test coverage of the fallback path.

**`src/aleph/services/storage/garbage_collector.py`** (line 41): Nit: variable name `ipfs_client` points to `pinning_client`. Consider renaming to `pinning_client` for clarity, consistent with the other changes in the PR.

src/aleph/handlers/content/store.py
```diff
@@ -80,7 +80,9 @@ class IpfsFileStats:
 async def _get_file_stats_from_ipfs(
     cid: ItemHash, ipfs_service: IpfsService, stat_timeout: int
 ) -> IpfsFileStats:
-    ipfs_client = ipfs_service.ipfs_client
+    # Stat is a storage operation: when a separate pinning service is
+    # configured, route it there. Falls back to the main daemon otherwise.
+    ipfs_client = ipfs_service.pinning_client
```
odesenfans: Could you encapsulate this in an `ipfs_client()` method?

odesenfans (changes requested):
Encapsulate the pinning_client / ipfs_client in a dedicated method.

foxpatch-aleph (dismissed):
Cleanly separates storage operations (routed through pinning_client/storage_client) from P2P operations (ipfs_client). All code paths and test mocks are consistently updated. The property-based access for storage_client provides a clean abstraction point.

**`src/aleph/services/ipfs/service.py`** (line 92): The `storage_client` property provides a clean abstraction point for routing storage ops. Consider whether `storage_client` or `pinning_client` is the canonical name going forward — having both aliases could cause confusion long-term.

foxpatch-aleph (dismissed):
This PR correctly routes all IPFS storage operations through the storage_client (which maps to the configured pinning daemon or falls back to the main daemon), while keeping P2P operations on ipfs_client. The rename from pinning_client to storage_client accurately reflects the broader role of this client beyond just pinning. All code paths are consistently updated, tests are properly modified, and single-daemon deployments are unaffected. A cosmetic nit: _should_use_separate_pinning_client could be renamed for consistency, but this is non-blocking.

**`src/aleph/services/ipfs/service.py`** (line 414): Minor: the function name `_should_use_separate_pinning_client` and its docstring still reference "pinning" rather than "storage". Consider renaming to `_should_use_separate_storage_client` for consistency with the new attribute name.

**`tests/api/test_storage.py`** (line 87): Nit: `ipfs_client.files.stat` is mocked on line 87 but production code now routes `files.stat` through `storage_client`. This mock is dead code — harmless but could be removed.

foxpatch-aleph (dismissed):
This PR cleanly routes all IPFS storage-role operations through `pinning_client`, while preserving P2P operations (swarm, id, pubsub) on `ipfs_client`. The fallback logic (`pinning_client = pinning_client or ipfs_client`) ensures single-daemon deployments are unaffected. All changed call sites were verified against the separation of concerns, tests were correctly updated, and no bugs or security issues were found.

**`src/aleph/services/ipfs/service.py`** (line 416): Nit: function name `_should_use_separate_pinning_client` now undersells the role — it controls whether storage operations (not just pinning) go to a separate daemon. Consider renaming to `_should_use_separate_storage_client` for clarity, though this is non-blocking.

foxpatch-aleph (dismissed):
Clean refactor that routes all IPFS storage operations (pin/unpin, add, cat, dag.get, block.stat, files.stat, streamed iterators) through the pinning_client method. Single-daemon deployments are unaffected because pinning_client falls back to ipfs_client when no separate pinning service is configured. The code introduces a method accessor instead of a public attribute, which centralizes the routing decision. Tests are correctly updated to mock pinning_client() as a method call. Well-commented, correct, and no security concerns.

**`src/aleph/services/ipfs/service.py`** (line 82): The new method-based accessor vs attribute approach is clean. Consider if _should_use_separate_pinning_client should eventually be folded into pinning_client() to truly centralize the routing decision, but that's beyond the scope of this PR.

**`src/aleph/services/ipfs/service.py`** (line 271): get_ipfs_content_iterator and get_ipfs_directory_iterator now route through _pinning_client. Consider adding explicit test coverage for the routing of these streaming paths (currently mocked at the service level, not the client level).

foxpatch-aleph (approved):
Clean refactor routing all IPFS storage operations through `pinning_client()` method instead of `ipfs_client`, preserving the fallback for single-daemon deployments. All callers and tests are properly updated. No bugs or security issues found.

**`src/aleph/services/ipfs/service.py`** (line 94): Good pattern: centralizing the routing decision behind a method rather than exposing the raw attribute. This makes it easy to evolve the routing logic later.

**`src/aleph/services/ipfs/service.py`** (line 110): Log message text updated from "pinning operations" to "storage operations" — consistent with the semantic change. Nice touch.

**`src/aleph/web/controllers/ipfs.py`** (line 193): Correctly changed from attribute access `pinning_client.files.stat` to method call `pinning_client().files.stat`. All tests updated accordingly.

**`src/aleph/handlers/content/store.py`** (line 86): Correctly uses `pinning_client()` method instead of `ipfs_client`. The inline comment clearly explains the rationale.

PR #1183, opened 2026-06-10
feature: IPNS support (registration, tracking, republishing, gateway)

odesenfans:
## Summary

Adds first-class, paid IPNS name support to Aleph Cloud. A STORE message with the new `item_type: ipns` registers an IPNS name (`item_hash` is the base36 `k51...` name):

- **Publish**: the message carries a user-signed IPNS record (`ipns_record`, base64, 10 KiB cap). The node verifies it against the name via kubo, pins the record's value CID within the paid quota (`max_size_mib`), stores the registration in the new `ipns_records` table and injects the record into the DHT. No node ever holds a private key.
- **Update**: same message shape with a higher record sequence number. Updates self-order by sequence, so out-of-order processing needs no chain logic (`ref` is forbidden for IPNS content). The registration keeps exactly one re-pointable pin; replaced CIDs get the existing grace-period treatment and superseded messages stop billing.
- **Track-only**: omit `ipns_record` and the node fetches the current record from the DHT, then proceeds identically (keep-alive for names managed elsewhere).
- **Republish worker**: a periodic job (default 4h, `ipfs.ipns.*` config) re-injects records (kubo DHT records expire within ~24-48h) and re-resolves names to adopt newer records, with per-record failure isolation. Over-quota content keeps the last good CID pinned (`status=over_quota`); expired records stop republishing but stay pinned (`status=expired`).
- **API**: `GET /api/v0/ipns/{name}` (registration state, used by clients to bootstrap sequence numbers), `GET /api/v0/ipns/{name}/raw` (302 to `/api/v0/storage/raw/{cid}`), `GET /api/v0/addresses/{address}/ipns`. Registered names only: CCNs are not open IPNS resolvers.
- **Uploads**: `add_file`/`add_car` accept IPNS STORE metadata; the uploaded (root) CID must match the signed record's value instead of `item_hash`.
- **Pricing**: flat name fee (new `CostType.IPNS`, covers republish work) plus storage billed on the declared quota, so billing never changes without a message. Defaults live in `DEFAULT_PRICE_AGGREGATE` under the new `ipns` product and are tunable via the network pricing aggregate.

## Dependencies and rollout

- Depends on aleph-message branch [`ipns-support`](https://github.com/aleph-im/aleph-message/tree/ipns-support) (new `ItemType.ipns` shape rule + `StoreContent` fields), consumed as a git dependency in `pyproject.toml`. To be swapped for a released version once stabilized.
- Rollout note: nodes running released aleph-message reject IPNS STOREs as malformed (k51 names fail `ItemHash` validation), a clean failure. The network needs to upgrade before IPNS traffic flows, standard for additive message-type changes.
- Migration `0060_c7d2e9f4a1b8` creates `ipns_records`; `file_pins` gains only a new polymorphic type value.

## Test plan

- [x] aleph-message: 17 new unit tests (shape rule, StoreContent validators); suite at baseline (90 passed)
- [x] pyaleph: 44 new tests across models, accessors, IpfsService kubo wrappers, pricing, handler (publish/update/track/stale/forget/quota/cost-migration), republish worker, API endpoints, uploads
- [x] Full suite: 864 passed, 10 skipped; only pre-existing environmental errors (ethereum tests need a local anvil)
- [ ] End-to-end against a live kubo daemon (record signing via SDK, DHT round-trip)

### DIFF
--- a/deployment/migrations/versions/0060_c7d2e9f4a1b8_add_ipns_records.py
+++ b/deployment/migrations/versions/0060_c7d2e9f4a1b8_add_ipns_records.py
@@ -1,0 +1,41 @@
+"""add ipns_records
+
+Revision ID: c7d2e9f4a1b8
+Revises: b9c4f1e6a2d7
+Create Date: 2026-06-10
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = "c7d2e9f4a1b8"
+down_revision = "b9c4f1e6a2d7"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "ipns_records",
+        sa.Column("name", sa.String(), nullable=False),
+        sa.Column("owner", sa.String(), nullable=False),
+        sa.Column("item_hash", sa.String(), nullable=False),
+        sa.Column("record", sa.LargeBinary(), nullable=True),
+        sa.Column("record_sequence", sa.BigInteger(), nullable=True),
+        sa.Column("record_validity", sa.TIMESTAMP(timezone=True), nullable=True),
+        sa.Column("max_size_mib", sa.Integer(), nullable=False),
+        sa.Column("resolved_cid", sa.String(), nullable=True),
+        sa.Column("last_resolved", sa.TIMESTAMP(timezone=True), nullable=True),
+        sa.Column("last_republished", sa.TIMESTAMP(timezone=True), nullable=True),
+        sa.Column("status", sa.String(), nullable=False),
+        sa.Column("created", sa.TIMESTAMP(timezone=True), nullable=False),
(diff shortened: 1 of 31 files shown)


foxpatch-aleph (approved):
Well-structured PR adding comprehensive IPNS support. The code is clean, the architecture separating fetch/process/forget is consistent with existing patterns, the sequence-based self-ordering handles race conditions correctly, cost migration on updates is sound, and test coverage is thorough (44 new tests). No security issues. The only minor concern is that nothing explicitly prevents IPNS messages from having a non-None `ref` field, but such messages would just pass through the existing dependency check harmlessly.

**`src/aleph/handlers/content/store.py`** (line 323): Consider checking for `content.ref` in `_fetch_ipns` and raising `InvalidMessageFormat` if set, since the design doc and implementation assume `ref` is forbidden for IPNS content. Currently a rogue IPNS message with a ref would silently ignore it.

PR #1184, opened 2026-06-10
fix(storage): phase out unauthenticated upload

odesenfans:
## Summary

Anyone can currently push bytes into a CCN with no identity and no payment, and the worst offender is accidental: the two add_json endpoints accept up to ~100 MiB of anonymous JSON because they are only bounded by aiohttp's global client_max_size, 4x the 25 MiB limit someone deliberately set for anonymous add_file. Combined with a 24h grace period and a daily GC sweep, anonymous bytes can squat a node's disk for up to ~48h, renewable forever, at zero cost.

This is phase 1 (mitigation, no breaking changes) of a two-phase plan. Phase 2, announced by the deprecation signal added here, will require a signed message on every upload and remove the anonymous paths entirely once SDKs support the one-step flow.

### Changes

- Enforce the 25 MiB unauthenticated limit on both add_json endpoints with a streaming cap that aborts mid-body instead of buffering 100 MiB. Undecodable bodies (bad JSON, non-UTF-8) now return 422 instead of 500.
- Emit a `Deprecation: true` response header and an operator warning log on all four anonymous upload paths (storage/ipfs add_json, storage/ipfs add_file without signed metadata). Authenticated uploads are unaffected.
- Shorten default `storage.grace_period` 24h -> 6h and `storage.garbage_collector_period` 24h -> 4h. Worst-case lifetime of unclaimed anonymous bytes drops from ~48h to ~10h. Operators with explicit config overrides keep their values.
- Fix a data-loss race the shorter GC interval would have amplified: authenticated uploads (add_file, add_car) previously had NO pin row between upload and STORE message processing, so a GC sweep during a pending-queue backlog could delete a paying user's file. All uploads now get an upload-time grace pin bridging that gap.
- Treat an empty metadata part on /ipfs/add_file as unauthenticated (25 MiB cap + deprecation signal); previously it was processed without signature or balance check but received the authenticated 1 GiB cap.

### Notes for reviewers and node operators

- Behavior change: files orphaned by FORGET now become reclaimable after 6h instead of 24h (same `storage.grace_period` knob, accepted consequence).
- Knob asymmetry (pre-existing): both add_json endpoints use `storage.max_unauthenticated_upload_file_size`; `ipfs/add_file` uses `ipfs.max_unauthenticated_upload_file_size`. Tuning the ipfs knob does not affect /ipfs/add_json.
- Client-visible changes: 413 above 25 MiB on add_json (was accepted up to ~100 MiB), 422 on undecodable JSON (was 500). The `Deprecation` header uses the boolean draft form; the phase 2 PR can switch to RFC 9745 `Deprecation: @<timestamp>` plus `Sunset` once the removal date is fixed.
- The deprecation warning log fires once per successful anonymous upload; busy public nodes will see noticeable log volume, which is the point of the campaign.

## Test Plan

- [x] New tests: streaming 413 on both add_json endpoints (including chunked transfer without Content-Length), 422 on invalid JSON and non-UTF-8 bodies, Deprecation header present on all four anonymous paths and absent on authenticated ones, grace pin present after authenticated storage/ipfs/CAR uploads, empty-metadata regression test, config defaults pin test.
- [x] Full test suite green locally (786 passed) excluding tests/chains/test_ethereum.py, which requires a local anvil node and fails identically on main.


foxpatch-aleph (commented):

PR #1185, opened 2026-06-12
feat(api): time-range filtering for the credit history endpoint

odesenfans:
Adds optional start_date / end_date query params (Unix epoch seconds, inclusive bounds on message_timestamp) to GET /api/v0/addresses/{address}/credit_history. Applies to both pagination modes and to pagination_total. Epoch values are validated and converted to aware UTC datetimes in the pydantic schema, so malformed or out-of-range values (negative, non-numeric, inf, NaN) all return 422. Motivation: clients cannot currently answer 'what credit activity happened in the last N days' without paging through the entire history. message_timestamp is already indexed; no migration.

### DIFF
--- a/src/aleph/db/accessors/balances.py
+++ b/src/aleph/db/accessors/balances.py
@@ -847,6 +847,8 @@ def _apply_credit_history_filters(
     payment_method: Optional[str] = None,
     has_expiration: Optional[bool] = None,
     exclude_payment_method: Optional[List[str]] = None,
+    start_date: Optional[dt.datetime] = None,
+    end_date: Optional[dt.datetime] = None,
 ) -> Select:
     """Apply common filters to a credit history query."""
     if tx_hash is not None:
@@ -871,6 +873,10 @@ def _apply_credit_history_filters(
         query = query.where(
             AlephCreditHistoryDb.payment_method.notin_(exclude_payment_method)
         )
+    if start_date is not None:
+        query = query.where(AlephCreditHistoryDb.message_timestamp >= start_date)
+    if end_date is not None:
+        query = query.where(AlephCreditHistoryDb.message_timestamp <= end_date)
     return query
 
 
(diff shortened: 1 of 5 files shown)


foxpatch-aleph (dismissed):
This PR adds optional start_date/end_date query parameters (Unix epoch seconds) to the credit history endpoint, with proper validation, conversion to aware UTC datetimes, and filtering at the DB layer. The implementation is clean, correct, and well-tested across both the API and DB unit tests.

**`src/aleph/schemas/api/accounts.py`** (line 202): The `cls` parameter is unused; consider removing it or adding `@classmethod` for clarity. Works either way in Pydantic v2 but the current form is slightly misleading.

foxpatch-aleph (dismissed):
Well-structured PR that adds optional time-range filtering to the credit history endpoint. The epoch-to-datetime conversion is correctly validated with proper edge case handling, filters are properly threaded through all three accessor functions, and the test coverage is thorough with both API-level validation tests and DB-level filtering tests, including inclusive-bounds verification. A minor naming nit in one test and the lack of start_date <= end_date validation are the only observations, neither blocking.

**`tests/api/test_credit_history.py`** (line 44): The test name says "snake_case_time_params_also_accepted" but the assertion expects 422 because the value is negative. The comment explains the intent, but the name is misleading. Consider `test_credit_history_validates_snake_case_params` or similar.

**`src/aleph/schemas/api/accounts.py`** (line 197): Optional nits: consider whether to validate start_date <= end_date with a @model_validator. Currently an inverted range silently returns empty results. Not a correctness bug but could improve developer experience.

foxpatch-aleph (approved):
This PR adds optional start_date and end_date query parameters to the credit history endpoint, allowing clients to filter results within a time window without paginating through the full history. The implementation is clean and consistent: the DB accessor layer adds simple WHERE clause filters on message_timestamp, the Pydantic schema validates inputs (converting Unix epoch seconds to aware UTC datetimes, with date range inversion checks), and all three endpoint variants pass through the new parameters. Tests cover validation rejection, time filtering correctness, and inclusive boundary semantics. No issues found.

PR #1186, opened 2026-06-12
feat(api): direction (incoming/outgoing) filter for credit history

odesenfans:
Adds an optional direction=incoming|outgoing query param to GET /api/v0/addresses/{address}/credit_history, filtering on the sign of amount. payment_method alone cannot distinguish transfer-in from transfer-out and the set of top-up payment methods is open-ended, so a sign-based filter is the robust way to split top-ups from spending. Zero-amount entries match neither direction. Composes with the time filters: direction=incoming&startDate=... answers 'did this address top up in the last X days' in one request. Stacked on the time-filter PR (#1185).

### DIFF
--- a/src/aleph/db/accessors/balances.py
+++ b/src/aleph/db/accessors/balances.py
@@ -18,6 +18,7 @@ from aleph.toolkit.constants import (
     CREDIT_PRECISION_MULTIPLIER,
 )
 from aleph.toolkit.timestamp import timestamp_to_datetime, utc_now
+from aleph.types.credit import CreditFlow
 from aleph.types.db_session import DbSession
 from aleph.types.sort_order import SortByCreditHistory, SortOrder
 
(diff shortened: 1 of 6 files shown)


foxpatch-aleph (commented):

foxpatch-aleph (approved):
Well-scoped PR that adds a direction (incoming/outgoing) filter to the credit history endpoint. The implementation is clean — a new CreditFlow enum type, three-line filter in the shared query function, proper Pydantic validation, and thorough test coverage including the zero-amount edge case. No bugs or security concerns.

**`src/aleph/db/accessors/balances.py`** (line 882): The `if/elif` pattern naturally excludes zero-amount entries. Consider adding a brief inline comment: `# Zero-amount entries match neither direction` to make this non-obvious behavior explicit at a glance.

PR #1188, opened 2026-06-12
feat(api): resourceTypes filter for credit history (storage vs compute spend)

odesenfans:
Adds an optional `resourceTypes` query param to the credit history listing and summary endpoints: a comma-separated list of message types for the billed resource (STORE for storage, INSTANCE and PROGRAM for compute). An entry matches if its resource type is any of the listed values.

Expense rows reference the billed resource via `coalesce(nullif(origin, ''), origin_ref)`, the same effective-origin rule `get_total_consumed_credits` uses; the filter resolves it against `messages.item_hash` (PK) with an EXISTS subquery. Entries whose origin does not resolve to a known message (including forgotten messages) match no type.

Combined with the time-range filter, `resourceTypes=INSTANCE,PROGRAM&startDate=...` on the summary endpoint answers "compute spend this week" in one request, and `resourceTypes=STORE` the storage equivalent.

The param is named `resourceTypes` (plural, a list) following the messages endpoint's `msgTypes` convention. No migration.

### DIFF
--- a/src/aleph/db/accessors/balances.py
+++ b/src/aleph/db/accessors/balances.py
@@ -17,12 +17,17 @@ from typing import (
     Union,
 )
 
-from aleph_message.models import Chain
+from aleph_message.models import Chain, MessageType
 from sqlalchemy import case, func, select, text
 from sqlalchemy.dialects.postgresql import insert as pg_insert
 from sqlalchemy.sql import ColumnElement, Select
 
-from aleph.db.models import AlephBalanceDb, AlephCreditBalanceDb, AlephCreditHistoryDb
+from aleph.db.models import (
+    AlephBalanceDb,
+    AlephCreditBalanceDb,
+    AlephCreditHistoryDb,
+    MessageDb,
+)
 from aleph.toolkit.constants import (
     CREDIT_PRECISION_CUTOFF_TIMESTAMP,
     CREDIT_PRECISION_MULTIPLIER,
@@ -861,6 +866,7 @@ def _apply_credit_history_filters(
     start_date: Optional[dt.datetime] = None,
     end_date: Optional[dt.datetime] = None,
     direction: Optional[CreditFlow] = None,
+    origin_type: Optional[MessageType] = None,
 ) -> Select:
     """Apply common filters to a credit history query."""
     if tx_hash is not None:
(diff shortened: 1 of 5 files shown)


foxpatch-aleph (approved):
Clean PR that adds an optional originType filter to the credit history listing and summary endpoints. The filter uses an EXISTS subquery with the same effective-orion resolution pattern used elsewhere in the codebase, correctly matching expenses against the message type of their billed resource. Tests cover invalid/invalid values, empty results, filtering both STORE and INSTANCE entries, composition with the summary aggregate, and the unresolvable case. No bugs or issues found.

**`src/aleph/web/controllers/accounts.py`** (line 847): The OpenAPI enum is hardcoded — if MessageType is ever extended the docs will drift. Consider deriving from `MessageType` enum values. Non-blocking as this follows the same pattern used for `direction` elsewhere.

**`src/aleph/web/controllers/accounts.py`** (line 1126): Same enum hardcoding pattern as the listing endpoint, same consideration applies.

foxpatch-aleph (dismissed):
Cleanly adds an origin_type filter to credit history endpoints. The EXISTS subquery against MessageDb is correct and consistent with the existing effective_origin pattern. Tests cover the filter, summary composition, validation, and edge cases. Only minor nits.

**`src/aleph/schemas/api/accounts.py`** (line 15): `model_validator` is imported but appears unused in the shown diff — it's used at line 220 for the date range validator. This import is correct, just noting it's not dead code.

foxpatch-aleph (dismissed):
Well-structured PR that adds an originType filter to the credit history endpoint. The implementation correctly reuses the existing effective-origin logic (coalesce(nullif(origin,''), origin_ref)), composes with all other filters, and handles edge cases like unresolvable entries. Tests cover both API-level validation and DB-level filtering. Clean, minimal code changes throughout.

**`tests/db/test_credit_balances.py`** (line 2855): Consider adding a brief comment explaining that an unresolvable row (both ref='' and execution_id='') produces effective_origin='', which doesn't match any message item_hash, so it correctly matches no origin_type filter.

foxpatch-aleph (approved):
Cleanly implements an optional resourceTypes query parameter for credit history listing and summary endpoints. The filter uses an EXISTS subquery with the effective_origin coalesce pattern already established in the codebase. Tests comprehensively cover both the API layer (invalid values, valid empty results, end-to-end filtering) and the DB accessor layer (multi-type matching, unresolvable origin exclusion). The Pydantic validator reuse is elegant. No bugs or security issues found.

PR #1190, opened 2026-06-16
perf(db): composite index for credit-history address+timestamp queries

odesenfans:
## Problem

A production slow-query log showed the credit-history query taking **~10s**:

```
duration: 10249.229 ms  statement: SELECT credit_history.* FROM credit_history
WHERE credit_history.address = '0x...'
ORDER BY credit_history.message_timestamp ASC, credit_history.credit_ref ASC, credit_history.credit_index ASC
```

This is the default invocation of `get_address_credit_history()` (`GET /api/v1/accounts/{address}/credit-history`): filter by `address`, order by `(message_timestamp, credit_ref, credit_index)`, no LIMIT.

`credit_history` only had single-column indexes on `address` and `message_timestamp`, so the planner had to either sort the entire per-address result set or scan the whole `message_timestamp` index filtering by address. Both are slow on high-volume addresses.

## Fix

Add a composite btree index on `(address, message_timestamp, credit_ref, credit_index)`. The query becomes a single index range scan on `address` with rows already in sort order, eliminating the sort node. A single index serves both ASC and DESC because all sort keys share direction.

The new index is a left-prefix superset of `ix_credit_history_address`, so that standalone index is dropped in the same migration.

## Changes

- `src/aleph/db/models/balances.py`: drop `index=True` on `address`, add the composite `Index` via `__table_args__`.
- `deployment/migrations/versions/0060_a1d4e7b2c9f6_*.py`: `CREATE INDEX CONCURRENTLY` the composite index + `DROP INDEX CONCURRENTLY ix_credit_history_address` (downgrade reverses both). Follows the autocommit pattern from migration 0058.

## Testing

- `tests/api/test_credit_history.py` (17), `tests/db/test_credit_balances.py` + `tests/api/test_cursor_pagination.py` (64): all pass. The suite runs the full alembic chain, so the migration applies cleanly.
- Formatting (black/isort) and ruff: clean. Alembic: single head `a1d4e7b2c9f6`.

## Out of scope

- Non-default `sort_by` values (expiration_date, amount, etc.) still sort; rare enough to defer.
- The `pagination=0` default still returns all rows for an address (separate API-behavior concern).

### DIFF
--- a/deployment/migrations/versions/0060_a1d4e7b2c9f6_credit_history_address_timestamp_index.py
+++ b/deployment/migrations/versions/0060_a1d4e7b2c9f6_credit_history_address_timestamp_index.py
@@ -1,0 +1,83 @@
+"""Add composite index credit_history(address, message_timestamp, credit_ref, credit_index)
+
+Revision ID: a1d4e7b2c9f6
+Revises: b9c4f1e6a2d7
+Create Date: 2026-06-16
+
+The credit-history endpoint filters by address and orders by
+(message_timestamp, credit_ref, credit_index) - the default sort. With only
+single-column indexes on address and message_timestamp, the planner either
+sorts the full per-address set or scans the whole message_timestamp index
+filtering by address, costing ~10s on high-volume addresses.
+
+This composite index serves the filter and the full sort key in one index
+range scan; the single btree covers both ASC and DESC because all sort keys
+share direction. It is a left-prefix superset of the standalone
+ix_credit_history_address index, which is therefore dropped here.
+"""
+
+import logging
+
+from alembic import op
+from sqlalchemy import text
+
+revision = "a1d4e7b2c9f6"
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
Clean, well-scoped performance fix. Adds a composite btree index on (address, message_timestamp, credit_ref, credit_index) to serve the default credit-history query in a single index range scan, replacing the standalone address index. The migration safely uses CONCURRENTLY DDL with the autocommit pattern established by migration 0058. Code changes are minimal and correct.

**`src/aleph/db/models/balances.py`** (line 80): The comment mentions the index covers both ASC and DESC because all sort keys share direction — this is accurate. Note that the default sort_order in the accessor is actually DESC, not ASC as the PR description says, but the index still works for both since all three sort components flip together.

PR #1200, opened 2026-06-18
feat(p2p): gRPC IPC client for P2P service 2.0

odesenfans:
P2P service overhaul. Replaces the RabbitMQ + HTTP control split with a thin async gRPC client to the p2p-service. Fully wire-compatible: gossipsub, topics and the message format are unchanged, so old and new nodes interoperate during the rollout.

## What changes
- New `P2PGrpcClient` (generated stubs + a small async wrapper) replacing the `aleph-p2p-client` package. Keepalive, `wait_for_ready` at boot, deterministic stream cleanup, structured error mapping (wrong-peer / unreachable on dial).
- All call sites ported to gRPC: `pub_p2p`, the incoming pubsub channel, the dial/reconnect paths. RabbitMQ leaves the network-layer boundary (it stays for pyaleph's internal job queues); the API server opens its own MQ connection via `make_mq_conn`.
- `Publish(echo=true)` replaces the client-side loopback hack with an explicit service-side contract.
- New `refresh_preferred_peers_job`: pushes the active CCN set from the CORECHANNEL aggregate to the service as preferred peers (multiaddr-validated, self-filtered, tolerant of a service that does not implement it yet).
- Reconnect backoff in the consumer loops (gRPC fails fast where AMQP used to auto-reconnect).

## Operator notes
- Requires `alephim/p2p-service >= 2.0.0` (PRs aleph-im/p2p-service#67, #68). The compose samples are bumped accordingly and the p2p-service no longer depends on RabbitMQ.
- Pubsub durability changes: the broker-buffered exchange becomes a best-effort gRPC stream (server-side drop-oldest, 1024 cap). Missed messages during a reconnect are recovered by the normal sync paths, same as a gossipsub gap today.
- Prometheus scrape target for the service moves from the old control port to `metrics_port` (4040).

## Notes
- `protobuf` is held on the 6.x line (`>=6.33.5,<7`): the committed gRPC stubs are protobuf-6 gencode and the runtime guard requires a major-6 runtime. Lifting it needs a protobuf-7 `grpcio-tools`, which does not exist yet (latest 1.81.1 still pins protobuf<7).

## Testing
gRPC client unit tests (in-process server) and preferred-peers job tests. Full suite green; validated against the current pinned dependency set in a clean install.


foxpatch-aleph (dismissed):
Clean, well-structured PR that replaces the RabbitMQ+HTTP-based P2P client with a thin async gRPC client. The `P2PGrpcClient` wraps the generated stubs with proper error mapping (DialWrongPeerException/DialFailedException), deterministic stream cleanup, and keepalive configuration. All call sites are ported — the old `aleph-p2p-client` dependency is completely removed. The new `refresh_preferred_peers_job` extracts CCNs from the corechannel aggregate and pushes them to the service. Tests are thorough with unit tests for all RPCs and job-level tests for the preferred peers logic. Linting/formatter exclusions are correctly configured for generated code. No bugs found, no security issues.

**`src/aleph/services/p2p/client.py`** (line 86): Minor: the local `stub` variable is created but never used — `self._stub` in `__init__` is created from `channel` directly. This is dead code but harmless. Consider removing the redundant `stub = ...` line.

**`src/aleph/services/p2p/protocol.py`** (line 20): The `seen_hashes` deque has a fixed `maxlen=200000` with no time-based eviction. In high-throughput or long-running scenarios, this could fill up with old hashes reducing its effectiveness. Not a regression (same as before), but worth monitoring.

**`tests/services/p2p/test_grpc_client.py`** (line 181): Minor: the `assert isinstance(exc_info.value, Exception)` on line 184 is redundant since `grpc.aio.AioRpcError` always extends `Exception`. Not harmful, but unnecessary.

foxpatch-aleph (approved):
Well-structured migration from RabbitMQ+HTTP to gRPC for the P2P service. The P2PGrpcClient is cleanly designed with proper error mapping, the generated stubs are correctly excluded from linting, and the new tests cover the client thoroughly with an in-process gRPC server. Minor observations include an inconsistency in how `connect` vs `identify` handle errors (BaseException vs raw propagation), a testing gap in `test_job_survives_rpc_failure` that doesn't verify the failed call was actually attempted, and the lack of dedicated tests for the `incoming_channel` and `monitor_hosts_p2p` loops. None of these are blocking.

**`src/aleph/services/p2p/client.py`** (line 97): `identify()` lacks the error/cleanup guard that `connect()` has (lines 87-93). If the service becomes unreachable between `connect` and an `identify()` call, the raw AioRpcError propagates. The docstring says this is allowed, so not a bug — but the inconsistency is worth noting.

**`tests/services/p2p/test_preferred_peers.py`** (line 233): `test_job_survives_rpc_failure` asserts `client.calls[0]` after recovery but doesn't verify that the first (failed) call was actually attempted. Since `FailOnceP2PClient` raises before recording to `self.calls`, you can't distinguish "failed once then succeeded" from "never failed, only ran once". Consider adding a `failures` counter to the mock to make this assertion more precise.

**`src/aleph/services/p2p/protocol.py`** (line 63): The exponential backoff here (0.1 → 10.0) and in `monitor.py` (2.0 → 10.0) uses different starting values but the same cap and multiplier. Consider whether the backoff should be configurable or referenced from a single constant, but this is a nit and not blocking.

PR #1201, opened 2026-06-18
feat(p2p): fetch protocol, P2P-only broadcast, ALIVE consume stop (release N+1)

odesenfans:
Release N+1 of the P2P overhaul (sections 4.4-4.6 of `docs/plans/2026-06-09-p2p-overhaul-design.md`).

**Stacked on #1200** (release N). Base is `feat/p2p-grpc-client` so this diff shows only the N+1 changes; review and merge #1200 first, then this retargets to `main` automatically.

## What changes
- **Content fetch over `/aleph/fetch/1.0.0`.** The storage fetch order becomes local -> P2P `Fetch` RPC -> IPFS -> legacy HTTP. Fetched bytes are hash-verified before use and a hash mismatch falls through to the next source. Against a p2p-service without the fetch RPC the call returns `UNIMPLEMENTED` and degrades silently, so rolling upgrades are safe in both orders. The gRPC client is wired into every `StorageService` instance (job processes connect best-effort).
- **Broadcast is now P2P-required.** IPFS pubsub publishing is removed (consumption stays for pre-N nodes). API-visible: a P2P publish failure now returns `error` / HTTP 500 where it previously degraded to `warning` / 200 via IPFS, and this also fixes a latent bug where `/api/v0/p2p/pubsub/pub` always returned 200 on failure.
- **ALIVE is no longer consumed** (still emitted on the P2P topic for old nodes). The legacy HTTP-fallback `api_servers` list is now built from the active CCN registry instead of the ALIVE-fed peers table.
- **Deployment:** p2p-service pinned to v2.0 with `content_dir` + `ipfs_api_url` (the service reads pyaleph's content store read-only and the local Kubo for CIDs, replacing the HTTP callback), plus a persistent peerstore volume.

## Operator notes
- Set `p2p.ipfs_api_url` to the node that holds pinned content (the pinning node if `ipfs.pinning` offloads storage; the local daemon otherwise).
- A p2p-service outage now rejects message submissions instead of silently accepting them without broadcasting. Plan accordingly for restarts.

## Rollout
N+1 behavioral changes (stop IPFS pubsub publish, stop ALIVE consume) should ship only after the majority of active CCNs run N, per the design's gate. A successful fetch transfer requires both peers to speak `/aleph/fetch/1.0.0`; against the current all-old fleet, fetch requests fall through to IPFS as designed.

## Testing
Storage fetch chain, P2P-only broadcast, and the pub endpoints validated; preferred-peers/api_servers jobs covered. Full targeted suite green, validated against the bumped dependency set (aiohttp 3.14.1, sqlalchemy 2.0.51, pydantic 2.13.4, ...) in a clean install.


foxpatch-aleph (approved):
Well-structured PR delivering the N+1 release of the P2P overhaul. The fetch chain (local -> P2P gRPC Fetch -> IPFS -> legacy HTTP) is correct with proper hash verification and fall-through. ALIVE consumption removal is clean and the replacement of the peers-table-based api_servers with the corechannel aggregate is architecturally sound. P2P-only broadcast is a proper simplification. Tests cover the new functional paths well. Minor suggestions only, no blocking issues.

**`src/aleph/storage.py`** (line 113): Minor: the `if not chunks: return None` guard at line 140 is unreachable — `b"".join([])` already produces an empty bytes object which is falsy, so the `return content if content else None` at line 143 handles the empty case identically. Not a bug, just dead code.

**`src/aleph/storage.py`** (line 266): The legacy HTTP fallback sets `source = ContentSource.P2P`, which is the same enum value used for the new gRPC P2P fetch at line 247. Consider a different source value (or a log message) to distinguish these paths operationally before the legacy code is removed in N+2. Non-blocking since this is slated for deletion.

PR #1203, opened 2026-06-22
fix(api): catch base ConnectionError when streaming WS history

odesenfans:
## Problem

On 0.10.2, the messages WebSocket endpoint logs `[ERROR] Error handling request` with a `ConnectionError: Connection lost` (direct cause `BrokenPipeError`) whenever a client disconnects while the server is still streaming the `history` backlog. These benign client disconnects get logged as server errors and forwarded to Sentry, drowning out real errors.

```
File "/opt/pyaleph/src/aleph/web/controllers/messages.py", line 655, in _send_history_to_ws
    await ws.send_str(aleph_json.dumps(msg_dict).decode("utf-8"))
  ...
ConnectionError: Connection lost
```

## Root cause

`messages_ws` wrapped the history send in `except ConnectionResetError`. But when the peer is gone, aiohttp's `_drain_helper` raises a **bare** `ConnectionError("Connection lost")`. In Python's hierarchy:

```
OSError → ConnectionError → {ConnectionResetError, BrokenPipeError, ...}
```

`ConnectionResetError` and `BrokenPipeError` are *siblings*; both derive from `ConnectionError`. A bare `ConnectionError` is the **parent**, so `except ConnectionResetError` does not catch it — the exception escapes and is logged as a 500.

## Change

Catch the base `ConnectionError` in the history path (the actual fix). The two other WS send paths already caught `(ConnectionResetError, ConnectionError)`; simplified those to the base class for consistency (no behavior change):

- `messages.py` — history send (the bug)
- `messages.py` — live broadcast send (cleanup)
- `main.py` — metrics WS send (cleanup)

## Test

Adds `test_ws_history_connection_error_is_swallowed`, which drives `messages_ws` with a mocked history send that raises the exact bare `ConnectionError("Connection lost")` and asserts the handler returns cleanly without registering the client (would fail against the old `except ConnectionResetError`).

> Note: tests run in CI — local hatch/venv environments are currently broken (stale SQLAlchemy 1.4 in `.venv`; the hatch testing env can't build a native Rust dep).

### DIFF
--- a/src/aleph/web/controllers/main.py
+++ b/src/aleph/web/controllers/main.py
@@ -139,7 +139,7 @@ class StatusBroadcaster:
         try:
             await ws.send_json(payload)
             return True
-        except (ConnectionResetError, ConnectionError):
+        except ConnectionError:
             return False
 
 
--- a/src/aleph/web/controllers/messages.py
+++ b/src/aleph/web/controllers/messages.py
@@ -300,7 +300,7 @@ class MessageBroadcaster:
             payload = json_no_content if client.exclude_content else json_full
             await client.ws.send_str(payload)
             return True
-        except (ConnectionResetError, ConnectionError):
+        except ConnectionError:
             return False
 
 
@@ -750,7 +750,7 @@ async def messages_ws(request: web.Request) -> web.WebSocketResponse:
                     history=history,
                     query_params=query_params,
                 )
-            except ConnectionResetError:
+            except ConnectionError:
                 LOGGER.info("Could not send history, aborting message websocket")
                 return ws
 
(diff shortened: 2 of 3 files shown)


foxpatch-aleph (approved):
The PR correctly fixes a real bug: the history WS send path caught only ConnectionResetError, but aiohttp raises a bare ConnectionError("Connection lost") on peer disconnect, which is the parent class and was not caught. Widening to except ConnectionError is the right fix. The two cleanup changes (removing redundant ConnectionResetError from (ConnectionResetError, ConnectionError) tuples) are no-ops since ConnectionError already subsumes ConnectionResetError. The test is well-targeted: it mocks _send_history_to_ws to raise the exact bare ConnectionError, asserts the handler returns ws cleanly, and verifies broadcaster.add was never called (confirming early return). The test would fail against the old except ConnectionResetError, validating it actually exercises the bug. All three WS send paths in the codebase are now consistently handling ConnectionError.

PR #1205, opened 2026-06-24
feat(api): resource filter for credit history and summary endpoints

odesenfans:
## Summary

Adds an optional `resource` query filter to the credit history listing and summary endpoints, to make it easy to inspect all credit spend tied to a specific resource (a VM or a file) when debugging.

- `GET /api/v0/addresses/{address}/credit_history?resource=<hash>`
- `GET /api/v0/addresses/{address}/credit_history/summary?resource=<hash>`

An entry matches when its **effective origin** - `coalesce(nullif(origin, ''), origin_ref)` - equals the given hash. This is the same effective-origin rule the existing `resourceTypes` filter and `get_total_consumed_credits` use, so an expense matches by the resource's message hash regardless of whether the hash sits in `origin` (execution_id-style) or `origin_ref` (volume-style).

The filter is named `resource` rather than `vmHash` because it matches any billed resource message hash (VMs and files alike), not just VMs.

Unlike `resourceTypes`, this filter compares the credit-history row's own columns and does **not** join against `messages`, so entries whose resource message has been forgotten still match - which is exactly what you want when debugging. It AND-composes with the existing filters (`direction`, time range, `resourceTypes`, ...). No DB migration.

### Changes
- `schemas/api/accounts.py`: `resource` field on the shared `CreditHistoryFilterParams` (inherited by both endpoints).
- `db/accessors/balances.py`: `resource` filter in `_apply_credit_history_filters`, threaded through `get_address_credit_history`, `count_address_credit_history`, and `get_address_credit_history_summary`.
- `web/controllers/accounts.py`: wired through all four accessor call sites (listing first-page, cursor pagination, count, summary) + OpenAPI docs on both handlers.

## Test Plan
- [x] Accessor test: matches hash in `origin` and in `origin_ref`, excludes unrelated entries, summary aggregation + count, no-match case (`tests/db/test_credit_balances.py`).
- [x] Endpoint tests: listing + summary filtering on both endpoints, composition with `direction`, valid-but-no-data -> 404/zeros (`tests/api/test_credit_history.py`).
- [x] Full `tests/api/test_credit_history.py` + `tests/db/test_credit_balances.py` green.
- [x] `linting:all` clean (ruff, black, isort, mypy).

### DIFF
--- a/src/aleph/db/accessors/balances.py
+++ b/src/aleph/db/accessors/balances.py
@@ -867,6 +867,7 @@ def _apply_credit_history_filters(
     end_date: Optional[dt.datetime] = None,
     direction: Optional[CreditFlow] = None,
     resource_types: Optional[List[MessageType]] = None,
+    vm_hash: Optional[str] = None,
 ) -> Select:
     """Apply common filters to a credit history query."""
     if tx_hash is not None:
@@ -912,6 +913,14 @@ def _apply_credit_history_filters(
             )
             .exists()
         )
+    if vm_hash is not None:
+        query = query.where(
+            func.coalesce(
+                func.nullif(AlephCreditHistoryDb.origin, ""),
+                AlephCreditHistoryDb.origin_ref,
+            )
+            == vm_hash
+        )
     return query
 
 
@@ -933,6 +942,7 @@ def get_address_credit_history(
     end_date: Optional[dt.datetime] = None,
     direction: Optional[CreditFlow] = None,
     resource_types: Optional[List[MessageType]] = None,
+    vm_hash: Optional[str] = None,
     sort_by: SortByCreditHistory = SortByCreditHistory.MESSAGE_TIMESTAMP,
     sort_order: SortOrder = SortOrder.DESCENDING,
     after_sort_value: Optional[Any] = None,
@@ -989,6 +999,7 @@ def get_address_credit_history(
         end_date=end_date,
         direction=direction,
         resource_types=resource_types,
+        vm_hash=vm_hash,
     )
 
     # Cursor-based keyset pagination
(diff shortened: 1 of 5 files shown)


foxpatch-aleph (dismissed):
Clean, well-structured PR that adds an optional vmHash filter to the credit history listing and summary endpoints. The effective-origin logic (coalesce(nullif(origin, ''), origin_ref)) is consistent with the existing get_total_consumed_credits function and the resourceTypes filter. The parameter is correctly threaded through all four accessor call sites (cursor pagination, first-page listing, count, summary) and both endpoint handlers. SQLAlchemy parameterizes the value automatically, so there is no SQL injection risk. Test coverage is thorough: accessor tests verify matching via both the origin and origin_ref columns, no-match cases, summary aggregation, and count; API tests cover both endpoints, composition with the direction filter, and the valid-but-no-data edge case. The vm_hash field on CreditHistoryFilterParams follows the same Optional[str] + alias convention as the existing filters.

foxpatch-aleph (approved):
The PR correctly adds an optional `resource` query filter to the credit history listing and summary endpoints, using the same `coalesce(nullif(origin, ''), origin_ref)` effective-origin pattern already established by the `resource_types` filter and `get_total_consumed_credits`. The filter is properly threaded through all four accessor call sites (cursor pagination, first-page, count, summary), uses parameterized SQL (no injection risk), and AND-composes with existing filters. Test coverage is thorough at both the DB accessor and API levels, covering both origin columns, no-match cases, summary aggregation, and filter composition. No bugs, security issues, or missing coverage identified.

**`src/aleph/schemas/api/accounts.py`** (line 205): Minor: The `resource` field has no `alias` (unlike `resource_types` which uses `alias="resourceTypes"`). Since `resource` is a single word, no camelCase alias is needed, so this is fine — just noting for consistency awareness. The PR description title still references `vmHash` but the code correctly uses `resource` per the latest refactor commit.

PR #1209, opened 2026-06-30
refactor(api): drop the global client_max_size override, cap uploads per-route

odesenfans:
Stacked on #1208 (base is its branch; retarget to `main` once #1208 merges). It depends on #1208: dropping the global cap is only safe once the IPFS routes already clone up.

## Why

`client_max_size` is the body limit for every endpoint that reads the body buffered (`request.json()` / `.read()` / `.post()`), not just uploads. Wiring it to `storage.max_file_size` (100 MiB) meant:
- the unauthenticated `/storage/add_json` and `/ipfs/add_json` endpoints buffered up to 100 MiB of JSON in memory per request;
- unrelated limits were coupled (this is what made `ipfs.max_upload_file_size` unreachable, see #1208).

## Change

- `create_aiohttp_app` defaults to aiohttp's own 1 MiB `client_max_size`; `api_entrypoint` stops passing `storage.max_file_size`.
- Endpoints that accept larger bodies raise it per-route with `Request.clone()`:
  - `storage_add_file` -> `storage.max_file_size` (unchanged effective limit).
  - `add_storage_json` / `add_ipfs_json` -> `max_unauthenticated_upload_file_size`. These are unauthenticated and buffered in memory, so this also **tightens** them from 100 MiB to 25 MiB. Flagging the number explicitly: 25 MiB matches the existing unauthenticated content limit and is generous for JSON, but say so if you want a different value.
  - `/ipfs/add_file`, `/ipfs/add_car` already clone (#1208).
- Message / pubsub / prices endpoints keep no explicit override: their bodies are bounded by `MAX_INLINE_SIZE` (200 KB), well under 1 MiB.

Net effect: every body-size limit is now explicit and per-route, the default is aiohttp's conservative 1 MiB, and the unauthenticated JSON endpoints are no longer a 100 MiB in-memory vector.

## Tests

`tests/api/test_storage.py`:
- `test_add_storage_json_above_global_cap`: JSON above the 1 MiB global but within the unauthenticated limit is accepted. `request.json()` is gated by `client_max_size` on every aiohttp version, so this actively guards the lift even on the pinned 3.13.x.
- `test_add_storage_json_above_unauthenticated_cap`: JSON above 25 MiB is rejected.
- `test_storage_add_file_above_global_cap`: a multipart upload above the 1 MiB global succeeds.

Full `tests/api/` suite: 237 passed, 6 skipped.

### DIFF
--- a/src/aleph/api_entrypoint.py
+++ b/src/aleph/api_entrypoint.py
@@ -61,9 +61,11 @@ async def configure_aiohttp_app(
         )
         signature_verifier = SignatureVerifier()
 
-        app = create_aiohttp_app(
-            max_file_size=config.storage.max_file_size.value,
-        )
+        # Use aiohttp's small default client_max_size for the app. Upload
+        # endpoints raise it per-route via Request.clone() (see ipfs.py /
+        # storage.py); wiring it to storage.max_file_size would have exposed
+        # every buffered endpoint to that size in memory.
+        app = create_aiohttp_app()
 
         # Reuse the connection of the P2P client to avoid opening two connections
         mq_conn = p2p_client.mq_client.connection
--- a/src/aleph/api_entrypoint.py
+++ b/src/aleph/api_entrypoint.py
@@ -61,10 +61,6 @@ async def configure_aiohttp_app(
         )
         signature_verifier = SignatureVerifier()
 
-        # Use aiohttp's small default client_max_size for the app. Upload
-        # endpoints raise it per-route via Request.clone() (see ipfs.py /
-        # storage.py); wiring it to storage.max_file_size would have exposed
-        # every buffered endpoint to that size in memory.
         app = create_aiohttp_app()
 
         # Reuse the connection of the P2P client to avoid opening two connections
--- a/src/aleph/web/__init__.py
+++ b/src/aleph/web/__init__.py
@@ -9,9 +9,15 @@ import jinja2
 from aiohttp import web
 from aiohttp_swagger3 import SwaggerDocs, SwaggerInfo, SwaggerUiSettings
 
-from aleph.toolkit.constants import DEFAULT_MAX_FILE_SIZE
 from aleph.version import get_version
 from aleph.web.controllers.routes import register_routes
+
+# Default app-wide aiohttp client_max_size, capping every request body read via
+# request.json()/.read()/.post(). Endpoints that accept larger uploads raise it
+# per-route with Request.clone() (see ipfs.py / storage.py). Kept at aiohttp's
+# own default (1 MiB) so buffered endpoints are not exposed to large in-memory
+# bodies; message submissions are bounded well below this by MAX_INLINE_SIZE.
+DEFAULT_CLIENT_MAX_SIZE = 1024 * 1024
 
 
 def init_cors(app: web.Application):
@@ -35,7 +41,7 @@ def init_cors(app: web.Application):
 
 def create_aiohttp_app(
     with_swagger: bool = True,
-    max_file_size: int = DEFAULT_MAX_FILE_SIZE,
+    max_file_size: int = DEFAULT_CLIENT_MAX_SIZE,
 ) -> web.Application:
     app = web.Application(client_max_size=max_file_size)
 
--- a/src/aleph/web/controllers/storage.py
+++ b/src/aleph/web/controllers/storage.py
@@ -86,7 +86,14 @@ async def add_ipfs_json_controller(request: web.Request):
     config = get_config_from_request(request)
     grace_period = config.storage.grace_period.value
 
-    data = await request.json()
+    # Unauthenticated JSON upload, buffered fully in memory by request.json().
+    # The app-wide client_max_size is now aiohttp's small default, so lift it to
+    # this endpoint's own cap (the unauthenticated content limit) for the body
+    # read only. The original request keeps the app state used afterwards.
+    body_request = request.clone(
+        client_max_size=config.ipfs.max_unauthenticated_upload_file_size.value
+    )
+    data = await body_request.json()
     with session_factory() as session:
         output = {
             "status": "success",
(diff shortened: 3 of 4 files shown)


foxpatch-aleph (dismissed):
This PR correctly drops the global client_max_size override (previously wired to storage.max_file_size = 100 MiB) and replaces it with aiohttp's conservative 1 MiB default, lifting the cap per-route via request.clone() only on endpoints that need larger bodies. The clone pattern is used correctly: the original request is retained for app-state-dependent operations (broadcast_and_process_message, getters), while the clone is used solely for body reads. The add_ipfs_json_controller and add_storage_json_controller correctly use their respective config sections (config.ipfs / config.storage) for max_unauthenticated_upload_file_size, both defaulting to 25 MiB. This tightens unauthenticated JSON endpoints from 100 MiB to 25 MiB of in-memory buffering, a meaningful security improvement. Tests cover the storage JSON endpoint lift (2 MiB accepted), rejection (26 MiB rejected), and multipart upload above the global cap. The only gaps are minor: no equivalent cap-lift test for the IPFS JSON endpoint (/api/v0/ipfs/add_json), and the 26 MiB rejection test buffers a large payload in memory which is slower than ideal. Neither is blocking.

**`src/aleph/web/controllers/storage.py`** (line 94): This endpoint uses `config.ipfs.max_unauthenticated_upload_file_size` while `add_storage_json_controller` uses `config.storage.max_unauthenticated_upload_file_size`. Both default to 25 MiB so this is correct, but worth noting the asymmetry — if the two config values ever diverge, the IPFS JSON endpoint will follow the IPFS config and the storage JSON endpoint will follow the storage config, which is the intended behavior.

**`tests/api/test_storage.py`** (line 844): Missing test coverage: the `add_ipfs_json_controller` endpoint (IPFS_ADD_JSON_URI) received the same `request.clone()` treatment but has no equivalent above-global-cap / above-unauthenticated-cap test. Consider adding a parallel test pair for `/api/v0/ipfs/add_json` to guard the per-route lift on that endpoint as well.

**`tests/api/test_storage.py`** (line 850): Sending 26 MiB of JSON to test the 413 rejection path works but buffers the entire payload in memory on both client and server side, making this test slower than necessary. An alternative would be to lower `max_unauthenticated_upload_file_size` in the test fixture to a small value (e.g. 1 MiB) and send a slightly larger body, achieving the same assertion with far less memory. Not blocking.

**`src/aleph/web/__init__.py`** (line 44): The parameter name `max_file_size` is now misleading since it controls `client_max_size`, not a file size limit. Renaming it to `client_max_size` would improve clarity but would require updating all callers (api_entrypoint.py, conftest.py, test_ipfs.py). Not blocking since there are only three callers, but worth considering in a follow-up.

foxpatch-aleph (dismissed):
Correct and well-motivated refactor. The global client_max_size is dropped to aiohttp's 1 MiB default and lifted per-route via Request.clone(client_max_size=...) on the three body-reading endpoints. I verified the clone() kwarg is valid in the installed aiohttp, that each handler obtains app-state getters from the original request before cloning (so the clone is used solely for body reads), that the 25 MiB unauthenticated caps are consistently applied to both /storage/add_json and /ipfs/add_json, and that the existing create_aiohttp_app(max_file_size=...) call site in tests/api/test_ipfs.py still compiles. The new tests actively guard the per-route lift (request.json() is gated by client_max_size on all aiohttp versions) and confirm the cap is tightened rather than removed. Two minor non-blocking notes: (1) no test was added for /api/v0/ipfs/add_json, which received the same treatment as /api/v0/storage/add_json but sources its cap from config.ipfs rather than config.storage; (2) the small_global_cap_client fixture docstring in tests/api/test_ipfs.py:1176-1179 is now stale (it claims api_entrypoint wires client_max_size to storage.max_file_size, which this PR removes). Neither blocks merge.

**`tests/api/test_storage.py`** (line 830): Coverage gap: this test guards /api/v0/storage/add_json, but /api/v0/ipfs/add_json received the same clone() treatment in src/aleph/web/controllers/storage.py:93-96 and sources its cap from config.ipfs.max_unauthenticated_upload_file_size (a different config key). Consider adding a mirror test against IPFS_ADD_JSON_URI so a future config divergence between config.ipfs and config.storage unauthenticated caps is caught.

**`tests/api/test_ipfs.py`** (line 1176): Stale docstring: "api_entrypoint wires the app-wide client_max_size to storage.max_file_size via create_aiohttp_app" is no longer true after this PR (api_entrypoint now uses aiohttp's 1 MiB default). Worth updating to match the new wiring to avoid misleading future readers.

src/aleph/api_entrypoint.py
```diff
@@ -61,9 +61,11 @@ async def configure_aiohttp_app(
...
         signature_verifier = SignatureVerifier()
 
-        app = create_aiohttp_app(
-            max_file_size=config.storage.max_file_size.value,
-        )
+        # Use aiohttp's small default client_max_size for the app. Upload
+        # endpoints raise it per-route via Request.clone() (see ipfs.py /
+        # storage.py); wiring it to storage.max_file_size would have exposed
+        # every buffered endpoint to that size in memory.
+        app = create_aiohttp_app()
```
odesenfans:
```suggestion
        app = create_aiohttp_app()
```

foxpatch-aleph (approved):
Sound security hardening: drops the global client_max_size from 100 MiB to aiohttp's 1 MiB default and lifts it per-route via request.clone() for the three affected endpoints. The pattern is consistent with #1208, the config selection is correct (ipfs vs storage sections), and the storage_add_file two-tier check is preserved. The unauthenticated JSON endpoints are tightened from 100 MiB to 25 MiB, closing an in-memory DoS vector. Tests guard the new caps for the storage endpoints. Minor nits: a stale comment in ipfs.py referencing the old wiring, no symmetric test for the add_ipfs_json lift, and the legacy max_file_size parameter name in create_aiohttp_app.

**`src/aleph/web/controllers/ipfs.py`** (line 103): Stale after this PR: the comment says the global cap is "wired to storage.max_file_size", but this PR changes the default to 1 MiB (DEFAULT_CLIENT_MAX_SIZE). Consider updating to reflect the new default so future readers aren't misled.

**`src/aleph/web/__init__.py`** (line 44): The parameter is still named `max_file_size` but now controls `client_max_size`, not a file size. The default is 1 MiB (a body buffer cap), not 100 MiB. Renaming to `client_max_size` would clarify intent, though it would require updating the caller in tests/api/test_ipfs.py:1181. Not blocking.

**`tests/api/test_storage.py`** (line 829): There's no symmetric test for the `add_ipfs_json` endpoint (IPFS_ADD_JSON_URI), which received the same clone() lift. A test mirroring test_add_storage_json_above_global_cap for /ipfs/add_json would ensure the ipfs config path is also guarded against regressions.

PR #1210, opened 2026-06-30
feat(api): expose upload size limits at GET /storage/limits.json

odesenfans:
## What

Adds a read-only `GET /api/v0/storage/limits.json` endpoint that returns this node's configured upload size limits (in bytes), grouped by storage engine:

```json
{
  "storage": {
    "max_file_size": 104857600,
    "max_unauthenticated_upload_file_size": 26214400
  },
  "ipfs": {
    "max_upload_file_size": 1073741824,
    "max_upload_car_size": 4294967296,
    "max_unauthenticated_upload_file_size": 26214400
  }
}
```

## Why

Clients currently have no way to discover a node's upload caps short of trial-and-error 413 responses. These limits are per-node configuration and **vary between CCNs** (this was the root cause of a recent prod incident where the effective `/ipfs/add_file` cap differed across nodes), so a client cannot safely hardcode them. With this endpoint a client can pick the right engine for a given file size: anything larger than `storage.max_file_size` must be uploaded through the IPFS engine (`/api/v0/ipfs/add_file`).

Grouping by engine (rather than a flat list) keeps the two independent `max_unauthenticated_upload_file_size` values (one under `storage`, one under `ipfs`) unambiguous.

## Security

The endpoint reflects only an explicit allowlist of size integers, never the broader config object, so it cannot grow into a config-dump. It discloses nothing that the existing aiohttp 413 responses (`Maximum request body size N exceeded`) do not already reveal, and knowing a limit does not raise it.

## Routing note

The literal `limits.json` route is registered before the `/storage/{file_hash}` dynamic route so it is not captured as a hash. A test guards this.

## Tests

- `test_get_storage_limits` — asserts the full response matches the configured defaults.
- `test_get_storage_limits_not_captured_by_hash_route` — guards the route ordering.

Full `tests/api/test_storage.py` + `tests/api/test_ipfs.py` (50 tests) pass; mypy clean.

### DIFF
--- a/src/aleph/web/controllers/routes.py
+++ b/src/aleph/web/controllers/routes.py
@@ -151,6 +151,9 @@ def register_routes(app: web.Application, swagger: Optional[SwaggerDocs]):
             "/api/v0/storage/add_json",
             storage.add_storage_json_controller,
         ),
+        # Register the literal path before the {file_hash} route below, which
+        # would otherwise capture "limits.json" as a hash.
+        web.get("/api/v0/storage/limits.json", storage.get_storage_limits),
         web.get("/api/v0/storage/{file_hash}", storage.get_hash),
         web.get("/api/v0/storage/raw/{file_hash}", storage.get_raw_hash),
         web.get(
--- a/src/aleph/web/controllers/storage.py
+++ b/src/aleph/web/controllers/storage.py
@@ -927,3 +927,78 @@ async def get_file_pins_count(request: web.Request) -> web.Response:
     with session_factory() as session:
         count = count_file_pins(session=session, file_hash=item_hash)
     return web.json_response(data=count)
+
+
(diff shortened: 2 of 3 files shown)


foxpatch-aleph (approved):
A clean, well-scoped endpoint that exposes an explicit allowlist of per-node upload size limits grouped by storage engine. The config field accesses match existing usage patterns elsewhere in the codebase, the Pydantic response models prevent any config-dump growth, and the disclosed values are already observable via 413 responses. Route ordering is correctly guarded by both a comment and a test. No correctness, security, or testing gaps identified.

**`tests/api/test_storage.py`** (line 907): Minor cosmetic nit: the docstring has a stray space after the opening quotes (`""" "limits.json"...`). This is harmless but slightly odd to read.

PR #1212, opened 2026-07-01
fix(ipfs): drop the dead pinning timeout knob

odesenfans:
## Context

While investigating a 1 GiB IPFS upload timing out via the Rust CLI, the suspected cause was the server-side IPFS pin timeout in `/api/v0/ipfs/add_file`. Investigation showed that timeout does not actually exist at runtime.

## The bug

`config.ipfs.pinning.timeout` (default 60s) and the `timeout` parameters on the IPFS client factories are passed to `aioipfs.AsyncIPFS(read_timeout=...)`, but **aioipfs 0.7.1 silently ignores `read_timeout`**: `AsyncIPFS.__init__` stores it in `self._read_timeout`, then builds its client session via `get_session()` called with no arguments, so the value never reaches aiohttp.

The effective timeouts are aioipfs's own session defaults: `total=1800s` (30 min) and `sock_read=600s` (10 min). Those are generous, which is also why the 1 GiB failure surfaced as a client-side abort rather than a server pin timeout.

This made the knob a footgun:
- Bumping the default looked like a fix but was a no-op.
- Wiring it through would newly impose a low cap that does not exist today.

## The fix

Remove the dead `timeout` / `read_timeout` plumbing and the `config.ipfs.pinning.timeout` key, and document that aioipfs's session defaults govern the pin duration. No runtime behavior change.

## Tests

Adds `tests/services/test_ipfs_common.py` asserting the client factories never forward `read_timeout`.

### DIFF
--- a/src/aleph/config.py
+++ b/src/aleph/config.py
@@ -263,8 +263,10 @@ def get_defaults():
                 "port": 5001,
                 # Scheme of the ipfs pinning service.
                 "scheme": "http",
-                # Timeout for pinning operations (seconds)
-                "timeout": 60,
+                # NOTE: there is deliberately no pinning timeout here. The
+                # aioipfs client ignores read_timeout (see make_ipfs_client),
+                # so the pin duration is governed by aioipfs's own session
+                # defaults (30 min total / 10 min inter-read).
             },
             # Timeout for file stat requests (seconds)
             "stat_timeout": 30,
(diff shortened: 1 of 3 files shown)


foxpatch-aleph (approved):
Clean, well-documented removal of a dead configuration knob. The PR correctly removes the `timeout`/`read_timeout` parameter from all three IPFS client factory functions and the corresponding `config.ipfs.pinning.timeout` config key. All call sites have been verified to be compatible with the new signatures. The inline comments thoroughly explain why the parameter was removed (aioipfs silently ignores `read_timeout`) and what the effective timeouts are. The three new tests appropriately assert that `read_timeout` is never forwarded to `AsyncIPFS`. No runtime behavior change, no security concerns.

PR #1213, opened 2026-07-01
fix(ipfs): stream uploads to IPFS instead of buffering in memory

odesenfans:
## Context

Part of investigating large (1 GiB) IPFS uploads via `/api/v0/ipfs/add_file`.

## The bug

The handler buffered the entire upload in memory before pinning:

```python
temp_file = await uploaded_file.open_temp_file()
file_content = await temp_file.read()   # entire 1 GiB into a single bytes object
...
cid = await ipfs_service.add_bytes(file_content)
```

So a 1 GiB upload was held fully in RAM (on top of the temp file already on disk) before being sent to the daemon. Under concurrent large uploads this is a real memory-pressure risk.

## The fix

Add `IpfsService.add_file`, which streams the file from its path via `aioipfs.add` (an async generator that chunks the upload to the daemon), and call it from the handler with the temp file path instead of reading the file into memory.

The stored object and the client-visible hash are unchanged:
- The CID is derived from the file bytes only (we do not wrap in a directory, so the filename is not part of it), so `add(path, cid_version=0)` yields the same CID as `add_bytes(bytes, cid_version=0)`.
- `aioipfs.add` pins by default (`pin=true`), matching the previous behavior.

## Tests

- Unit tests for `add_file` (streams from path, forwards `cid_version=0`, never calls `add_bytes`; raises if the daemon yields no entry).
- Updated `tests/api/test_ipfs.py` and `tests/api/test_storage.py` fixtures: the pin call is now `add_file`, so the "never pinned" assertions target `add_file`.

### DIFF
--- a/src/aleph/services/ipfs/service.py
+++ b/src/aleph/services/ipfs/service.py
@@ -309,6 +309,35 @@ class IpfsService:
         result = await self.storage_client.add_bytes(value, cid_version=cid_version)
         return result["Hash"]
 
+    async def add_file(
+        self, file_path: Union[str, pathlib.Path], cid_version: int = 0
+    ) -> str:
+        """Add a file to IPFS by streaming it from disk.
+
+        Unlike add_bytes, this does not buffer the whole file in memory:
+        aioipfs streams the file from ``file_path`` in chunks. Used by the
+        upload endpoints, where a single file can be up to 1 GiB and reading
+        it fully into RAM would be wasteful (and, at scale, a memory-pressure
+        risk).
+
+        The resulting CID is identical to ``add_bytes`` for the same content:
+        the CID is derived from the file bytes only. The filename is not part
+        of it because we do not wrap the file in a directory.
+        """
+        # ``add`` is an async generator yielding one entry per file added.
+        # We add a single file (no directory wrapping), so exactly one entry
+        # is produced; keep the last one defensively.
+        result: Optional[Dict] = None
+        async for entry in self.storage_client.add(
+            str(file_path), cid_version=cid_version
+        ):
+            result = entry
+
+        if result is None:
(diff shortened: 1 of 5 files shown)


foxpatch-aleph (dismissed):
The PR correctly replaces in-memory buffering with streaming via aioipfs.add, eliminating the memory-pressure risk for large uploads. The CID equivalence claim is accurate (no directory wrapping means the filename doesn't affect the hash), and pinning behavior is preserved since both add and add_bytes rely on the daemon's default pin=true. The temp file cleanup path still works because the finally block calls cleanup() which safely handles the case where open_temp_file() was never called. Unit tests cover both the happy path and the no-entry edge case, and the test fixture fixes are correct — including a latent bug in test_storage.py where the mock returned a dict instead of a string.

**`tests/api/test_ipfs.py`** (line 90): The add_bytes mock is still set up in both fixtures (here and line 1191) but is no longer called by the IPFS add_file handler. This is harmless since other code paths may use it, but if no test in this file exercises add_bytes, it could be removed to reduce fixture noise. Non-blocking.

foxpatch-aleph (approved):
The PR correctly replaces the in-memory buffering of uploaded files with streaming via `aioipfs.add`, which reads the file from disk in chunks. The new `IpfsService.add_file` method is well-documented, forwards `cid_version=0` to match `add_bytes`, and the IPFS daemon pins by default (matching prior behavior). The temp file cleanup path is preserved through the existing `finally` block. Test coverage is thorough: unit tests verify streaming, `cid_version` forwarding, `add_bytes` non-use, and the no-entry edge case; existing handler tests are correctly updated to mock `add_file` instead of `add_bytes`. The `test_storage.py` fixture correctly retains both mocks since `StorageService.add_file` still calls `add_bytes` internally.

PR #1214, opened 2026-07-01
fix(ipfs): build add_file multipart with BufferedReaderPayload

odesenfans:
## Problem

The streaming `add_file` path merged in #1213 crashes in production on every IPFS file upload:

```
File "/opt/pyaleph/src/aleph/web/controllers/ipfs.py", line 197, in ipfs_add_file
  cid = await ipfs_service.add_file(uploaded_file.get_temp_file_path())
File "/opt/pyaleph/src/aleph/services/ipfs/service.py", line 331, in add_file
  async for entry in self.storage_client.add(...)
File ".../aioipfs/api.py", line 1399, in add
  file_payload = payload.BytesIOPayload(...)
File ".../aiohttp/payload.py", line 851, in __init__
  self._size = len(self._value.getbuffer()) - self._value.tell()
AttributeError: '_io.BufferedReader' object has no attribute 'getbuffer'
```

## Root cause

`aioipfs.AsyncIPFS.add()`, for a single file, opens it with `open(path, "rb")` (an `io.BufferedReader`) and wraps it in aiohttp's `BytesIOPayload`. On aiohttp >= 3.13, `BytesIOPayload.__init__` calls `value.getbuffer()`, which only exists on `io.BytesIO`, not on a `BufferedReader`. We ship aiohttp 3.13.5, so the call always raises.

aioipfs 0.7.1 is the latest release and still has this bug (its directory branch correctly uses `BufferedReaderPayload`; only the single-file branch is wrong), so there is no upstream version to pick up.

## Fix

Build the multipart request ourselves with aiohttp's `BufferedReaderPayload` (the correct payload for a file object, which streams it from disk in chunks) and post it through `CoreAPI.add_generic`. This keeps the memory win from #1213 (the file is never buffered in RAM) without the crash.

## Why the tests missed it

The existing tests mocked `storage_client.add`, so the real aioipfs/aiohttp payload construction never ran. The updated tests capture the multipart handed to `add_generic` and assert it carries a streaming `BufferedReaderPayload`, never an in-memory `BytesIOPayload`/`BytesPayload`, so a regression here would now fail in CI.

## Verification

- New/updated unit tests in `tests/services/test_ipfs_service.py` pass (16 passed).
- API upload suites `tests/api/test_ipfs.py` + `tests/api/test_storage.py` pass (50 passed).
- End to end against a real kubo 0.41 daemon: streaming `add_file` yields the **same CID** as `add_bytes` for the same content (`QmWiEuiWvVUE7dH3uswANBT7LRb54Nnu2YZMShaLMzkV7D`) and the content is retrievable via `cat`.

### DIFF
--- a/src/aleph/services/ipfs/service.py
+++ b/src/aleph/services/ipfs/service.py
@@ -333,23 +333,26 @@ class IpfsService:
         ``AttributeError`` on aiohttp >= 3.13. ``BufferedReaderPayload`` is the
         correct payload for a file object and streams it straight from disk.
         """
-        file_path = str(file_path)
-        # add_generic and _build_add_params live on the CoreAPI, which
-        # AsyncIPFS exposes as ``.core`` (only a handful of core methods are
-        # proxied on the client itself, and these two are not among them).
+        path = pathlib.Path(file_path)
+        # ``cid-version`` is a kubo /api/v0/add query parameter (stable across
+        # aioipfs versions), so we build the params dict directly rather than
+        # coupling to aioipfs's private _build_add_params. add_generic and the
+        # FormDataWriter have no public equivalent, so those we do reach for on
+        # the CoreAPI (AsyncIPFS.core): the client proxies only a handful of
+        # core methods and add_generic is not among them.
+        params = {"cid-version": str(cid_version)}
         core = self.storage_client.core
-        params = core._build_add_params(cid_version=cid_version)
 
         result: Optional[Dict] = None
-        with open(file_path, "rb") as fd:
+        with open(path, "rb") as fd:
             with aioipfs.multi.FormDataWriter() as mpwriter:
                 file_payload = aiohttp.payload.BufferedReaderPayload(
                     fd,
                     content_type="application/octet-stream",
-                    headers={"Abspath": file_path},
+                    headers={"Abspath": str(path)},
                 )
                 file_payload.set_content_disposition(
-                    "form-data", name="file", filename=pathlib.Path(file_path).name
+                    "form-data", name="file", filename=path.name
                 )
                 mpwriter.append_payload(file_payload)
 
--- a/src/aleph/services/ipfs/service.py
+++ b/src/aleph/services/ipfs/service.py
@@ -7,7 +7,9 @@ from typing import AsyncIterable, Dict, Optional, Self, Union, cast
 
 import aiofiles
 import aiohttp
+import aiohttp.payload
 import aioipfs
+import aioipfs.multi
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (dismissed):
The fix correctly works around an aioipfs 0.7.1 bug by building the multipart request directly with BufferedReaderPayload (which streams from disk) instead of relying on aioipfs's add() that incorrectly wraps a file in BytesIOPayload. The approach mirrors what aioipfs's own directory branch does. Resource management is sound (file and FormDataWriter are properly scoped with context managers). Tests are well-structured: they capture the actual multipart payload and assert it uses BufferedReaderPayload, closing the gap that let the original bug slip through. The reliance on aioipfs's private _build_add_params is a necessary coupling given no public API exists, and it's well-documented. The synchronous open() in async context is acceptable since only the fd creation is blocking; actual I/O is chunked via aiohttp's payload system.

**`src/aleph/services/ipfs/service.py`** (line 336): Minor nit: `file_path = str(file_path)` is followed by `pathlib.Path(file_path).name` on line 352, which re-parses the string back into a Path. You could keep the original Path object and call `.name` on it, then `str()` only where the string is needed (open, Abspath header). Not a blocking issue.

**`src/aleph/services/ipfs/service.py`** (line 341): Calling `core._build_add_params()` accesses a private aioipfs API. This is well-documented in the comment above and there is no public alternative, but it does create a coupling that could break on an aioipfs upgrade. Consider adding a guard or a fallback if aioipfs ever renames this method.

foxpatch-aleph (approved):
The fix correctly works around an aioipfs 0.7.1 bug by building the IPFS add multipart request manually with BufferedReaderPayload instead of relying on aioipfs.AsyncIPFS.add(), which crashes on aiohttp >= 3.13 due to BytesIOPayload calling getbuffer() on a BufferedReader. The approach mirrors aioipfs's own directory-branch code and preserves the streaming/memory-safety properties from #1213. The cid-version parameter is correctly forwarded as a kubo query param. Tests are substantially improved: they now create a real temp file, capture the multipart payload, and assert on payload types, which would have caught the original regression. No security or correctness issues found.

**`tests/services/test_ipfs_service.py`** (line 322): Accessing `captured["mpart"]._parts` relies on a private attribute of aiohttp's MultipartWriter. This is acceptable for a regression test but will break silently if aiohttp renames or restructures this internal. Consider adding a brief comment noting this dependency on a private API, or alternatively iterating via `mpart._parts` with a fallback assertion message that makes the fragility visible.

PR #1215, opened 2026-07-01
perf(credits): drain expense lots set-based instead of per-address

odesenfans:
## Problem

Applying a credit expense message drained each address's lots in a Python loop: one `SELECT ... FOR UPDATE` plus a `session.flush()` per address (`_consume_address_credits`). A large expense turned into tens of thousands of synchronous DB round-trips, and the ever-growing SQLAlchemy identity map made each per-address flush progressively more expensive. Observed on a production node: a 27,143-address expense ran for 2+ minutes and appeared to stall near the end.

```
[INFO] Updating credit balances expense for 27143 addresses   19:33:59
[INFO] Done updating credit balances                          19:36:08   (2m09s)
```

## Change

Replace the per-address loop with a set-based drain that processes the whole message in **two statements**:

1. `SELECT ... FOR UPDATE` that locks every eligible lot for the batch in a deterministic global order (`address, message_timestamp, credit_ref, credit_index`), preserving the old per-address lock serialization and avoiding deadlocks. It is a separate statement because Postgres forbids `FOR UPDATE` alongside a window function.
2. A single `UPDATE ... FROM` whose window function does the FIFO allocation: `prev_cumulative` is the summed `amount_remaining` of strictly-earlier lots per address, so each lot yields `LEAST(amount_remaining, requested - prev_cumulative)`; fully-covered lots drop out via the `WHERE`, and over-draw is capped because a lot is never decremented below zero.

The history-row build and the bulk `COPY` are unchanged. The transfer path keeps the per-lot `_consume_address_credits` helper (single sender, not the hotspot).

### Semantics preserved
Lot eligibility (still-funded, granted at or before the expense, not yet expired at the expense instant), FIFO order, over-draw dropping, per-entry `-amount` history rows, the `amount <= 0` no-op, and the precision multiplier (still applied in Python before the arrays are built).

## Impact

~55,000+ round-trips (plus the growing identity-map flush cost) collapse to 2 statements + the existing history COPY. The expense should drop from minutes to well under a second, and it no longer holds `FOR UPDATE` locks for the whole duration.

## Tests

- `tests/db/test_credit_balances.py`: 57 passed (54 pre-existing + 3 new).
- `tests/message_processing/test_process_posts.py`: 11 passed.
- The 3 new characterization tests were confirmed green against the old loop first, then against the new implementation: independent per-address FIFO, over-draw isolation, batch-vs-sequential equivalence, and same-address-twice cumulative draining.

## Follow-up (not in this PR)

Worth checking `credit_balances` bloat / autovacuum on the slow node, since table bloat also inflates the per-query latency.

### DIFF
--- a/src/aleph/db/accessors/balances.py
+++ b/src/aleph/db/accessors/balances.py
@@ -18,7 +18,8 @@ from typing import (
 )
 
 from aleph_message.models import Chain, MessageType
-from sqlalchemy import case, func, select, text
+from sqlalchemy import BigInteger, String, bindparam, case, func, select, text
+from sqlalchemy.dialects.postgresql import ARRAY
 from sqlalchemy.dialects.postgresql import insert as pg_insert
 from sqlalchemy.sql import ColumnElement, Select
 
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
Well-executed set-based optimization that correctly replaces a per-address Python loop with two SQL statements. The window-function FIFO logic is sound: `prev_cumulative` via `ROWS BETWEEN UNBOUNDED PRECEDING AND 1 PRECEDING` combined with `LEAST(amount_remaining, requested - prev_cumulative)` and the `WHERE requested > prev_cumulative` filter correctly implements FIFO draining, over-draw capping, and same-address cumulative summing. The separate `FOR UPDATE` lock statement with deterministic global ordering preserves the old per-address serialization while preventing deadlocks. All parameters are properly bound with correct types. The three new tests cover the key semantics (independent FIFO, over-draw isolation, batch-vs-sequential equivalence, same-address-twice draining). The caller interface is unchanged, making this a transparent internal optimization.

foxpatch-aleph (approved):
The set-based FIFO drain is correct and well-reasoned. The window function logic accurately reproduces the per-address FIFO consumption: prev_cumulative (sum of strictly-earlier lots' amount_remaining) combined with LEAST(amount_remaining, requested - prev_cumulative) and the WHERE guard ensures lots are drained in emission order, never go negative, and over-draw is silently dropped. The two-statement approach (lock then update) is necessary because PostgreSQL forbids FOR UPDATE alongside window functions, and the deterministic lock ordering prevents deadlocks. Same-address multiple entries are correctly handled via GROUP BY address + SUM(amount). The amount <= 0 short-circuit, eligibility filters, and per-entry history rows are all preserved. The three new tests cover independent FIFO, over-draw isolation, batch-vs-sequential equivalence, and cumulative same-address draining. The existing test suite already covers expired/backdated lot filtering through the single-address path, which now routes through the batch SQL. This is a clean, high-impact performance optimization.

**`tests/db/test_credit_balances.py`** (line 3110): Minor coverage gap: no batch-specific test verifies that expired lots or lots with message_timestamp > expense_ts are correctly skipped by the set-based drain. The existing single-address tests exercise these filters through the new batch path, but a dedicated batch test with a mix of eligible and expired/backdated lots would make the intent explicit and guard against future regressions in the SQL WHERE clause.

PR #1216, opened 2026-07-01
fix(credit): correct daily cost accounting and removal reason for credit payments

amalcaraz:
Three related fixes around credit-paid message removal, found while debugging an instance that was "removed" on one node and "processed" on another.

1. Credit balance cron daily-cost accounting. Costs in account_costs are stored per-second, but the cron treated them as hourly: it compared the balance against `cost * 24` (3600x too low) and then decremented the running balance by only `cost` (one second) per resource. Use `cost * DAY` (86400) for both the threshold and the per-resource reservation, so the check matches the ingest-time guard in cost_validation.validate_balance_for_payment (sum of per-second costs * DAY) and an account cannot keep more resources than it can fund for a day.

2. Removal reason in the message status endpoint. It hardcoded balance_insufficient for every REMOVING/REMOVED message, even credit-paid ones. Derive the reason from the persisted payment_type column so credit-paid removals report credit_insufficient and hold/superfluid ones report balance_insufficient.

Tests: tests/jobs/test_credit_balance_job.py covers the credit cron removal threshold and the per-resource daily reservation; tests/api/test_get_message.py covers the API removal reason for both payment types.

### DIFF
--- a/src/aleph/jobs/cron/credit_balance_job.py
+++ b/src/aleph/jobs/cron/credit_balance_job.py
@@ -23,7 +23,7 @@ from aleph.db.models.cron_jobs import CronJobDb
 from aleph.db.models.messages import MessageDb, MessageStatusDb
 from aleph.jobs.cron.cron_job import BaseCronJob
 from aleph.services.cost import calculate_storage_size
-from aleph.toolkit.constants import CREDIT_ONLY_CUTOFF_TIMESTAMP, MiB
+from aleph.toolkit.constants import CREDIT_ONLY_CUTOFF_TIMESTAMP, DAY, MiB
 from aleph.toolkit.timestamp import utc_now
 from aleph.types.db_session import DbSession, DbSessionFactory
 from aleph.types.message_status import MessageStatus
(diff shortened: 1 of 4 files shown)


foxpatch-aleph (dismissed):
Both fixes are correct and well-tested. The daily-cost fix changes the multiplier from 24 (hourly) to DAY (86400) for both the removal threshold and the per-resource reservation, correctly matching the ingest-time check in cost_validation.validate_balance_for_payment. The _removal_reason helper correctly derives the reason from the persisted payment_type column, reporting credit_insufficient for credit-paid messages and balance_insufficient for hold/superfluid. Tests cover the underfunded removal, funded retention, per-resource reservation with multiple instances, and both payment-type branches of the API reason. One pre-existing redundancy (duplicate get_message_status call at lines 59 and 78) is present in the changed file but was not introduced by this PR.

**`src/aleph/jobs/cron/credit_balance_job.py`** (line 59): Pre-existing nit: `get_message_status` is called at line 59 and its result is immediately overwritten by a second call at line 78 without being used in between. The first call is a redundant DB query that could be removed.

foxpatch-aleph (approved):
Both fixes are correct and well-tested. The credit balance cron fix correctly changes the daily cost calculation from `cost * 24` (which treated per-second costs as hourly) to `cost * DAY` (86400), matching the ingest-time guard in `validate_balance_for_payment`. The per-resource reservation is similarly corrected to decrement by the full daily cost rather than the per-second cost, preventing an account from keeping more resources than it can fund for a day. The removal reason fix correctly derives the reason from the persisted `payment_type` column instead of hardcoding `BALANCE_INSUFFICIENT`. Tests cover the key scenarios: underfunded removal, adequately funded retention, per-day reservation with multiple instances, and API removal reason for both payment types.

**`src/aleph/jobs/cron/credit_balance_job.py`** (line 62): Pre-existing (not introduced by this PR): `get_message_status(session, item_hash)` is called on line 62 and again on line 81, with the first result overwritten by the second. The line 62 call is dead code and could be removed in a follow-up.

**`tests/api/test_get_message.py`** (line 304): Minor: no test case for superfluid payment type, but the behavior is correct by design — `_removal_reason` falls through to `BALANCE_INSUFFICIENT` for anything other than `credit`, which matches the PR description. Optional to add a superfluid REMOVING/REMOVED fixture for completeness.

PR #1219, opened 2026-07-08
feat(billing): preserve termination metadata and price forgotten/removed STOREs

amalcaraz:
### Changes

- **Forgotten metadata**: `forget_message()` now snapshots `owner`,
  `payment_type` (NULL coalesced to hold), file `size` and `forgotten_at`
  into `forgotten_messages`. Migration 0061 adds the columns, backfills
  `forgotten_at` from still-indexed FORGET messages, and builds the indexes
  concurrently (rerun-safe).
- **Removed metadata**: new `removed_messages` table with a two-phase
  lifecycle — size snapshotted when a message enters REMOVING (the file row
  still exists at that point; the GC deletes it before flipping the status),
  row deleted if the message recovers, `removed_at` stamped by the GC at
  REMOVING→REMOVED. All writes (including the denormalized status
  dual-writes) are gated on the guarded status-flip rowcount to survive
  GC↔recovery races. Migration 0063.
- **Terminal-status queries**: `messages.json?msgStatuses=forgotten|removed`
  (single status only) windows and sorts on `forgotten_at`/`removed_at`,
  supports `paymentTypes` (coalesced to hold for removed) and returns 400 for
  unsupported narrowing filters. Single-message endpoints expose the new
  fields as epoch floats.
- **Pricing**: `/api/v0/price/{hash}` computes live prices for forgotten and
  removed STOREs from the preserved metadata through the estimation path —
  same floor and pricing aggregate as live messages. Missing metadata
  (legacy rows) still returns 410.
- **v2 expenses**: `aleph_credit_expense` entries may carry `count`/`size`,
  persisted to `credit_history` as `expense_count`/`expense_size_mib`
  (migration 0062). v1 messages are unaffected (NULL columns).
- **Spam floor**: `MIN_STORE_COST_MIB` raised 25 → 100 MiB (credit-only).

### Behavior changes worth reviewing

- `/price/{hash}` on a forgotten/removed STORE now returns 200 with a
  computed cost instead of 410 (only when metadata exists).
- Forgotten/removed responses expose `owner` and `size` — required for
  billing, but it narrows what FORGET hides. Deliberate, flagged for review.
- Precomputed `account_costs` keep the old 25 MiB floor until a
  recalculation; estimates and forgotten/removed pricing use 100 MiB
  immediately (documented at the constant).

Migrations: 0061 → 0062 → 0063, all rerun-safe, concurrent index builds.

### DIFF
--- a/deployment/migrations/versions/0063_f4b8a2d7c1e9_removed_messages_table.py
+++ b/deployment/migrations/versions/0063_f4b8a2d7c1e9_removed_messages_table.py
@@ -1,20 +1,33 @@
-"""Add removed_messages billing metadata table
+"""Add removed_messages snapshot table
 
 Revision ID: f4b8a2d7c1e9
 Revises: e4a8c2d6f1b9
 Create Date: 2026-07-06
 
-Two-phase removal record: the balance/credit-balance cron jobs snapshot the
-file size at PROCESSED->REMOVING (the garbage collector deletes the files row
+Snapshot table for removed messages, mirroring forgotten_messages: at
+REMOVING->REMOVED the garbage collector copies the billing metadata of the
+messages row here and deletes the row (confirmations/costs/metrics follow
+via ON DELETE CASCADE).
+
+Two-phase lifecycle: the balance/credit-balance cron jobs snapshot the file
+size at PROCESSED->REMOVING (the garbage collector deletes the files row
 before the status flips to REMOVED, so the size must be captured while the
-message is still alive), recovery deletes the row, and the garbage collector
-stamps removed_at at REMOVING->REMOVED.
+message is still alive), recovery deletes the record, and the garbage
+collector fills the metadata and stamps removed_at at REMOVING->REMOVED.
 
-No backfill: legacy REMOVED rows have an unknown removal time (mirrors legacy
-forgotten rows).
+removed_at is node-local (each node's GC finalizes removals on its own
+schedule) and NOT deterministic across nodes — unlike forgotten_at there is
+no sender-declared removal time to share.
 
-The removed list endpoint windows/sorts on removed_at, so its index is built
-CONCURRENTLY (same pattern as migration 0060) to avoid blocking writes.
+Existing REMOVED messages are moved into the snapshot: their metadata is
+copied (size best-effort from a still-existing files row, NULL otherwise;
+removed_at unknown -> NULL, mirroring legacy forgotten rows) and their
+messages rows are deleted.
+
+The removed list endpoint filters by owner and windows/sorts on removed_at,
+so a composite (owner, removed_at) index and a plain removed_at index are
+built CONCURRENTLY (same pattern as migration 0060) to avoid blocking
+writes.
 """
 
 from alembic import op
@@ -36,9 +49,64 @@ def upgrade() -> None:
             """
         CREATE TABLE IF NOT EXISTS removed_messages (
             item_hash VARCHAR PRIMARY KEY,
+            type VARCHAR,
+            chain VARCHAR,
+            sender VARCHAR,
+            signature VARCHAR,
+            item_type VARCHAR,
+            time TIMESTAMPTZ,
+            channel VARCHAR,
+            owner VARCHAR,
+            payment_type VARCHAR,
             size BIGINT,
             removed_at TIMESTAMPTZ
         )
+        """
+        )
+    )
+
+    # Move existing REMOVED messages into the snapshot. ON CONFLICT keeps
+    # reruns (and rows already snapshotted while REMOVING) intact except for
+    # the metadata, which only the messages row knows; an existing size
+    # snapshot always wins over the files lookup.
+    op.execute(
+        text(
+            """
+        INSERT INTO removed_messages (
+            item_hash, type, chain, sender, signature, item_type, time,
+            channel, owner, payment_type, size, removed_at
+        )
+        SELECT
+            m.item_hash, m.type, m.chain, m.sender, m.signature, m.item_type,
+            m.time, m.channel, m.owner,
+            COALESCE(m.payment_type, 'hold'),
+            f.size,
+            NULL
+        FROM messages m
+        JOIN message_status ms ON ms.item_hash = m.item_hash
+        LEFT JOIN files f ON f.hash = m.content_item_hash
+        WHERE ms.status = 'removed'
+        ON CONFLICT (item_hash) DO UPDATE SET
+            type = EXCLUDED.type,
+            chain = EXCLUDED.chain,
+            sender = EXCLUDED.sender,
+            signature = EXCLUDED.signature,
+            item_type = EXCLUDED.item_type,
+            time = EXCLUDED.time,
+            channel = EXCLUDED.channel,
+            owner = EXCLUDED.owner,
+            payment_type = EXCLUDED.payment_type,
+            size = COALESCE(removed_messages.size, EXCLUDED.size)
+        """
+        )
+    )
+    op.execute(
+        text(
+            """
+        DELETE FROM messages m
+        USING message_status ms
+        WHERE ms.item_hash = m.item_hash
+          AND ms.status = 'removed'
         """
         )
     )
@@ -56,6 +124,13 @@ def upgrade() -> None:
         conn.execute(
             text(
                 "CREATE INDEX CONCURRENTLY IF NOT EXISTS "
+                "ix_removed_messages_owner_removed_at "
+                "ON removed_messages (owner, removed_at)"
+            )
+        )
+        conn.execute(
+            text(
+                "CREATE INDEX CONCURRENTLY IF NOT EXISTS "
                 "ix_removed_messages_removed_at "
                 "ON removed_messages (removed_at)"
             )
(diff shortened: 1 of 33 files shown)


foxpatch-aleph (commented):

foxpatch-aleph (commented):
Well-designed PR with careful race-condition handling, rerun-safe migrations, and comprehensive test coverage. The two-phase removal lifecycle and guarded status flips are correct. One minor naming issue in ForgottenMessage.serialize_time where the parameter 'dt' shadows the 'datetime as dt' import, but it's not a runtime bug. The code is production-ready with this cosmetic nit.

**`src/aleph/schemas/api/messages.py`** (line 224): The parameter name `dt` shadows the module-level `import datetime as dt`. While this works at runtime (annotations are evaluated in the enclosing scope where `dt` is still the module, and `dt.timestamp()` calls the method on the parameter value), it's confusing and could trip up mypy or readers. Consider renaming the parameter to `value` for consistency with `serialize_forgotten_at` below.

deployment/migrations/versions/0063_f4b8a2d7c1e9_removed_messages_table.py
```diff
@@ -0,0 +1,69 @@
...
+    # must not error on the already-created table.
+    op.execute(
+        text(
+            """
+        CREATE TABLE IF NOT EXISTS removed_messages (
+            item_hash VARCHAR PRIMARY KEY,
+            size BIGINT,
+            removed_at TIMESTAMPTZ
+        )
+        """
```
odesenfans, line 64: Are you sure we need this table? Seems like the only purpose is to track `removed_at`, right?

src/aleph/schemas/api/messages.py
```diff
@@ -204,6 +214,21 @@ class ForgottenMessage(BaseModel):
...
+
+    @field_serializer("time")
+    def serialize_time(self, dt: dt.datetime, _info) -> float:
+        return dt.timestamp()
+
+    @field_serializer("forgotten_at")
+    def serialize_forgotten_at(
+        self, value: Optional[dt.datetime], _info
+    ) -> Optional[float]:
+        return value.timestamp() if value is not None else None
```
odesenfans, line 267: Can we not serialize to timestamp though? I used human-readable formats for easier debugging, timestamps are probably a bit faster but I think datetimes are generally better.

src/aleph/toolkit/constants.py
```diff
@@ -336,7 +336,13 @@ class ProductPriceType(str, Enum):
...
 DEFAULT_MAX_UNAUTHENTICATED_UPLOAD_FILE_SIZE = 25 * MiB
 DEFAULT_MAX_UPLOAD_CAR_SIZE = 4 * GiB
-MIN_STORE_COST_MIB = 25  # Minimum MiB cost for pure STORE messages
+# Minimum MiB billed for pure STORE messages, applied to credit payment only
+# (services/cost.py::_calculate_storage_costs). Bumped 25 -> 100 as a
+# small-file spam deterrent. Transitional inconsistency after the bump:
+# precomputed account_costs rows keep the old floor until a cost
+# recalculation runs, while estimates and forgotten/removed pricing use the
+# new floor immediately.
+MIN_STORE_COST_MIB = 100
```
odesenfans: That's a big increment. Wasn't 25 okay-ish?

odesenfans (changes requested):
A few things to discuss:

* The removed_messages table: can we avoid it?
* The 25 -> 100 MiB change
* timestamps vs datetimes in JSON output

odesenfans:
## Security review finding #1 — Inconsistent `payment_type` NULL-coalescing in forgotten query

**Severity:** Medium

The **removed** messages query correctly coalesces NULL payment types:

```python
# make_matching_removed_messages_query (src/aleph/db/accessors/messages.py:541-545)
func.coalesce(MessageDb.payment_type, "hold").in_([pt.value for pt in payment_types])
```

The **forgotten** messages query does **not**:

```python
# make_matching_forgotten_messages_query (src/aleph/db/accessors/messages.py:577-580)
ForgottenMessageDb.payment_type.in_([pt.value for pt in payment_types])
```

Legacy forgotten rows (forgotten before this PR) have `payment_type = NULL`, which per billing semantics means "hold". `forget_message()` does coalesce to "hold" at forget time (line 840), so new rows are fine — but legacy rows are not. Filtering with `paymentTypes=hold` will silently **exclude** these legacy rows, while the removed endpoint correctly includes them.

**Recommendation:** Apply `func.coalesce(ForgottenMessageDb.payment_type, "hold")` to the forgotten query for consistency, or document the behavioral difference if it is intentional.

odesenfans:
## Security review finding #2 — GC may commit status flip without `removed_at` on exception

**Severity:** Low

In `_check_and_update_removing_messages` (`src/aleph/services/storage/garbage_collector.py:96-133`), the status flip and `mark_removed_message()` are queued in the same transaction:

```python
result = session.execute(make_message_status_upsert_query(...))  # line 98-106
if result.rowcount > 0:
    session.execute(sa_update(MessageDb)...)                        # dual-write
    mark_removed_message(session=session, ...)                      # line 124 — can throw
```

If `mark_removed_message()` throws (unlikely for a simple upsert, but possible on a DB constraint or connection issue), the broad `except Exception` at line 128 catches it and the loop continues. The subsequent `session.commit()` at line 137 would then commit the `REMOVING->REMOVED` status flip **without** the `removed_at` stamp on `removed_messages`.

The result: a message in `REMOVED` status with `removed_at = NULL`, treated as a "legacy" removal and excluded from date-windowed queries — a silent data-integrity gap.

**Recommendation:** Either:
- (a) Wrap the status flip + dual-write + `mark_removed_message` in a savepoint (`session.begin_nested()`) so a failure rolls back all three, or
- (b) Stamp `removed_at` *before* the status flip, so the record exists regardless of what happens next.

foxpatch-aleph (commented):

foxpatch-aleph (commented):

foxpatch-aleph (commented):

odesenfans (approved):

foxpatch-aleph (commented):
Well-structured PR with thorough test coverage and sound race-condition handling via guarded status flips and per-message savepoints. The two-phase removed_messages lifecycle is correctly implemented, and the forgotten/removed pricing path through the estimation aggregate is a clean reuse of existing logic. One discrepancy between the PR description (MIN_STORE_COST_MIB raised 25→100) and the actual code (still 25) needs clarification. A few minor observations around error handling completeness and the _forget_item_hash fall-through for REMOVED messages, but no blocking issues.

**`src/aleph/toolkit/constants.py`** (line 339): The PR description states 'MIN_STORE_COST_MIB raised 25 → 100 MiB (credit-only)' but the constant is still 25. Either the change was omitted from this PR or the description is inaccurate. The tests reference MIN_STORE_COST_MIB from the constants so they pass either way, but the documented behavior change does not match the code.

**`src/aleph/handlers/content/forget.py`** (line 202): When message_status is REMOVED, the code logs but does not return — it falls through to the PROCESSED/REMOVING check at line 214, which raises ForgetTargetNotFound. This is pre-existing behavior (not introduced by this PR), but now that REMOVED messages have their rows deleted from messages, the get_message_by_item_hash at line 225 would also fail. Since the status check at line 202 already identifies the case, an early return here would be cleaner and avoid a confusing ForgetTargetNotFound error for a message that was simply already removed.

**`src/aleph/web/controllers/prices.py`** (line 225): The _price_store_from_billing_metadata function catches RuntimeError and raises HTTPNotFound, but the forgotten/removed pricing paths (_price_forgotten_store_message / _price_removed_store_message) catch (ValidationError, ValueError) and raise the 'gone' HTTPGone. If get_total_and_detailed_costs raises a RuntimeError (e.g. pricing aggregate not found), the client gets a 404 rather than a 410 for a message that is known to be forgotten/removed. This is a minor UX inconsistency — a 410 would be more semantically accurate for a terminated message whose pricing infrastructure is temporarily unavailable.

**`src/aleph/web/controllers/messages.py`** (line 1050): When a REMOVED message's snapshot is not found, HTTPGone (410) is raised. This is correct and consistent with the forgotten message path (which raises HTTPNotFound at line 1008 for a missing forgotten_messages row). However, the inconsistency between 410 for removed and 404 for forgotten is pre-existing and not introduced by this PR.

**`deployment/migrations/versions/0063_f4b8a2d7c1e9_removed_messages_table.py`** (line 87): The data-move DELETE at line 87 deletes messages rows with status='removed'. If any FK constraints or triggers reference these rows beyond the known CASCADE on confirmations/costs/metrics, the migration could fail on nodes with custom schema extensions. The comment at line 97 ('ON DELETE CASCADE') documents the known cascades, but a pre-migration check for orphaned FK references might be worth considering for production nodes.

**`src/aleph/db/accessors/messages.py`** (line 734): In remove_message(), the messages row is deleted after the upsert. If the DELETE triggers an ON DELETE CASCADE that somehow affects removed_messages (unlikely given the PK is item_hash, not an FK to messages), the snapshot could be lost. This is safe given the current schema (removed_messages has no FK to messages), but worth noting that the ordering (upsert first, delete second) is the correct one — reversing it would lose the data.

PR #1220, opened 2026-07-09
feat: add V-PROGRAM message type support (phase 1)

odesenfans:
## What

Phase-1 CCN support for the new `V-PROGRAM` message type (SEV-SNP verifiable programs, `VerifiableProgramContent` from aleph-im/aleph-message#158): parse, price, store, and display.

- **Parsing**: `PendingVProgramMessage`, `CONTENT_TYPE_MAP` entry, tag extraction from `content.metadata.tags`.
- **API**: `VProgramMessage` response model wired into `MESSAGE_CLS_DICT`, the `AlephMessage` discriminated union, and the headers content format, so V-PROGRAM rows render in every /messages endpoint (list filters, single message, pending, rejected, websocket).
- **Pricing**: V-Programs are priced on the existing `instance_confidential` tier (compute units + declared extra `volumes`; the runtime bundle and workload are STORE-paid artifacts and deliberately carry no execution-volume cost). Cost-estimation schemas and both price endpoints (`GET /api/v0/price/{item_hash}`, `POST /api/v0/price/estimate`) support the type.
- **Processing**: dedicated `VProgramMessageHandler` validates the credit balance (V-Programs are credit-only by schema) and persists cost rows; recurring credit billing picks the rows up generically. `process`/`forget` keep no side tables in phase 1: the message row is the source of truth.

No DB migration is needed (the message `type` column is a plain varchar).

## Phase-1 scope (deliberate)

- No measurement validation and no runtime/workload ref checks (phase 2 will add CCN-side measurement recomputation per the confidential VM protocol design).
- No `vms` table rows, no `VmType` member: v-programs do not appear in vms accessors yet, and forgetting a STORE file referenced by a v-program workload is not blocked.
- No amend/`replaces` chain validation.

## Merge gate

`pyproject.toml` pins `aleph-message` to the head commit of aleph-im/aleph-message#158. Swap it for the released version once that PR ships in a release, and re-run the type checks, before merging.

## Tests

13 new tests (schema round-trips with the canonical cross-SDK fixture, cost tier selection, end-to-end pipeline processing/rejection against Postgres, price and display endpoints). Full suite: 871 passed, 10 skipped, 0 failed; black/ruff/isort clean; mypy clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1222, opened 2026-07-10
feat(vprogram): give V-Programs a vms representation and forget-protection

odesenfans:
Follow-up to #1220 (V-PROGRAM phase 1).

## Problem

Phase 1 (#1220) deliberately kept `VProgramMessageHandler.process` a no-op: no `vms` rows, no `VmType` member. That left a genuine footgun: `get_vms_dependent_volumes` — the only gate the FORGET handler uses to protect files referenced by executables — could not see V-Program references, so **forgetting a STORE file used by a live V-Program (workload image, runtime manifest, hash trees, verified volumes) was not blocked**, silently breaking the deployment.

## Changes

- **`VmType.VPROGRAM`** — pure Python change: the `vms.type` column is a varchar-backed `ChoiceType`, no DDL needed for the enum member.
- **`VProgramDb(VmBaseDb)`** polymorphic subclass. Following the `ProgramDb` precedent (scalar per-type fields as columns on the single-inheritance table, lists as side tables):
  - runtime manifest + workload as nullable columns on `vms` (`runtime_ref`, `runtime_comment`, `workload_ref`, `workload_hash_tree`, `workload_roothash`);
  - the verified volume list in a new `vprogram_verified_volumes` table with a `position` column — order matters, it's the measured cmdline order (`verified_volumes=h1,h2,...`) — cascading on `vms` deletion like the other volume tables (migration 0064).
  - No `vm_versions` rows: V-Programs are immutable at the schema level (no `allow_amend`/`replaces`), there is no amend chain to track. Cross-type amends are already impossible (`check_dependencies` resolves `replaces` via `get_program`, which won't return a V-Program row).
  - The verification block (SNP policy, launch measurements) stays in the message content: the SNP policy is 64-bit and does not fit the int4 column holding the SEV policy of instances.
- **Handler**: `process` writes the row, `forget_message` deletes it (volumes follow via `ON DELETE CASCADE`).
- **Forget-protection**: `get_vms_dependent_volumes` now also scans `runtime_ref`, `workload_ref`, `workload_hash_tree` (columns on `vms`, no extra join) and `vprogram_verified_volumes.ref`/`.hash_tree`. Hash trees are STORE files too and are as load-bearing as the data images.
- **Bug fix in the same query**: switched `scalar_one_or_none()` to `first()`. Several rows can legitimately match one file hash — two VMs sharing a runtime ref (common: official runtimes), or two volumes of one VM referencing the same file — which previously raised `MultipleResultsFound` instead of blocking the forget. Regression-tested.

## Not in scope

- The async removal path (`removed_messages` GC, #1219) does not consult `get_vms_dependent_volumes` for any executable type; that insolvency-removal ordering question is pre-existing and shared with instances/programs.
- `check_dependencies`-time existence checks for referenced STOREs (the `find_missing_volumes` analog) — candidate follow-up.

## Tests

- `tests/db/test_vprograms_db.py` (new): accessors, all five ref/hash-tree columns block via `get_vms_dependent_volumes` (parametrized), `MultipleResultsFound` regression, cascade deletion.
- `tests/message_processing/test_process_vprograms.py`: processing now asserts the full vms row (identity, refs, positional volumes); new end-to-end lifecycle test — forget of the referenced STORE is **rejected** with `FORGET_NOT_ALLOWED`, forgetting the V-Program deletes its rows, after which the STORE forget succeeds.

Locally: the vprogram/forget/instance/program/db/removal suites pass and mypy is clean (317 files); migrations are exercised by the test harness (alembic upgrade to head). Full-suite run left to CI.

### DIFF
--- a/deployment/migrations/versions/0064_a7c3e9f2d5b1_vprogram_vms_rows.py
+++ b/deployment/migrations/versions/0064_a7c3e9f2d5b1_vprogram_vms_rows.py
@@ -1,0 +1,59 @@
+"""Add the vms representation of V-PROGRAM messages
+
+Revision ID: a7c3e9f2d5b1
+Revises: f4b8a2d7c1e9
+Create Date: 2026-07-10
+
+V-Programs get a row in the single-inheritance vms table (type='vprogram').
+The runtime manifest and workload are single refs, so they live in nullable
+columns on vms (like the program-specific columns); the positional verified
+volume list gets its own table, cascading on vms deletion like the other
+volume tables.
+
+These rows exist so that get_vms_dependent_volumes sees the STORE files a
+V-Program references (runtime manifest, workload image and hash tree,
+verified volumes and their hash trees) and the forget handler blocks
+forgetting them while the V-Program is alive.
+
+No data backfill: no V-PROGRAM message can have reached PROCESSED before
+this migration (the message type ships in the same release).
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "a7c3e9f2d5b1"
+down_revision = "f4b8a2d7c1e9"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column("vms", sa.Column("runtime_ref", sa.String(), nullable=True))
+    op.add_column("vms", sa.Column("runtime_comment", sa.String(), nullable=True))
+    op.add_column("vms", sa.Column("workload_ref", sa.String(), nullable=True))
+    op.add_column("vms", sa.Column("workload_hash_tree", sa.String(), nullable=True))
+    op.add_column("vms", sa.Column("workload_roothash", sa.String(), nullable=True))
+    op.create_table(
+        "vprogram_verified_volumes",
+        sa.Column("vm_hash", sa.String(), nullable=False),
+        sa.Column("position", sa.Integer(), nullable=False),
+        sa.Column("ref", sa.String(), nullable=False),
+        sa.Column("hash_tree", sa.String(), nullable=False),
+        sa.Column("roothash", sa.String(), nullable=False),
+        sa.Column("comment", sa.String(), nullable=False),
+        sa.ForeignKeyConstraint(
+            ["vm_hash"], ["vms.item_hash"], ondelete="CASCADE"
+        ),
+        sa.PrimaryKeyConstraint("vm_hash", "position"),
+    )
+
+
+def downgrade() -> None:
+    op.drop_table("vprogram_verified_volumes")
+    op.drop_column("vms", "workload_roothash")
+    op.drop_column("vms", "workload_hash_tree")
+    op.drop_column("vms", "workload_ref")
+    op.drop_column("vms", "runtime_comment")
+    op.drop_column("vms", "runtime_ref")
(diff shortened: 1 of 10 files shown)


foxpatch-aleph (dismissed):
Well-structured PR that correctly extends the single-inheritance vms table with a VProgramDb polymorphic subtype and a positional verified-volumes side table. The forget-protection query correctly scans all five ref/hash-tree columns, and the scalar_one_or_none() to first() bug fix with .limit(1) is a legitimate fix for the MultipleResultsFound edge case. The migration is clean (valid hex revision ID, correct down_revision chain, proper CASCADE FK). Tests are thorough at both the DB level (parametrized ref coverage, MultipleResultsFound regression, cascade deletion) and the end-to-end level (full forget lifecycle). One minor latent gap: repair.py's mark_processed_message_as_rejected does not handle MessageType.v_program, but this is not a live bug since that function is currently only called for PROGRAM messages.

**`src/aleph/repair.py`** (line 67): Latent gap: `mark_processed_message_as_rejected` only cleans up `vms` rows for `program` and `instance` types. If this function is ever extended to reject V-PROGRAM messages, the `VProgramDb` row and its `vprogram_verified_volumes` would be orphaned (the CASCADE only fires on `delete_vm`, which is not called here for v_program). Not a live bug today since the only caller (line 240) filters by `MessageType.program`, but worth adding `MessageType.v_program` to the tuple for symmetry, or a comment noting the exclusion is intentional.

foxpatch-aleph (dismissed):
Well-structured PR that gives V-PROGRAMs a vms representation to close a forget-protection gap. The implementation faithfully follows existing patterns (ProgramDb for scalar columns, side tables for lists, cascade deletion). The scalar_one_or_none() to first()+limit(1) fix in get_vms_dependent_volumes is correct and necessary for the multi-join query. The repair.py change to include v_program in the rejection cleanup prevents orphaned vms rows. Test coverage is thorough across all new code paths.

foxpatch-aleph (approved):
Well-structured PR that closes a genuine forget-protection gap for V-Programs by adding a vms representation (VProgramDb polymorphic subclass + VProgramVerifiedVolumeDb side table), wiring the handler to write/delete it, and extending get_vms_dependent_volumes to scan the new ref/hash-tree columns. The scalar_one_or_none() → scalars().first() fix with .limit(1) correctly handles the MultipleResultsFound regression. The repair path was correctly extended to clean up v-program vms rows on rejection. Migration revision IDs are valid hex, down_revision chains correctly to 0063, and the no-backfill rationale is sound. Test coverage is thorough: accessor tests, parametrized dependent-volume tests for all five ref columns, a multiple-match regression test, cascade deletion, an end-to-end forget lifecycle test, and a repair rejection test. No correctness, security, or performance issues found.

**`src/aleph/handlers/content/vprogram.py`** (line 80): The str() casts on ref/hash_tree are inconsistent with vm.py (line 211) which passes content.runtime.ref directly without casting. Harmless since ItemHash is a str subclass, but the inconsistency is slightly surprising. Consider dropping the casts for consistency, or adding a brief comment explaining why they're here.

**`src/aleph/handlers/content/vprogram.py`** (line 84): workload_roothash is not str()-cast while workload_ref and workload_hash_tree are. If roothash is a plain string in the message model (not an ItemHash) this is correct, but the inconsistency is worth a one-line comment for clarity.

**`src/aleph/db/models/vms.py`** (line 243): VProgramVerifiedVolumeDb.comment is nullable=False, so if the V-Program message schema ever allows comment=None for verified volumes, the insert would fail with an IntegrityError. This is fine if the schema mandates a non-None comment (which the test content suggests), but worth confirming against the aleph_message schema definition.

PR #1225, opened 2026-07-20
feat(cli): add --repair/--no-repair switch to toggle boot repair

odesenfans:
## Summary

Adds a `--repair` / `--no-repair` CLI switch controlling whether repair operations run at node startup. `--repair` is the default, so existing behaviour is unchanged; operators can pass `--no-repair` to skip repair on boot.

## Changes

- **`src/aleph/cli/args.py`** — new `--repair` argument using `argparse.BooleanOptionalAction`, which creates the paired `--repair` / `--no-repair` flags and defaults to `True`.
- **`src/aleph/commands.py`** — the `repair_node(...)` startup call is now gated behind `if args.repair:` (logging when disabled), mirroring the existing `--no-jobs` pattern.
- **`tests/cli/test_args.py`** — new tests covering the default (`True`), `--repair` (`True`), and `--no-repair` (`False`).

## Testing

- `hatch run testing:test tests/cli/test_args.py` — 3 passed (written test-first).
- `hatch run linting:fmt` / `linting:all` — ruff, black, isort, yamlfix pass. The mypy step fails on a pre-existing SQLAlchemy-stubs plugin conflict (aborts at plugin-load before any file is checked); reproduces identically on the untouched base.

### DIFF
--- a/src/aleph/cli/args.py
+++ b/src/aleph/cli/args.py
@@ -30,6 +30,14 @@ def parse_args(args):
     )
     parser.add_argument("--no-jobs", action="store_true", dest="no_jobs", default=False)
     parser.add_argument(
+        "--repair",
+        dest="repair",
+        help="Run repair operations at startup (enabled by default). "
+        "Use --no-repair to skip them.",
+        action=argparse.BooleanOptionalAction,
+        default=True,
+    )
+    parser.add_argument(
         "-v",
         "--verbose",
         dest="loglevel",
--- a/src/aleph/commands.py
+++ b/src/aleph/commands.py
@@ -192,9 +192,12 @@ async def main(args: List[str]) -> None:
         )
         await stack.enter_async_context(chain_connector)
 
-        await repair_node(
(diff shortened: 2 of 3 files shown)


foxpatch-aleph (approved):
Small, well-scoped PR that adds a --repair/--no-repair CLI switch using argparse.BooleanOptionalAction, defaulting to True to preserve existing behavior. The gate in commands.py correctly mirrors the existing --no-jobs pattern, logging when repair is disabled. Tests cover the three meaningful argument cases. No correctness, security, or quality concerns; the only minor gap is the absence of a test exercising the commands.main gate itself, but that is non-blocking given the trivial if/else logic.

**`tests/cli/test_args.py`** (line 17): Tests cover parse_args output thoroughly. A non-blocking suggestion: consider adding a test (in tests/ for commands.py) verifying that repair_node is called when args.repair is True and skipped when False, to cover the behavior gate in commands.py:195-200. Not required given the trivial if/else logic.

PR #1233, opened 2026-08-03
fix(pricing): backfill missing credit price from default in pricing timeline

aliel:
A pricing aggregate update that replaced a nested price object without a `credit` field, the on-chain shape during the 2025-02 → 2025-10 window, before credit pricing existed, resolved to cost_credit=0 for every item priced in that window, because get_pricing_timeline built each model from the shallow-merged on-chain elements only, with no fallback to the default.

Deep-merge each on-chain product type onto DEFAULT_PRICE_AGGREGATE so unset leaves (e.g. INSTANCE.price.compute_unit.credit = 14250) fall back to the default instead of zeroing. On-chain values still win where present, and product types never priced on chain stay absent (timeline semantics unchanged). The shallow aggregate-merge semantics themselves are untouched.

Adds a regression test asserting a gap-window INSTANCE compute_unit with no credit inherits the default 14250 while its on-chain payg/holding win.

### DIFF
--- a/src/aleph/services/pricing_utils.py
+++ b/src/aleph/services/pricing_utils.py
@@ -20,6 +20,29 @@ from aleph.types.cost import ProductPricing
 from aleph.types.db_session import DbSession
 
 
+def _deep_merge(base: dict, override: dict) -> dict:
+    """Recursively merge ``override`` onto a copy of ``base``.
+
+    Nested dicts are merged key-by-key so a *partial* override inherits any
+    leaf it does not set from ``base`` instead of dropping it. Non-dict values
+    (and dict-vs-non-dict conflicts) are taken from ``override``.
+
+    Used to backfill the pricing model from ``DEFAULT_PRICE_AGGREGATE``: a
+    pricing aggregate update that replaced e.g. ``INSTANCE.price.compute_unit``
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
The fix is correct and well-targeted. _deep_merge recursively backfills missing leaves from DEFAULT_PRICE_AGGREGATE onto each on-chain product type's merged data, so a partial update that replaced e.g. INSTANCE.price.compute_unit without a credit field now inherits the default credit (14250) instead of resolving to cost_credit=0. On-chain values still win where present, product types never priced on chain stay absent (timeline semantics preserved), and the shallow aggregate-merge semantics are untouched. The regression test covers the core scenario. The _deep_merge implementation is clean and handles dict-vs-non-dict conflicts correctly by taking the override. No security concerns.

**`tests/services/test_pricing_utils.py`** (line 379): The test asserts compute_unit.credit is backfilled but doesn't verify storage.credit. The gap-window element's storage dict also lacks a credit field, so it too gets backfilled from the default (0.17967489030626108). Adding an assertion for instance.price.storage.credit would make the test more complete, since this is the same class of bug affecting a different leaf.

PR #1234, opened 2026-08-04
fix(vprogram): check store dependencies of V-PROGRAM messages at processing time

odesenfans:
## Problem

V-PROGRAM messages had no dependency existence check: `VProgramMessageHandler` did not override `check_dependencies`, and `find_missing_volumes` only branches on `ProgramContent` / `InstanceContent`. A V-Program referencing store files unknown to the node was processed successfully and then parked invisibly at the scheduler, whose hard lookup was the only guard.

This is the pyaleph phase-2 work item recorded in the #1222 review notes (the forget-protection half already landed there via `get_vms_dependent_volumes`).

## Fix

Add `check_dependencies` to `VProgramMessageHandler`. It checks the file pins for the five store references of the content:

- `runtime.ref` (runtime manifest)
- `workload.ref` and `workload.hash_tree`
- each verified volume's `ref` and `hash_tree`

V-Program references are direct store item hashes (there is no `use_latest` / file tag mechanism for measured content), so the pin check is sufficient. Missing refs raise `VmVolumeNotFound`, giving the same retry-then-reject behavior as programs with missing volumes (`VM_VOLUME_NOT_FOUND`).

Out of scope: the balance-removal/GC path not consulting `get_vms_dependent_volumes` for any executable type is a separate, pre-existing gap.

## Tests

- New `test_process_vprogram_missing_refs`: a V-Program with no seeded refs is rejected with `VM_VOLUME_NOT_FOUND` and the error details list all five missing refs (written first, watched fail against main).
- Existing processing and API tests now seed the referenced file pins (mirroring `insert_volume_refs` in the program tests) so they still exercise their original paths (processing, cost persistence, credit rejection, forget blocking).
- `tests/message_processing/`, `tests/api/test_vprogram_api.py`, `tests/db/test_vprograms_db.py`, `tests/services/test_cost_service_vprogram.py`, `tests/test_repair.py`, `tests/jobs/test_check_removing_messages.py`: all passing.

### DIFF
--- a/src/aleph/handlers/content/vprogram.py
+++ b/src/aleph/handlers/content/vprogram.py
@@ -3,6 +3,7 @@ from typing import List, Set
 
 from aleph_message.models import PaymentType, VerifiableProgramContent
 
+from aleph.db.accessors.files import find_file_pins
 from aleph.db.accessors.vms import delete_vm
 from aleph.db.models import MessageDb, VProgramDb, VProgramVerifiedVolumeDb
 from aleph.db.models.account_costs import AccountCostsDb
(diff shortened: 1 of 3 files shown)


foxpatch-aleph (approved):
The PR correctly adds a `check_dependencies` override to `VProgramMessageHandler` that validates all store references (runtime, workload, hash trees, verified volumes) exist as file pins before processing, mirroring the existing pattern in `vm.py`. The implementation is straightforward, uses the existing `find_file_pins` accessor and `VmVolumeNotFound` exception, and is consistent with the codebase conventions. All tests exercising the processing pipeline are updated to seed the required file pins, and a new test validates the rejection path when refs are missing. No correctness, security, or quality issues found.

PR #1236, opened 2026-08-06
fix(credits): clean up expense ledger rows referencing forgotten messages

odesenfans:
Adds migration 0065: removes credit_history expense rows whose source messages have been forgotten (constrained to the 2026-08-05 window), keeps them in a backup audit table, and rebuilds the credit_balances lot cache for the affected addresses.

Idempotent (safe on re-run, on already-repaired nodes, and on fresh syncs). Runs automatically at startup; no operator action required beyond deploying the release. Full technical background is in the migration docstring.

### DIFF
--- a/deployment/migrations/versions/0065_b3d7f1a9c4e2_remove_double_charged_credit_expenses.py
+++ b/deployment/migrations/versions/0065_b3d7f1a9c4e2_remove_double_charged_credit_expenses.py
@@ -147,7 +147,7 @@ def upgrade() -> None:
     #    manual repair SQL already ran and created the table.
     conn.execute(sa.text(f"""
             CREATE TABLE IF NOT EXISTS {BACKUP_TABLE}
-                (LIKE credit_history INCLUDING DEFAULTS)
+                (LIKE credit_history INCLUDING DEFAULTS INCLUDING INDEXES)
             """))
     conn.execute(sa.text(f"""
             INSERT INTO {BACKUP_TABLE}
@@ -166,7 +166,7 @@ def upgrade() -> None:
         sa.text(f"DELETE FROM credit_history ch WHERE {BAD_ROWS_PREDICATE}")
     ).rowcount
     logger.info(
-        "Removed %d double-charged credit_history row(s) " "(backup in %s)",
+        "Removed %d double-charged credit_history row(s) (backup in %s)",
         deleted,
         BACKUP_TABLE,
     )
--- a/deployment/migrations/versions/0065_b3d7f1a9c4e2_remove_double_charged_credit_expenses.py
+++ b/deployment/migrations/versions/0065_b3d7f1a9c4e2_remove_double_charged_credit_expenses.py
@@ -1,0 +1,196 @@
+"""Remove the 2026-08-05 double-charged credit expenses
+
+Revision ID: b3d7f1a9c4e2
+Revises: a7c3e9f2d5b1
+Create Date: 2026-08-06
+
+On 2026-08-05 17:00-17:32 UTC an expense-repricing run republished ~53 days
+of already-billed hourly aleph_credit_expense posts as NEW posts (type
+aleph_credit_expense + ref) instead of amends (type "amend"). The credit
+ledger is additive per message, so every address in those posts was charged
+the full window a second time. The bad posts were later FORGOTTEN, but
+forgetting does not reverse credit_history, and the corrected "amend" run
+never touches balances (amend is not a credit_balances post type).
+
+A ledger row is bad iff its credit_ref (the source message hash) now points
+to a forgotten_messages entry: legitimate hourly expense messages are still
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (commented):
Well-structured, idempotent migration that correctly identifies and removes double-charged credit_history rows from the 2026-08-05 incident. The frozen copy of the lot-rebuild logic faithfully matches the application code, the SQL is valid PostgreSQL, and the idempotency design handles all three scenarios (re-run, manual repair, fresh sync). The downgrade is sound. No correctness or security bugs found. The main gap is the absence of test coverage — the codebase has a precedent for testing migration SQL (migrations 0061, 0063), and given the financial sensitivity of credit ledger manipulation, tests for the delete predicate, backup population, lot rebuild, and idempotency would be expected. A minor efficiency note: the backup table is created without a primary key or indexes, so the anti-join and DISTINCT queries will seq-scan it.

**`deployment/migrations/versions/0065_b3d7f1a9c4e2_remove_double_charged_credit_expenses.py`** (line 149): The backup table is created with `LIKE credit_history INCLUDING DEFAULTS`, which copies columns and defaults but not the primary key or indexes. Consider adding `INCLUDING INDEXES` or manually creating a primary key on `(credit_ref, credit_index)` to avoid sequential scans on the anti-join (line 159) and `SELECT DISTINCT address` (line 129) when the number of backed-up rows is large.

**`deployment/migrations/versions/0065_b3d7f1a9c4e2_remove_double_charged_credit_expenses.py`** (line 142): No tests exist for this migration. The codebase has a precedent of testing migration SQL (see `tests/db/test_messages.py` lines 664, 716, 742 for migrations 0061 and 0063). Given the financial sensitivity of credit ledger manipulation, consider adding tests that verify: (1) the BAD_ROWS_PREDICATE selects only the intended rows, (2) the backup table is populated correctly, (3) the lot rebuild produces correct balances after deletion, and (4) re-running the upgrade is a no-op.

**`deployment/migrations/versions/0065_b3d7f1a9c4e2_remove_double_charged_credit_expenses.py`** (line 169): The adjacent string literals `"...row(s) " "(backup in %s)"` are concatenated at parse time. This works but the split across two literals is unnecessary — a single string would be clearer.

foxpatch-aleph (commented):
A well-structured data-repair migration that correctly identifies and removes double-charged credit_history rows, backs them up for audit/downgrade, and rebuilds the credit_balances lot cache. The frozen copy of _rebuild_credit_lots_for_address is a good defensive choice to prevent drift. The idempotency design (CREATE IF NOT EXISTS, anti-join backup, replay-based rebuild) is sound. The core logic is correct and the SQL is safe from injection. The main observations are non-blocking: the backup table is missing CONSTRAINTS (NOT NULL, PK) from the LIKE clause, there are no tests for this migration, and the downgrade uses SELECT * which is slightly fragile given the id column.

**`deployment/migrations/versions/0065_b3d7f1a9c4e2_remove_double_charged_credit_expenses.py`** (line 150): Consider adding `INCLUDING CONSTRAINTS` to the LIKE clause: `CREATE TABLE IF NOT EXISTS {BACKUP_TABLE} (LIKE credit_history INCLUDING DEFAULTS INCLUDING INDEXES INCLUDING CONSTRAINTS)`. Without it, the backup table is missing NOT NULL constraints and the primary key on (credit_ref, credit_index). While the copied data won't violate these constraints in practice, an audit table should enforce the same integrity as the source. The PK would also provide a DB-level deduplication guarantee complementing the NOT EXISTS anti-join.

**`deployment/migrations/versions/0065_b3d7f1a9c4e2_remove_double_charged_credit_expenses.py`** (line 191): Using `SELECT *` for the downgrade INSERT is slightly fragile: credit_history has an autoincrement `id` column (not part of the PK) that gets copied from the backup table. This works today because there's no unique constraint on `id`, but a future migration adding one would break this downgrade silently. Consider using an explicit column list that excludes `id` (letting the sequence assign new values), or at minimum document that `id` values are intentionally reused.

**`deployment/migrations/versions/0065_b3d7f1a9c4e2_remove_double_charged_credit_expenses.py`** (line 63): No tests exist for this migration. Given the complexity (backup table creation, anti-join idempotency, delete predicate, lot rebuild, downgrade restore), test coverage would be valuable — especially for the idempotency paths (re-run, manually-repaired node, fresh node) and the downgrade round-trip. The existing test suite in tests/db/test_credit_balances.py has extensive lot-rebuild tests that could be leveraged as a pattern.

**`deployment/migrations/versions/0065_b3d7f1a9c4e2_remove_double_charged_credit_expenses.py`** (line 126): Minor: `_rebuild_backup_table_addresses` gets addresses from the backup table, not from the actual DELETE. On a manually-repaired node (bad rows already deleted, backup table empty), no addresses are rebuilt. This is fine because the `_repair_credit_balances` startup hook in repair.py rebuilds all addresses on every startup, but a brief comment noting this dependency would help future readers understand why the no-address case is safe.

PR #1237, opened 2026-08-06
chore(ipfs): enable pubsub via `Pubsub.Enabled` instead of the deprecated flag

RezaRahemtola:
Kubo deprecated `--enable-pubsub-experiment` in v0.40.0 ([ipfs/kubo#11110](https://github.com/ipfs/kubo/pull/11110)), so since our v0.43.0 bump every daemon start logs:

```
ERROR cmd/ipfs kubo/daemon.go:426 The --enable-pubsub-experiment flag is deprecated. Use Pubsub.Enabled config option instead.
```

`001-update-ipfs-config.sh` now sets `Pubsub.Enabled`, and the four compose files drop the flag. `Pubsub.Enabled` is not a new key — the flag was always an alias for it. No removal version has been announced, so this is cleanup, not a deadline.

Tested against `ipfs/kubo:v0.43.0`: full container start with the script mounted at `/container-init.d/` and the new command line — config applies, daemon ready, pubsub publish/subscribe round-trips, no deprecation warning. Flag and config set together also works, so an operator who keeps the flag is not broken.

## Upgrade Guide (for the release notes)

> The Kubo pubsub flag is replaced by the `Pubsub.Enabled` config option. Do these in order:
>
> 1. Update `001-update-ipfs-config.sh` from this release, then `docker compose up -d ipfs`.
> 2. **Verify it applied:** `docker compose exec ipfs ipfs config Pubsub.Enabled` must print `true`.
>    If it errors, the config did not apply — `ipfs config` is rejected while a repo migration is pending, which is the case on the boot that upgrades the repo. Restart the container once more and re-check. **Do not continue until this returns `true`.**
> 3. Only then remove `--enable-pubsub-experiment` from the `ipfs` service command in `docker-compose.yml` and `docker compose up -d ipfs`.
>
> Leaving the flag in place is harmless — it logs a deprecation warning and is ignored once the config option is set. A CCN with neither the flag nor the config option has pubsub disabled and will not propagate messages, with no error surfaced.


foxpatch-aleph (dismissed):
Clean, correct migration from the deprecated --enable-pubsub-experiment flag to the Pubsub.Enabled config option. All four compose files with an IPFS service are updated consistently, the fifth (test) has no IPFS service. The init script addition follows the existing pattern (ipfs config with --json for booleans) and includes a clear comment. The upgrade guide in the PR description is thorough and covers the repo-migration edge case. No remaining references to the old flag exist anywhere in the repo.

**`deployment/scripts/001-update-ipfs-config.sh`** (line 17): Pre-existing pattern, not introduced by this PR: the script has no `set -e`, so if `ipfs config Pubsub.Enabled` fails (e.g. during a repo migration as noted in the upgrade guide), the script still exits 0 and prints "IPFS config updated!". The upgrade guide warns operators to verify with `ipfs config Pubsub.Enabled`, which is the right mitigation. Consider adding `set -e` in a future PR so any failed config call is surfaced automatically.

foxpatch-aleph (approved):
The PR correctly replaces the deprecated `--enable-pubsub-experiment` CLI flag with the `Pubsub.Enabled` config option across all four docker-compose files and the init script. The `--json 'true'` usage ensures a boolean value is set rather than a string. The additional repo migration logic for stripping legacy Provider/Reprovider keys is sound: the version guard (kubo >= 0.38) and repo-version check (18) correctly target the right migration, and the downgrade-to-17-then-upgrade-to-18 approach is a valid way to re-run the stripping migration. Script hardening with `set -eu` and quoting `$CONFIG_FILE` are good improvements. The `set -e` interaction with commands inside `if` conditions (line 30) is safe — POSIX shells exempt conditional contexts from `set -e`. No correctness, security, or logic issues found.

**`deployment/scripts/001-update-ipfs-config.sh`** (line 21): If `ipfs version --number` outputs something unexpected (e.g., a non-numeric minor version in a custom build), the `cut -d. -f2` result would cause `[ "$kubo_minor" -ge 38 ]` to error and, with `set -e`, abort the script. This is probably the desired behavior (fail loudly rather than silently skip migration), but worth noting that it means a malformed version string would prevent the container from starting.

**`deployment/scripts/001-update-ipfs-config.sh`** (line 44): The upgrade guide warns that `ipfs config` is rejected while a repo migration is pending. Since the repo migration (lines 26 or 32-33) runs before this line, the migration should be complete by this point. However, if the migration fails and `set -e` doesn't catch it (e.g., `ipfs repo migrate` exits 0 but leaves the repo in a partially-migrated state), this `ipfs config` call could fail. This is an edge case that the upgrade guide's "restart once more" advice covers for operators.

PR #1241, opened 2026-08-14
fix(credits): anchor time-based accounting on observed message time

odesenfans:
## Summary

Hardens time-sensitive economic paths against untrusted, sender-supplied message timestamps.

Time-based decisions (credit accounting, storage and payment tier cutoffs, and pricing timeline selection) previously keyed off the message-supplied timestamp. That value is not a trustworthy input for accounting decisions. This introduces a single trusted anchor, `MessageDb.observed_time`, and routes all such decisions through it:

- On-chain confirmation time when available (deterministic across nodes)
- Node reception time otherwise

Credit balance repair re-derives the same observed time from the referenced message, so replay converges deterministically across nodes once messages confirm, and falls back to the stored timestamp when the message row is absent.

No schema migration required.

## Changes

- New `MessageDb.observed_time` property and `observed_time_expr()` SQL helper
- Credit distribution / expense / transfer accounting use observed time
- Storage and payment tier cutoff checks use observed time
- Pricing timeline selection during cost recalculation uses observed time
- Credit repair replay re-derives observed time deterministically

## Tests

Adds regression coverage for the accounting, cutoff, and repair paths. Existing credit, store, program, instance, cost, and repair suites pass.

### DIFF
--- a/deployment/migrations/versions/0066_c1e8b4a9d3f2_snapshot_observed_time.py
+++ b/deployment/migrations/versions/0066_c1e8b4a9d3f2_snapshot_observed_time.py
@@ -1,0 +1,49 @@
+"""Preserve observed time on forgotten/removed message snapshots
+
+Revision ID: c1e8b4a9d3f2
+Revises: b3d7f1a9c4e2
+Create Date: 2026-08-14
+
+Adds an ``observed_time`` column to ``forgotten_messages`` and
+``removed_messages``. The pricing endpoints for forgotten/removed STORE
+messages select the pricing model by message time; using the sender-supplied
+``time`` there lets a backdated message be priced against an older model. The
+messages row is deleted when a message is forgotten/removed, so the trusted
+observed time (on-chain confirmation, else node reception) cannot be recovered
+by joining back later and must be snapshotted at forget/remove time.
+
+Legacy rows created before this column existed keep ``observed_time`` NULL;
+their source messages rows are already gone, so there is nothing to backfill
+from, and the pricing paths fall back to the stored ``time`` for those rows.
+"""
+
+from alembic import op
+from sqlalchemy import text
+
+# revision identifiers, used by Alembic.
+revision = "c1e8b4a9d3f2"
+down_revision = "b3d7f1a9c4e2"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.execute(
+        text(
+            """
+        ALTER TABLE forgotten_messages ADD COLUMN IF NOT EXISTS observed_time TIMESTAMPTZ;
+        ALTER TABLE removed_messages ADD COLUMN IF NOT EXISTS observed_time TIMESTAMPTZ;
+        """
+        )
+    )
+
+
+def downgrade() -> None:
+    op.execute(
+        text(
+            """
+        ALTER TABLE removed_messages DROP COLUMN IF EXISTS observed_time;
+        ALTER TABLE forgotten_messages DROP COLUMN IF EXISTS observed_time;
+        """
+        )
+    )
--- a/src/aleph/db/accessors/messages.py
+++ b/src/aleph/db/accessors/messages.py
@@ -696,6 +696,7 @@ def remove_message(session: DbSession, item_hash: str, removed_at: dt.datetime) 
             "channel",
             "owner",
             "payment_type",
+            "observed_time",
             "removed_at",
         ],
         select(
@@ -711,6 +712,9 @@ def remove_message(session: DbSession, item_hash: str, removed_at: dt.datetime) 
             # Messages without an explicit payment field are hold-paid;
             # capture the effective payment type so billing can rely on it.
             func.coalesce(MessageDb.payment_type, "hold"),
+            # Trusted time for pricing; the messages row is deleted below so it
+            # cannot be recovered by joining back later.
+            MessageDb.observed_time_expr(),
             literal(removed_at),
         ).where(MessageDb.item_hash == item_hash),
     )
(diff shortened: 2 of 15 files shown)


foxpatch-aleph (dismissed):
Well-designed security hardening that correctly routes time-sensitive economic decisions (credit accounting, cost cutoffs, pricing selection, repair replay) through a trusted observed time anchored on on-chain confirmation or node reception time, closing the sender-supplied timestamp forgery vector. The Python property and SQL expression are semantically equivalent, the repair path correctly falls back to the stored history timestamp when the message row is absent, and test coverage addresses the core attack vectors (backdating to mint, precision inflation, confirmation time preference, repair re-anchoring, and orphan fallback). Two remaining uses of sender-supplied time in forgotten/removed store pricing (prices.py:278,316) are the same class of vulnerability but require a schema change to fix and are appropriate for a follow-up; the forget cursor use of sender time appears intentional by design.

**`src/aleph/web/controllers/prices.py`** (line 278): `forgotten_message.time.timestamp()` is still used for pricing-model selection in `_price_forgotten_store_message`, the same class of vulnerability this PR fixes in `recalculate_message_costs` (line 617). The `forgotten_messages` snapshot table doesn't store `first_confirmed_at`/`reception_time`, so fixing this requires persisting observed time on the snapshot or joining back to `messages` at snapshot time. Worth tracking as a follow-up issue.

**`src/aleph/web/controllers/prices.py`** (line 316): Same as line 278: `removed_message.time.timestamp()` in `_price_removed_store_message` still uses sender-supplied time for pricing-model selection. The `removed_messages` snapshot table has the same limitation.

**`src/aleph/handlers/content/forget.py`** (line 176): `forgotten_at=forgotten_by.time` uses the sender-supplied FORGET message time for the deletion-time cursor. This is billing-adjacent, but the `ForgottenMessageDb` docstring explicitly acknowledges this is sender-declared by design to match the live list's time sort. Confirm this is intentionally out of scope for this PR.

**`src/aleph/toolkit/costs.py`** (line 31): The `date` variable is computed unconditionally but only used in the `else` branch (when `height is None`). When `height is not None`, block height is used exclusively and `date` is dead. Pre-existing structure, but the `observed_time` computation could be moved into the `else` branch for clarity. Minor nit.

odesenfans:
Addressed the review points in 69969713:

- **prices.py forgotten/removed STORE pricing:** now select the pricing model from the trusted observed time. Since the `messages` row is deleted at forget/remove, the observed time is snapshotted onto `forgotten_messages` / `removed_messages` at that point (new nullable column + migration 0066). Legacy rows without the column fall back to the stored time.
- **costs.py:** moved the observed-time read into the branch that uses it.
- **forget.py `forgotten_at`:** confirmed intentional. It is the sender-declared deletion time used only for list windowing/cursor, matching the live message list's declared-time semantics, and is not used for any pricing or credit decision.

Added coverage that the snapshots capture the observed time; affected suites pass.

foxpatch-aleph (changes requested):
The core implementation is correct and well-documented: observed_time properly anchors credit accounting, cutoffs, pricing, and repair replay on the trusted confirmation/reception time instead of the forgeable sender-supplied time. The migration is valid and the snapshot logic for forgotten/removed messages is sound. However, the test_pricing_recalculation.py fixture was not updated: the sample_messages fixture does not set first_confirmed_at or reception_time on MessageDb (both default to None/utc_now), so the ordering change from MessageDb.time.asc() to observed_time_expr().asc() makes test_chronological_processing_order non-deterministic because all three messages now sort by approximately-equal reception_time instead of deliberately staggered time values.

**`tests/api/test_pricing_recalculation.py`** (line 115): The sample_messages fixture creates MessageDb instances without setting first_confirmed_at or explicitly setting reception_time on MessageDb (only on MessageStatusDb at line 115). MessageDb.__init__ defaults reception_time to utc_now(), so all three messages get approximately the same reception_time. The PR changes the recalculation ordering from MessageDb.time.asc() to MessageDb.observed_time_expr().asc() = COALESCE(first_confirmed_at, reception_time). Since first_confirmed_at is None and reception_time is ~utc_now() for all three, the ordering in test_chronological_processing_order becomes non-deterministic. Fix: set reception_time=msg.time (or first_confirmed_at=msg.time) on each MessageDb in the fixture so observed_time matches the staggered time values the test expects.

odesenfans:
Fixed in 87e916d3. The `sample_messages` fixture in `test_pricing_recalculation.py` now staggers `reception_time` to match the message time, so `observed_time` reflects the intended chronological order and `test_chronological_processing_order` is deterministic (verified stable across repeated runs). Good catch.

foxpatch-aleph (approved):
Well-executed security hardening that correctly replaces forgeable sender-supplied timestamps with a trusted observed-time anchor (on-chain confirmation or node reception) across all economic paths: credit accounting, pricing model selection, storage/payment tier cutoffs, and credit balance repair. The `observed_time` property and its SQL counterpart are consistent, the repair replay correctly re-derives observed time with a safe fallback for absent message rows, the migration is idempotent and well-documented, and the test coverage directly verifies the attack scenarios (backdating to mint credits, backdating to amplify precision, confirmation-time anchoring for historical backfill). No remaining untrusted timestamp usage was found in economic paths. The only minor gap is the absence of dedicated regression tests for the cutoff functions with a divergent `observed_time` vs `message.time`, but the logic is simple and correct.

**`src/aleph/toolkit/costs.py`** (line 36): Consider adding a dedicated test where `observed_time` is set to a post-cutoff value while `message.time` is pre-cutoff, verifying the cutoff check uses `observed_time` and rejects the free/legacy tier. The existing store tests all set `observed_time = message.time`, so this specific attack vector isn't directly exercised for the cutoff functions.

**`src/aleph/web/controllers/prices.py`** (line 302): The guard checks `removed_message.time is None` but not `removed_message.observed_time`. If a legacy row had `observed_time IS NULL` and `time IS NOT NULL`, it would pass the guard and fall back to the untrusted `time` — which is the intended behavior for legacy rows. However, the guard also rejects rows where `time IS NULL` even if `observed_time IS NOT NULL`, which is slightly over-conservative. Not a bug, but worth noting for future cleanup.

PR #1244, opened 2026-08-17
fix(cron): chunk + yield + cap balance/credit removal to stop event-loop stalls

aliel:
## Problem
`credit_balance_job` and `balance_job` mark every unaffordable message for an account in **one synchronous, non-yielding loop inside a single transaction**. In production, account `0x1424…` had **14,545** unaffordable instances flagged in one tick:

- ~14,545 sequential DB round-trips **without yielding** blocked the main-process event loop → the API and other crons starved (the `balance` cron's bookkeeping session sat `idle in transaction` for over an hour).
- One long transaction held `message_counts` trigger counter-row locks the whole time, which the message-processing subprocess contends on.
- The node's DB is on an HDD shared with IPFS, so each round-trip pays a seek + WAL fsync — turning "slow" into "wedged for an hour".

`EXPLAIN` confirmed the per-message lookup is already an index scan; the issue is the unbounded, non-yielding, single-transaction batch, not a missing index.

## Change
For both `delete_messages` / `recover_messages`:
- **Commit in chunks** (`COMMIT_CHUNK = 500`) so the transaction and its counter-row locks stay small.
- **`await asyncio.sleep(0)`** each chunk (and during the read-only scan) so the API and other crons keep running.
- **`MAX_MESSAGES_PER_RUN = 5000`** cap so one tick can't process an unbounded batch; the remainder is handled next run. Safe because the flips are idempotent (`WHERE status='processed'` / `WHERE status='removing'`).
- Drop the duplicate `get_message_status` in the credit loop; lower the per-message log to `debug`.

The `PROCESSED→REMOVING` / `REMOVING→PROCESSED` flip semantics, STORE small-file exception, removal-record snapshot, and guarded dual-write are unchanged.

## Test
`test_credit_job_caps_messages_per_run`: with the cap shrunk to 2 and chunk to 1, 5 unaffordable instances flip 2 per run and drain over successive runs — proving the cap and idempotent resumption. `balance_job` receives the identical change (the `hold`-payment twin).

## Follow-ups (discussed, not in this PR)
- Move the DB onto SSD, off the IPFS spindle — the root I/O amplifier.
- Larger scale: set-based bulk updates, a dedicated reconciliation worker subprocess, and `message_counts` hot-row contention.

## Known trade-offs
- **Budget bounds attempted work, not successful flips.** `delete_messages`/`recover_messages` return `len(messages)` so skipped no-ops still count against the per-run budget. This is intentional: the cap exists to bound per-tick work (DB round-trips / event-loop time), and counting only flips would let a skip-heavy account iterate unboundedly. Documented in the method docstrings.
- **Deletes are processed before recovers within an account, sharing the budget.** If deletes exhaust the budget, that account's recovers are deferred to the next run (safe via idempotent re-scan). With `MAX_MESSAGES_PER_RUN = 5000` this is unlikely to matter, but an account with consistently many deletes could defer its recovers across ticks.
- **Balance-decrement semantics unchanged:** the running-balance decrement stays unconditional (a cost row with no message-status row still reserves balance), matching the pre-PR behavior.

## last_run resumption fix
The account queries (`get_updated_balance_accounts` / `get_updated_credit_balance_accounts`) are delta-based (`last_update >= last_run`), and the cron framework advanced `last_run` unconditionally after each run. That stranded the remainder of a capped account: once `last_run` moved to `now`, an account with no new balance activity fell out of the query window and its deferred messages were never picked up.

Fix: `BaseCronJob.run` now returns a completion signal — `False` means work was deferred (cap hit). `CronJob.__run_job` advances `last_run` only when the job did **not** return `False`, so the next tick re-queries the same accounts and drains the remainder. The balance/credit jobs return `not deferred`, where `deferred` is set on a cap-break or a truncated `to_delete`/`to_recover` slice. `ORDER BY address` keeps the drain order deterministic across ticks.

Covered by: return-signal assertions in the cap tests, a resumption test that mimics the framework's conditional `last_run` advance (proving a deferred account is not stranded), and a multi-account test (first account exhausts the budget, second is deferred then drained).

### DIFF
--- a/src/aleph/db/accessors/balances.py
+++ b/src/aleph/db/accessors/balances.py
@@ -236,10 +236,13 @@ def update_balances(
 
 
 def get_updated_balance_accounts(session: DbSession, last_update: dt.datetime):
+    # Ordered so that a per-run cap in the balance cron defers accounts
+    # deterministically (fair resumption) rather than by scan order.
     select_stmt = (
         select(AlephBalanceDb.address)
         .where(AlephBalanceDb.last_update >= last_update)
         .distinct()
+        .order_by(AlephBalanceDb.address)
     )
     return (session.execute(select_stmt)).scalars().all()
 
@@ -564,10 +567,13 @@ def get_updated_credit_balance_accounts(session: DbSession, last_update: dt.date
     """
     Get addresses that have had their credit history updated since the given timestamp.
     """
+    # Ordered so that a per-run cap in the credit cron defers accounts
+    # deterministically (fair resumption) rather than by scan order.
     select_stmt = (
         select(AlephCreditHistoryDb.address)
         .where(AlephCreditHistoryDb.last_update >= last_update)
         .distinct()
+        .order_by(AlephCreditHistoryDb.address)
     )
     return session.execute(select_stmt).scalars().all()
 
--- a/src/aleph/jobs/cron/balance_job.py
+++ b/src/aleph/jobs/cron/balance_job.py
@@ -110,7 +110,7 @@ class BalanceCronJob(BaseCronJob):
                     if index % COMMIT_CHUNK == 0:
                         await asyncio.sleep(0)
 
-                if to_delete:
+                if budget > 0 and to_delete:
                     LOGGER.info(
                         f"'{len(to_delete)}' messages to delete for account '{address}'..."
                     )
--- a/src/aleph/jobs/cron/balance_job.py
+++ b/src/aleph/jobs/cron/balance_job.py
@@ -116,10 +116,12 @@ class BalanceCronJob(BaseCronJob):
                         await asyncio.sleep(0)
 
                 if to_delete:
+                    processing = min(len(to_delete), budget)
                     if len(to_delete) > budget:
                         deferred = True
                     LOGGER.info(
-                        f"'{len(to_delete)}' messages to delete for account '{address}'..."
+                        f"Deleting '{processing}' of '{len(to_delete)}' messages "
+                        f"for account '{address}'..."
                     )
                     budget -= await self.delete_messages(session, to_delete[:budget])
 
@@ -128,8 +130,10 @@ class BalanceCronJob(BaseCronJob):
                     if budget <= 0 or len(to_recover) > budget:
                         deferred = True
                     if budget > 0:
+                        processing = min(len(to_recover), budget)
                         LOGGER.info(
-                            f"'{len(to_recover)}' messages to recover for account '{address}'..."
+                            f"Recovering '{processing}' of '{len(to_recover)}' messages "
+                            f"for account '{address}'..."
                         )
                         budget -= await self.recover_messages(
                             session, to_recover[:budget]
--- a/src/aleph/jobs/cron/balance_job.py
+++ b/src/aleph/jobs/cron/balance_job.py
@@ -127,8 +127,12 @@ class BalanceCronJob(BaseCronJob):
     async def delete_messages(
         self, session: DbSession, messages: List[ItemHash]
     ) -> int:
-        """Return the number of messages attempted (including skipped no-ops),
-        used by the caller to decrement the per-run budget."""
+        """Return the number of messages attempted, used by the caller to
+        decrement the per-run budget. Skipped no-ops (missing message,
+        small-file exception, or a status guard that prevents the flip) count
+        on purpose: the budget bounds the number of delete/recover attempts —
+        each a DB round-trip — not only the flips that succeed, because
+        bounding per-tick work is what the cap exists to do."""
         for index, item_hash in enumerate(messages, start=1):
             message = get_message_by_item_hash(session, item_hash)
 
@@ -195,8 +199,12 @@ class BalanceCronJob(BaseCronJob):
     async def recover_messages(
         self, session: DbSession, messages: List[ItemHash]
     ) -> int:
-        """Return the number of messages attempted (including skipped no-ops),
-        used by the caller to decrement the per-run budget."""
+        """Return the number of messages attempted, used by the caller to
+        decrement the per-run budget. Skipped no-ops (missing message,
+        small-file exception, or a status guard that prevents the flip) count
+        on purpose: the budget bounds the number of delete/recover attempts —
+        each a DB round-trip — not only the flips that succeed, because
+        bounding per-tick work is what the cap exists to do."""
         for index, item_hash in enumerate(messages, start=1):
             message = get_message_by_item_hash(session, item_hash)
             if message is None:
--- a/src/aleph/jobs/cron/balance_job.py
+++ b/src/aleph/jobs/cron/balance_job.py
@@ -127,6 +127,8 @@ class BalanceCronJob(BaseCronJob):
     async def delete_messages(
         self, session: DbSession, messages: List[ItemHash]
     ) -> int:
+        """Return the number of messages attempted (including skipped no-ops),
+        used by the caller to decrement the per-run budget."""
         for index, item_hash in enumerate(messages, start=1):
             message = get_message_by_item_hash(session, item_hash)
 
@@ -193,6 +195,8 @@ class BalanceCronJob(BaseCronJob):
     async def recover_messages(
         self, session: DbSession, messages: List[ItemHash]
     ) -> int:
+        """Return the number of messages attempted (including skipped no-ops),
+        used by the caller to decrement the per-run budget."""
         for index, item_hash in enumerate(messages, start=1):
             message = get_message_by_item_hash(session, item_hash)
             if message is None:
--- a/src/aleph/jobs/cron/balance_job.py
+++ b/src/aleph/jobs/cron/balance_job.py
@@ -48,21 +48,26 @@ class BalanceCronJob(BaseCronJob):
         self.session_factory = session_factory
         self.max_unauthenticated_upload_file_size = max_unauthenticated_upload_file_size
 
-    async def run(self, now: dt.datetime, job: CronJobDb):
+    async def run(self, now: dt.datetime, job: CronJobDb) -> bool:
         with self.session_factory() as session:
             accounts = get_updated_balance_accounts(session, job.last_run)
 
             LOGGER.info(f"Checking '{len(accounts)}' updated account balances...")
 
             # Budget the mutations performed this run so one tick cannot process
-            # an unbounded batch; the remainder is idempotently handled next run.
+            # an unbounded batch. `deferred` tracks whether any work was left
+            # undone; when True we return False so the caller does not advance
+            # last_run and the next tick re-queries the same accounts to drain
+            # the remainder (the account query is delta-based on last_run).
             budget = MAX_MESSAGES_PER_RUN
+            deferred = False
 
             for address in accounts:
                 if budget <= 0:
+                    deferred = True
                     LOGGER.info(
-                        "Per-run message cap (%d) reached; remaining accounts "
-                        "will be handled on the next run.",
+                        "Per-run message cap (%d) reached; deferring remaining "
+                        "accounts to the next run.",
                         MAX_MESSAGES_PER_RUN,
                     )
                     break
@@ -110,19 +115,29 @@ class BalanceCronJob(BaseCronJob):
                     if index % COMMIT_CHUNK == 0:
                         await asyncio.sleep(0)
 
-                if budget > 0 and to_delete:
+                if to_delete:
+                    if len(to_delete) > budget:
+                        deferred = True
                     LOGGER.info(
                         f"'{len(to_delete)}' messages to delete for account '{address}'..."
                     )
                     budget -= await self.delete_messages(session, to_delete[:budget])
 
-                if budget > 0 and to_recover:
-                    LOGGER.info(
-                        f"'{len(to_recover)}' messages to recover for account '{address}'..."
-                    )
-                    budget -= await self.recover_messages(session, to_recover[:budget])
+                if to_recover:
+                    # budget may be 0 here if deletes consumed it this account.
+                    if budget <= 0 or len(to_recover) > budget:
+                        deferred = True
+                    if budget > 0:
+                        LOGGER.info(
+                            f"'{len(to_recover)}' messages to recover for account '{address}'..."
+                        )
+                        budget -= await self.recover_messages(
+                            session, to_recover[:budget]
+                        )
 
                 session.commit()
+
+            return not deferred
 
     async def delete_messages(
         self, session: DbSession, messages: List[ItemHash]
--- a/src/aleph/jobs/cron/balance_job.py
+++ b/src/aleph/jobs/cron/balance_job.py
@@ -1,3 +1,4 @@
+import asyncio
 import datetime as dt
 import logging
 from typing import List, cast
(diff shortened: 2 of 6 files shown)


foxpatch-aleph (dismissed):
The PR correctly implements chunked commits, event-loop yielding, and a per-run cap to prevent event-loop stalls from unbounded single-tick processing. The idempotent status-flip guards (WHERE status=PROCESSED / WHERE status=REMOVING) make the chunked-commit and budget-resumption semantics safe: a partial run simply resumes on the next tick. The test validates the cap and chunked-commit path by shrinking both constants and asserting progressive draining across runs. Logic is correct, no off-by-one or race conditions introduced. Minor observations: (1) `balance_job` has no equivalent test despite receiving the identical change; (2) `delete_messages`/`recover_messages` return `len(messages)` (the passed-in count) rather than the count of actually-performed mutations, which slightly under-uses the budget when messages are skipped internally — this is conservative and safe but means the cap is more aggressive than intended; (3) the duplicated `COMMIT_CHUNK`/`MAX_MESSAGES_PER_RUN` constants could be shared, but this is a style nit.

**`src/aleph/jobs/cron/credit_balance_job.py`** (line 128): Consider returning the count of actually-flipped messages (e.g. increment a counter on `result.rowcount > 0`) rather than `len(messages)`. Currently, skipped messages (message is None, small-file exception, or status guard prevents the flip) still count against the budget, so the run processes fewer real mutations than the cap allows. This is conservative and safe (idempotent resumption picks up the rest), but it means a large account with many skip-able messages could exhaust the budget without doing much real work, slightly starving other accounts.

**`src/aleph/jobs/cron/balance_job.py`** (line 117): Same observation as credit_balance_job.py: `delete_messages` returns `len(messages)` rather than the count of actual flips. See comment on credit_balance_job.py line 128.

**`tests/jobs/test_credit_balance_job.py`** (line 455): Good test for the credit job. Since `balance_job` receives the identical chunking/cap change, consider adding a parallel test (or a parametrized variant) for `balance_job` to guard against divergence if the two implementations drift in the future.

foxpatch-aleph (commented):
The PR effectively addresses event-loop stalls by chunking commits, yielding to the event loop, and capping per-run mutations. The budget accounting via `to_delete[:budget]` slicing and `return len(messages)` is correct — budget can reach 0 but never go negative, and the `if budget <= 0: break` guard at the top of the account loop prevents starting new accounts with no budget. The idempotent resumption claim holds because the PROCESSED→REMOVING and REMOVING→PROCESSED flips are guarded by status checks. The duplicate `get_message_status` removal in credit_balance_job and the log-level downgrade from info to debug are both reasonable. The test correctly validates the cap and idempotent resumption with a shrunk cap/chunk. Two non-blocking observations: (1) the identical change was applied to balance_job.py but no equivalent cap test exists for it, and (2) the account-list queries (`get_updated_balance_accounts` / `get_updated_credit_balance_accounts`) have no `ORDER BY`, so when the budget is exhausted the set of accounts processed next run is non-deterministic — the operations are idempotent so this isn't a correctness issue, but it could lead to unfairness where the same accounts are repeatedly deferred.

**`tests/jobs/test_balance_job.py`** (line 1): The PR description states "balance_job receives the identical change" but there is no test for the cap/chunk behavior on `BalanceCronJob`. Consider adding an equivalent `test_balance_job_caps_messages_per_run` test to ensure the twin implementation doesn't regress.

**`src/aleph/jobs/cron/credit_balance_job.py`** (line 128): Minor observation: `delete_messages` and `recover_messages` return `len(messages)` (the input list length), not the count of messages actually mutated. This is fine for budgeting — it bounds attempted work, which is slightly more conservative — but the method name and return type could mislead a future caller into thinking the return value reflects successful mutations. A one-line docstring like `"""Return the number of messages attempted (including skipped no-ops)."""` would clarify intent.

**`src/aleph/jobs/cron/balance_job.py`** (line 61): When the per-run budget is exhausted, the `break` exits the account loop. However, `get_updated_balance_accounts` (balances.py:238) has no `ORDER BY`, so the set of accounts processed each run is non-deterministic. The operations are idempotent so this isn't a correctness bug, but it means the same accounts could be repeatedly deferred while others are always processed. Adding `.order_by(AlephBalanceDb.address)` to the query would make the cap behavior fair and predictable.

foxpatch-aleph (dismissed):
The PR correctly implements chunked commits, event-loop yielding, and a per-run cap to prevent event-loop stalls in the balance and credit crons. The budget accounting is sound — slicing with `to_delete[:budget]` and returning `len(messages)` ensures budget can never go negative. The deterministic `ORDER BY address` on the accounts queries ensures fair resumption across runs. The test coverage for the cap mechanism is solid, verifying idempotent draining over successive runs for both jobs. One subtle behavioral change in the credit job (moving `get_message_status` before the `remaining_credits` decrement means orphaned cost records no longer reduce the running balance) is arguably more correct and unlikely to matter in practice. The recovery-starvation scenario (deletes exhausting budget before recovers for the same account) is a safe, idempotent deferral. No security issues found.

**`src/aleph/jobs/cron/credit_balance_job.py`** (line 88): Behavioral change: moving `get_message_status` + `if status is None: continue` before the `remaining_credits` decrement means that cost records without a matching message status row no longer reduce the running balance. The original code decremented unconditionally. This is arguably more correct (an orphaned cost record shouldn't reserve balance), but the change is not mentioned in the PR description. In practice the impact is negligible since orphaned cost records should not exist, but it would be good to note this explicitly.

**`src/aleph/jobs/cron/balance_job.py`** (line 113): Minor: `to_delete` is not guarded by `if budget > 0` like `to_recover` is on line 119. When budget is 0, `to_delete[:0]` is an empty list, so `delete_messages` is called with `[]` — a harmless no-op but slightly wasteful. Consider adding `if budget > 0 and to_delete:` for symmetry, or document that the empty-list call is intentional.

**`tests/jobs/test_balance_job.py`** (line 459): The cap test covers the delete/removal path well. Consider adding a companion test for the recovery cap path (messages in REMOVING status that need recovery, where the cap limits how many are recovered per run). The code path is symmetric but currently untested for the recover side of the cap.

**`src/aleph/jobs/cron/credit_balance_job.py`** (line 128): If `delete_messages` exhausts the budget for an account, `to_recover` for that same account is skipped (`budget > 0` guard on line 130). This is safe due to idempotency — the account will be re-scanned next run — but an account with consistently many deletes could theoretically starve its recovers. With `MAX_MESSAGES_PER_RUN = 5000` this is unlikely in practice; worth noting as a known trade-off in the PR description.

foxpatch-aleph (dismissed):
Well-designed fix for a real production event-loop stall: chunked commits, event-loop yields during the read-only scan, and a per-run mutation cap with idempotent resumption. The budget semantics (counting attempts not successes) are correct and documented. The duplicate get_message_status call is removed, log levels are appropriately downgraded, and deterministic ordering is added to the accessor queries. Tests exercise the cap and chunked-commit paths with idempotent resumption over multiple runs. The remaining nits (unbounded to_delete/to_recover list building during scan, no test for the delete-then-recover budget interaction) are non-blocking.

**`src/aleph/jobs/cron/balance_job.py`** (line 113): Minor nit: when budget reaches 0, `to_delete[:0]` returns an empty list, so `delete_messages` is called with `[]` (returns 0, no-op). Consider guarding with `if to_delete and budget > 0:` to avoid the wasted call and misleading log message ("N messages to delete" followed by doing nothing). Same pattern in credit_balance_job.py line 124.

**`src/aleph/jobs/cron/balance_job.py`** (line 72): Observation: the scan loop builds the full `to_delete`/`to_recover` lists in memory before the budget slice is applied at line 117. For an account with 14,545 unaffordable messages this still materializes a large list. It's not a bug (hash strings are small) and the scan itself now yields, but if memory is ever a concern, the lists could be truncated as they're built once they exceed `budget`.

**`tests/jobs/test_balance_job.py`** (line 458): Minor coverage gap: the cap test only exercises `to_delete` (all messages unaffordable). There's no test verifying that when `to_delete` partially consumes the budget, the remaining budget correctly limits `to_recover` — e.g., an account with 3 messages to delete and 3 to recover, cap=4, should delete 3 and recover 1. Same gap in test_credit_balance_job.py.

**`src/aleph/jobs/cron/balance_job.py`** (line 62): Observation: the `budget <= 0` check is at the top of the per-account loop, but the read-only scan for the current account (get_total_costs + per-message get_message_status) still runs in full before the budget is checked against mutations. This means if budget=1, an account with 10,000 messages still does 10,000 DB round-trips for the scan (yielding every 500). The PR description acknowledges the budget bounds mutations not scans, so this is an accepted trade-off.

foxpatch-aleph (dismissed):
The PR correctly implements chunked commits, event-loop yielding, and a per-run mutation cap for both balance_job and credit_balance_job. The budget decrement logic is sound: deletes and recovers share a budget, skipped no-ops count intentionally to bound DB round-trips, and the idempotent status guards (WHERE status='processed' / WHERE status='removing') make deferral safe. The removal of the duplicate get_message_status call in credit_balance_job is a genuine improvement. The deterministic ordering of get_updated_balance_accounts / get_updated_credit_balance_accounts ensures fair resumption across ticks. The balance decrement remains unconditional (before the status None-check), preserving original semantics. Tests cover the delete-path cap for both jobs and the recover-path cap for balance_job. The code is well-documented with clear comments explaining non-obvious design decisions.

**`tests/jobs/test_credit_balance_job.py`** (line 455): Test coverage is asymmetric: balance_job has both a delete-cap test and a recover-cap test, but credit_balance_job only has the delete-cap test. Adding a recover-path cap test (twin of test_balance_job_caps_recovery_per_run) would ensure the chunked-commit + budget logic in credit_balance_job.recover_messages is exercised under the cap.

**`src/aleph/jobs/cron/balance_job.py`** (line 79): The read-only scan loop calls get_message_status per item_hash (a DB round-trip each) but does not count these against the per-run budget. An account with 50,000 cost rows where only 100 are unaffordable would do 50,000 round-trips in the scan without hitting the budget cap. The asyncio.sleep(0) yield every COMMIT_CHUNK mitigates event-loop blocking, but the budget doesn't bound this read work. This is acknowledged in the PR description as a known trade-off (budget bounds mutations, not reads), so it's acceptable — just worth keeping in mind for the follow-up set-based bulk update work.

foxpatch-aleph (changes requested):
The chunked-commit, yield, and per-run cap approach is well-designed and addresses the event-loop stall and long-transaction problems effectively. The idempotent status guards make partial processing safe. However, there is a correctness issue: when the per-run budget is exhausted, the cron job's last_run is still advanced by the caller (update_cron_job in cron_job.py:39), so accounts with more than MAX_MESSAGES_PER_RUN unaffordable messages will not have their remaining messages picked up in subsequent runs unless they have new balance activity. The tests don't catch this because they reuse the same cron_job object without advancing last_run between runs. Additionally, the scan loop iterates all cost entries for an account even when the remaining budget is smaller, doing unnecessary get_message_status round-trips.

**`src/aleph/jobs/cron/balance_job.py`** (line 68): Correctness concern: when the budget is exhausted and the loop breaks, the caller (cron_job.py:39) still advances `last_run` to `now` via `update_cron_job`. On the next cron tick, `get_updated_balance_accounts(session, job.last_run)` queries `last_update >= now`, so accounts that were skipped due to budget exhaustion won't reappear unless they have new balance updates after `now`. The PR description says "the remainder is idempotently handled by the next run" but this only holds if the accounts are re-queried. Consider either (a) not advancing `last_run` when the budget was exhausted, (b) re-querying skipped accounts, or (c) documenting this limitation explicitly. The tests at lines 516-523 don't catch this because they reuse the same `cron_job` object without updating `last_run` between runs.

**`src/aleph/jobs/cron/credit_balance_job.py`** (line 73): Same last_run advancement issue as balance_job.py: when the budget is exhausted, `last_run` is still advanced by the caller, so remaining accounts won't be re-queried on the next tick unless they have new credit history updates.

**`src/aleph/jobs/cron/balance_job.py`** (line 110): Performance note: the scan loop iterates ALL hold_costs for an account, calling get_message_status per entry, even if the remaining budget is smaller than len(hold_costs). For an account with 14,545 cost entries and only 1000 budget remaining, all 14,545 entries are scanned. The yields (asyncio.sleep(0)) prevent event-loop starvation, but the unnecessary DB round-trips for unprocessable messages could be avoided by breaking the scan early once to_delete + to_recover exceeds the remaining budget (though this would change which messages get processed, so it may be acceptable as-is).

**`tests/jobs/test_balance_job.py`** (line 516): The test reuses the same `cron_job` object across multiple `balance_job.run()` calls without advancing `last_run`, so it doesn't test the production scenario where `update_cron_job` advances `last_run` after each run. In production, the second run would query `get_updated_balance_accounts(session, now)` which may not return the same account. Consider adding a test that simulates `last_run` advancement between runs to verify the cap + resumption behavior under realistic conditions.

**`tests/jobs/test_credit_balance_job.py`** (line 495): Same test coverage gap as test_balance_job.py: the test doesn't advance `last_run` between successive `run()` calls, so it doesn't catch the case where skipped accounts disappear from the query on the next run.

**`tests/jobs/test_balance_job.py`** (line 459): Consider adding a multi-account test where the first account exhausts the budget and the second account is deferred to the next run. This is the primary production scenario described in the PR (many accounts, one tick can't process all) but no test covers it.

foxpatch-aleph (dismissed):
Well-designed fix for a real production event-loop stall. The budget/deferred/cap logic is correct across all edge cases I traced, the idempotent re-scan ensures deferred work drains over successive ticks, and the conditional last_run advancement in the cron framework is properly implemented with no regression for MetricsPartitionCronJob (returns None, which is not False). Test coverage is solid for the cap/delete/recover/resumption paths. Two minor non-blocking observations: log messages report the full to_delete/to_recover count rather than the budget-truncated count, and the balance job lacks the last_run resumption and multi-account deferral test twins that the credit job has (logic is identical).

**`src/aleph/jobs/cron/balance_job.py`** (line 122): Log says `len(to_delete)` messages to delete, but only `to_delete[:budget]` are actually processed when the cap truncates the list. Consider logging `min(len(to_delete), budget)` instead, or logging the total with a note about the cap. Same issue at line 132 for `to_recover`.

**`src/aleph/jobs/cron/credit_balance_job.py`** (line 136): Same log accuracy issue as balance_job: `len(to_delete)` overstates the actual work when budget truncates. Same at line 146 for `to_recover`.

**`tests/jobs/test_balance_job.py`** (line 458): Consider adding the last_run resumption and multi-account deferral test twins (mirroring test_credit_job_deferred_account_survives_last_run_advance and test_credit_job_defers_second_account_when_budget_exhausted) for completeness, since the balance job has the identical deferral logic.

foxpatch-aleph (approved):
Well-designed fix for event-loop blocking in balance/credit cron jobs. Chunked commits, per-run cap with idempotent resumption, and the conditional last_run advance are all correctly implemented and thoroughly tested. The backward-compatible BaseCronJob return type change is clean. No correctness, security, or race condition issues found.

**`src/aleph/jobs/cron/balance_job.py`** (line 126): When `budget` reaches exactly 0 after this line, `to_delete[:0]` evaluates to `[]` and `delete_messages` returns 0, so `budget -= 0` is a no-op. This path is actually unreachable because the `if budget <= 0: break` guard at line 66 ensures budget > 0 when entering the account loop, and `to_delete` is non-empty when we reach this line. No bug — just noting the invariant holds.

**`src/aleph/jobs/cron/credit_balance_job.py`** (line 108): Good removal of the duplicate `get_message_status` call that existed in the original code. The status is now fetched once, after the balance decrement, which is correct since the decrement must happen regardless of status (to preserve running-balance semantics for cost rows without message-status rows).

**`src/aleph/jobs/cron/cron_job.py`** (line 48): The `completed is not False` check correctly handles all three return values: `False` (deferred, don't advance), `True` (complete, advance), and `None` (MetricsPartitionCronJob, advance). This is a clean backward-compatible contract.

PR #1245, opened 2026-08-20
chore(deps): adopt measurement registers from aleph-message

odesenfans:
Adopts the register-map schema from aleph-im/aleph-message#159.

`LaunchMeasurement.digest` becomes a `registers` object, so a
multi-register TEE such as Intel TDX (MRTD plus RTMRs) fits later without a
breaking protocol change. SEV-SNP declares `{"launch": "<hex>"}`.

## Scope

No CCN code changes. V-PROGRAM measurement validation is not implemented
here yet, so pyaleph only sees the shape through the schema. This is the
dependency bump plus the test fixture.

`VPROGRAM_ITEM_HASH` is recomputed because `item_content` is generated from
the content dict. The new value matches the aleph-rs and aleph-vm fixtures
byte for byte, which keeps the existing cross-SDK canonical-serialization
check honest: if those three ever diverge, something is serializing
differently.

## Merge gate

The pin points at aleph-im/aleph-message#159, which must merge and be
reachable on `od/vprogram-schema` before this installs.

## Test plan

- `tests/messages/test_vprogram.py`: 4 passed

Design: aleph-vm `docs/superpowers/specs/2026-08-20-intel-tdx-design.md` §3, §4.1

https://claude.ai/code/session_01NaLi3sSGWh8EtWcVn64yni

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ dependencies = [
   "aiohttp-swagger3>=0.10.0",
   "aioipfs~=0.7.1",
   "alembic==1.18.5",
-  "aleph-message @ git+https://github.com/aleph-im/aleph-message.git@bc9c0f86958a746faeecbd6c6f96cc6f5b4538de",
+  "aleph-message @ git+https://github.com/aleph-im/aleph-message.git@4ad261ff25f595a64589d49bf36f514c402e7c5e",
   "aleph-nuls2==0.1.1",
   "aleph-p2p-client @ git+https://github.com/aleph-im/p2p-service-client-python@1e992aaacdb0071a34ffb130cbbba9123509e08b",
   "asyncpg==0.31.0",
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
This PR is a minimal, correct dependency bump plus test fixture update. The aleph-message git pin moves to a commit that introduces the registers-map schema for LaunchMeasurement, and the test fixture in test_vprogram.py is updated accordingly (digest -> registers.launch). The recomputed VPROGRAM_ITEM_HASH was independently verified to match the SHA-256 of the canonical JSON serialization of the updated content dict. No source code in src/ references the old digest field or LaunchMeasurement directly, so the schema change flows through cleanly via the dependency. The four existing tests cover parsing, content-type mapping, tag extraction, and formatting, which is sufficient for this fixture-only change.

PR #1247, opened 2026-08-21
fix(cron): bound metrics-partition DDL lock wait so it can't stall message processing

aliel:
## Problem
The message-processing subprocess froze for ~30 minutes right after "Persisting scoring metrics", which starved the RabbitMQ heartbeat (connection declared stuck after 1803s, in-flight `publish` cancelled) and left the CCN unable to process incoming messages — the pending backlog kept climbing.

Root cause is a lock coupling, not raw I/O (a few-hundred-row insert can't take 30 min):

- `insert_node_metrics` (`post.py:369`) does a synchronous `INSERT` into the partitioned `crn_metrics`/`ccn_metrics` — on the message-processing event loop.
- `MetricsPartitionCronJob` takes **ACCESS EXCLUSIVE** on the parent tables for `DETACH`/`CREATE PARTITION`.
- When the cron waits for that lock (behind an in-flight scoring INSERT), its **pending exclusive request queues ahead of subsequent INSERTs** (Postgres fair queuing). Those INSERTs — and therefore the whole message-processing loop — block until the cron's DDL resolves. On the contended HDD this stretched to ~30 min.

The `CancelledError` / "Future exception was never retrieved" and "cancelled by signal" are downstream symptoms (dropped MQ connection + supervisor restart), not the disease.

## Fix
Run each table's partition maintenance in its own transaction under a bounded `SET LOCAL lock_timeout = '5s'`. If the parent lock is contended, the DDL aborts, the job **defers that table and retries next run** (partition maintenance is idempotent), and the write path is never blocked for more than the timeout. Maintenance yields to message processing instead of the reverse.

Per-table transactions also mean a lock timeout on one table doesn't undo the other's maintenance.

## Test
`test_cron_defers_table_when_parent_lock_is_contended`: holds `ACCESS EXCLUSIVE` on `crn_metrics` in a second connection (as an in-flight INSERT effectively does), runs the cron with `lock_timeout` shrunk to 200ms, and asserts it returns without hanging or raising, logs the deferral warning for `crn_metrics`, leaves that table's new partition uncreated, and still creates `ccn_metrics`'s partition.

## Notes
- Pairs with the balance/credit cron event-loop fix (#1244) — same class of "a synchronous DB operation blocks the async loop" problem.
- Root amplifier remains the DB sharing an HDD spindle with IPFS; moving it to SSD removes the contention that turns a "brief" lock into a long one.
- If a stall recurs from a different lock source, `pg_stat_activity` + `pg_blocking_pids` during the block will identify it.

### DIFF
--- a/src/aleph/jobs/cron/metrics_partition_job.py
+++ b/src/aleph/jobs/cron/metrics_partition_job.py
@@ -25,6 +25,7 @@ import logging
 from typing import Iterable, List, Tuple
 
 from sqlalchemy import text
+from sqlalchemy.exc import OperationalError
 
 from aleph.db.models.cron_jobs import CronJobDb
 from aleph.jobs.cron.cron_job import BaseCronJob
@@ -40,6 +41,14 @@ from aleph.types.db_session import DbSession, DbSessionFactory
 LOGGER = logging.getLogger(__name__)
 
 PARTITIONED_TABLES = ("crn_metrics", "ccn_metrics")
+
+# Bound how long partition DDL may wait for the parent table's ACCESS EXCLUSIVE
+# lock. DETACH/CREATE PARTITION need that lock, and a scoring-metrics INSERT
+# holds ROW EXCLUSIVE; without a cap the maintenance job can queue ahead of — and
+# block — the message-processing INSERT path for a long time, stalling the whole
+# message-processing event loop. If the lock is contended the job defers this
+# table and retries next run (partition maintenance is idempotent).
+PARTITION_LOCK_TIMEOUT = "5s"
 
 
 class MetricsPartitionCronJob(BaseCronJob):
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (dismissed):
The fix correctly addresses a lock-coupling problem where partition DDL maintenance could block the message-processing INSERT path for extended periods. The approach of per-table transactions with bounded `SET LOCAL lock_timeout` is sound — if the DDL can't acquire the lock within 5s, it aborts, defers that table, and retries next run, while the other table still gets maintained. The test is thorough, verifying both the contended and uncontended table outcomes. Two minor observations: the `OperationalError` catch is broader than just `lock_timeout` (will also catch connection failures), though the retry-next-run behavior is still reasonable in those cases; and `run` returns `None` instead of `False` when a table is deferred, meaning `last_run` advances and the deferred table waits for the next scheduled interval — likely intentional given daily cron cadence, but worth confirming.

**`src/aleph/jobs/cron/metrics_partition_job.py`** (line 94): The `except OperationalError` catch is broader than just `lock_timeout` — it will also swallow genuine connection errors or other operational failures. The comment says "lock_timeout fired" but that's not the only possible cause. Not a bug (deferring and retrying next run is reasonable for any transient operational error), but the comment could note that other OperationalError causes are also acceptable to defer.

**`src/aleph/jobs/cron/metrics_partition_job.py`** (line 74): When a table is deferred due to lock contention, `run` returns `None` (implicit), so `last_run` advances per the `BaseCronJob` contract. The deferred table won't be retried until the next scheduled interval (likely 24h). This is probably fine for transient contention, but the `run` method *could* return `False` to signal incomplete work and trigger an immediate retry on the next tick, per the convention in `BaseCronJob.run`'s docstring.

foxpatch-aleph (approved):
Well-reasoned fix for a production issue where partition maintenance DDL could block the message-processing event loop. The bounded `SET LOCAL lock_timeout` with per-table transactions and deferral is the correct approach: maintenance yields to the write path instead of blocking it, and idempotent retry ensures partitions eventually get created. The `OperationalError` catch is appropriately broad for an idempotent job. The test uses real PostgreSQL to verify lock contention behavior, deferral logging, per-table isolation, and idempotent retry. No correctness, security, or quality issues found.

PR #1248, opened 2026-08-22
fix(perf): stop channels cache growth and O(n) pubsub dedup scan

aliel:
Two findings from the memory-leak analysis.

## 1. `channels.get_channels` cache grows with traffic (the RSS culprit)
`@cached(ttl=7200, cache=SimpleMemoryCache)` had **no `key_builder`**, so aiocache's default key includes `str(args)` — i.e. the **per-request `Session`'s repr (its memory address)**. Every request opened a fresh `with session_factory() as session`, producing a distinct key, so:

- every call was a **cache miss** that stored a **new entry** (the cache never hit, and each request re-ran `get_distinct_channels`), and
- `SimpleMemoryCache` has no max size, so entries accumulated one-per-request until the 2h TTL evicted them → RSS rises with request volume.

**Fix:** a constant `key_builder` (the channel list is independent of which Session is passed), so it caches as intended — one entry, real hits, no growth. Adds a regression test asserting two distinct `Session` objects resolve to a single underlying query (fails on the old per-Session key).

The other three `@cached` in `metrics.py` are fine — they take no args or app-level singletons (stable keys), so they cache correctly.

## 2. P2P pubsub dedup: O(n) membership on the hot ingress path
`incoming_channel` used `deque(maxlen=200000)` for the seen-message cache, so the per-message `key in seen_hashes` check was an **O(n) linear scan** (up to 200k) on every inbound pubsub message — a CPU/throughput drag on a busy node.

**Fix:** use an `OrderedDict` as a bounded set — **O(1) membership** with identical FIFO eviction and bound (`SEEN_HASHES_MAXLEN = 200_000`). Dedup semantics (by `(sender, item_hash, signature)`) are unchanged.

## Not in this PR
The analysis's findings #2–#4 (PendingTxPublisher / second IpfsService / subprocess engine not disposed on shutdown) are **one-shot per-process resource leaks**, not traffic-proportional RAM growth — they belong in a separate "clean shutdown / resource disposal" PR. #5 is latent dead code.

### DIFF
--- a/src/aleph/services/p2p/protocol.py
+++ b/src/aleph/services/p2p/protocol.py
@@ -1,6 +1,6 @@
 import asyncio
 import logging
-from collections import deque
+from collections import OrderedDict
 from typing import Any
 
 from aleph_p2p_client import AlephP2PServiceClient
@@ -12,6 +12,9 @@ from aleph.types.message_status import InvalidMessageException
 
 LOGGER = logging.getLogger(__name__)
 
+# Max entries kept in the per-topic dedup cache before FIFO eviction.
+SEEN_HASHES_MAXLEN = 200_000
+
 
 async def incoming_channel(
     p2p_client: AlephP2PServiceClient, topic: str, message_publisher: MessagePublisher
@@ -19,7 +22,10 @@ async def incoming_channel(
     LOGGER.debug("incoming channel started...")
 
     await p2p_client.subscribe(topic)
-    seen_hashes: deque[tuple[Any, Any, Any]] = deque([], maxlen=200000)
+    # Dedup cache: OrderedDict used as a bounded set — O(1) membership with
+    # FIFO eviction. A deque here made the per-message membership check an
+    # O(n) linear scan (up to SEEN_HASHES_MAXLEN) on the hot ingress path.
+    seen_hashes: "OrderedDict[tuple[Any, Any, Any], None]" = OrderedDict()
 
     while True:
         try:
@@ -37,23 +43,6 @@ async def incoming_channel(
                     # and such things...
                     try:
                         message_dict = await decode_pubsub_message(message.body)
-                        # Implemented an in-memory cache to avoid deal with the same messages different times.
-                        if (
-                            message_dict["sender"],
-                            message_dict["item_hash"],
-                            message_dict["signature"],
-                        ) in seen_hashes:
-                            # Messages are already ACKed on underlying implementation in p2p_client.receive_messages()
-                            # if the process don't have issues
-                            continue
-
-                        seen_hashes.append(
-                            (
-                                message_dict["sender"],
-                                message_dict["item_hash"],
-                                message_dict["signature"],
-                            )
-                        )
                     except InvalidMessageException:
                         LOGGER.warning(
                             "Received invalid message on P2P topic %s from %s",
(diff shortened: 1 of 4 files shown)


foxpatch-aleph (dismissed):
Both fixes are correct and well-documented. The channels.py key_builder fix properly addresses the cache leak by producing a constant cache key independent of the per-request Session. The protocol.py change from deque to OrderedDict correctly converts the O(n) membership check to O(1) while preserving FIFO eviction semantics. The test for the channels cache is a solid regression test. The only minor gap is the lack of a test for the pubsub dedup change, but the change is small and mechanical enough that this is non-blocking.

**`src/aleph/services/p2p/protocol.py`** (line 28): The string-literal type annotation `"OrderedDict[tuple[Any, Any, Any], None]"` is unnecessary since `OrderedDict` is imported at module level. A direct annotation `OrderedDict[tuple[Any, Any, Any], None]` would be cleaner. Non-blocking nit.

**`src/aleph/services/p2p/protocol.py`** (line 60): No test coverage for the dedup behavior change (FIFO eviction at SEEN_HASHES_MAXLEN, O(1) membership). The change is mechanical, but a small unit test asserting that duplicate messages are skipped and that eviction works would add confidence. Non-blocking.

foxpatch-aleph (dismissed):
Both fixes are correct and well-targeted. The channels cache fix uses a constant `key_builder` to avoid aiocache's default per-Session key (which included the Session object's memory address repr), eliminating the traffic-proportional cache growth. The pubsub dedup fix replaces a `deque` (O(n) membership scan) with an `OrderedDict`-backed bounded set giving O(1) membership with identical FIFO eviction semantics. The `_BoundedSeenCache` class is clean and well-documented, and the test coverage is appropriate for both changes. The PR correctly scopes itself to the two traffic-proportional issues and defers one-shot resource leaks to a separate PR.

**`src/aleph/web/controllers/channels.py`** (line 20): Minor note: aiocache's `key_builder` is called as `key_builder(f, *args, **kwargs)` where `f` is the decorated function. The `lambda *args, **kwargs: "used_channels"` works correctly since `*args` captures `f`, but a reader unfamiliar with aiocache's call convention might wonder why the function arg is silently absorbed. A comment like `# key_builder receives (f, *args, **kwargs); we ignore all of them` could help, though the existing comment above already explains the intent adequately.

**`src/aleph/services/p2p/protocol.py`** (line 38): Behavioral improvement over the original deque: if the same key is re-added, `OrderedDict.__setitem__` is a no-op on the value and does not create a duplicate or change insertion order. The old deque would have appended a duplicate entry, wasting a slot. This is strictly better for dedup purposes — no action needed, just noting the subtle semantic difference.

foxpatch-aleph (approved):
Both fixes are correct, well-documented, and well-tested. The channels cache key_builder fix correctly addresses the per-Session memory leak by using a constant key, and the diagnosis is verified against aiocache's actual key_builder call convention. The _BoundedSeenCache correctly replaces the O(n) deque membership scan with an OrderedDict-backed O(1) bounded set with identical FIFO eviction semantics. Behavioral equivalence of the dedup logic is preserved — the placement change (from inside the try to after the except) is functionally equivalent since both only dedup valid messages. No concurrency, security, or edge-case issues found.

PR #1249, opened 2026-08-22
fix(cleanup): close leaked MQ/IPFS/DB resources on shutdown

aliel:
Resource-disposal fixes for **one-shot per-process leaks** surfaced by the memory-leak analysis — connection/pool drift that lingers for the process lifetime (and lingers on the server after an ungraceful restart), *not* traffic-proportional RAM growth. Complements the earlier cache/dedup fixes.

## Changes
- **`PendingTxPublisher` owns and closes its MQ connection.** Previously the connection was a local in `make_pending_tx_exchange` and only the exchange was kept, so the connection was never closed. The publisher is now an async context manager (closes the connection on `__aexit__`) and is entered on the startup exit stack. `make_pending_tx_exchange` is inlined into `PendingTxPublisher.new`.
- **`network.listener_tasks` reuses the shared `IpfsService`/`StorageService`.** It was building a *second* `IpfsService` directly (not via `.new()`, not on any stack), abandoning its aiohttp pool for the process lifetime. It now takes the caller's already-managed services as parameters.
- **Subprocesses dispose the DB engine and close their standalone MQ connection on shutdown.** `fetch_pending_messages`, `process_pending_txs`, and `process_pending_messages` never disposed the engine (and the first two never closed their `make_mq_conn` connection). Added a `disposing_engine` helper and put the engine + connection in the outer `async with` so they're cleaned up on shutdown.
- **Removed dead code in `_make_pending_queue`.** The `if not channel:` branch opened an MQ connection it never closed; all callers pass a channel, so it was unreachable. `channel` is now required.

## Tests
Unit tests for the two new cleanup contracts: `disposing_engine` disposes on normal and exceptional exit, and `PendingTxPublisher` closes its connection on context exit. Both are DB-free.

## Scope note
These are graceful-shutdown / resource-hygiene fixes; they don't change runtime behavior on a live node (resources are used the same way, just closed on exit). Subprocess recycling already bounded these, so impact is connection drift on restart rather than live growth.

### DIFF
--- a/src/aleph/chains/chain_data_service.py
+++ b/src/aleph/chains/chain_data_service.py
@@ -225,26 +225,21 @@ class ChainDataService:
                 raise InvalidContent(error_msg)
 
 
-async def make_pending_tx_exchange(config: Config) -> aio_pika.abc.AbstractExchange:
-    mq_conn = await aio_pika.connect_robust(
-        host=config.p2p.mq_host.value,
-        port=config.rabbitmq.port.value,
-        login=config.rabbitmq.username.value,
-        password=config.rabbitmq.password.value,
-        heartbeat=config.rabbitmq.heartbeat.value,
-    )
-    channel = await mq_conn.channel()
-    pending_tx_exchange = await channel.declare_exchange(
-        name=config.rabbitmq.pending_tx_exchange.value,
-        type=aio_pika.ExchangeType.TOPIC,
-        auto_delete=False,
-    )
-    return pending_tx_exchange
-
-
 class PendingTxPublisher:
-    def __init__(self, pending_tx_exchange: aio_pika.abc.AbstractExchange):
+    def __init__(
+        self,
+        mq_conn: aio_pika.abc.AbstractConnection,
+        pending_tx_exchange: aio_pika.abc.AbstractExchange,
+    ):
+        self.mq_conn = mq_conn
         self.pending_tx_exchange = pending_tx_exchange
+
+    async def __aenter__(self) -> Self:
+        return self
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb) -> None:
+        # Own the MQ connection so it is closed on shutdown instead of leaked.
+        await self.mq_conn.close()
 
     @staticmethod
     def add_pending_tx(session: DbSession, tx: ChainTxDb):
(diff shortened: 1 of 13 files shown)


foxpatch-aleph (dismissed):
Well-scoped resource-disposal fixes that correctly close MQ connections, DB engines, and reuse shared IPFS/storage services instead of leaking duplicates. The changes are functionally correct, all callers of the modified functions are updated, and the removed dead code was indeed unreachable. Tests cover the new cleanup contracts. The only minor inconsistency is that PendingTxPublisher.__aexit__ doesn't use safe_async_cleanup like the other MQ connection closures in commands.py, so a close() failure during shutdown could mask the original exception, but this is a non-blocking nit.

**`src/aleph/chains/chain_data_service.py`** (line 242): Consider wrapping `self.mq_conn.close()` in a try/except (or using `safe_async_cleanup` like `commands.py:145` does for its MQ connection). If the RabbitMQ server is unreachable on shutdown, this close() will raise and could mask the original exception that triggered the exit. The other MQ resources in the shutdown path use `safe_async_cleanup` to log and swallow such errors.

foxpatch-aleph (dismissed):
This is a well-executed resource-disposal cleanup PR. The changes are correct: PendingTxPublisher now owns and closes its MQ connection via the async context manager protocol, listener_tasks correctly reuses shared services instead of building (and leaking) a second IpfsService, and all three subprocess tasks now dispose their DB engine and close their MQ connections on shutdown via async context managers. The dead code removal in _make_pending_queue is clean — all callers already pass a channel. Tests cover the two new cleanup contracts (disposing_engine on normal/exception exit, PendingTxPublisher closing its connection). A pre-existing MQ connection leak in PendingMessageProcessor (process_pending_messages.py) is out of scope for this PR. Minor note: there's an inconsistency in error handling — PendingTxPublisher uses safe_async_cleanup to swallow close errors while the subprocesses use raw async with mq_conn which propagates them — but this is non-blocking.

**`src/aleph/jobs/process_pending_messages.py`** (line 163): Pre-existing (not introduced by this PR): PendingMessageProcessor.new() creates an MQ connection at line 71 that is never closed. The async with pending_message_processor at line 196 calls MqWatcher.__aexit__ which only cancels the watcher task — it does not call PendingMessageProcessor.close(). Consider adding mq_conn disposal to the async with block or overriding __aexit__ to call close().

**`src/aleph/jobs/fetch_pending_messages.py`** (line 302): Minor inconsistency: async with mq_conn will propagate close() errors on shutdown, whereas PendingTxPublisher.__aexit__ uses safe_async_cleanup to log-and-swallow. Consider wrapping with safe_async_cleanup for consistency, though this is non-blocking since these are subprocess exit paths.

foxpatch-aleph (approved):
Well-structured resource-disposal PR that fixes four genuine connection/pool leaks on shutdown. The PendingTxPublisher now owns and closes its MQ connection via an async context manager; listener_tasks reuses the shared IpfsService/StorageService instead of building (and abandoning) a second one; subprocesses wrap their DB engine in disposing_engine and MQ connections in closing_quietly; and dead code in _make_pending_queue is removed. The new closing_quietly and disposing_engine helpers are clean, the safe_async_cleanup pattern is applied consistently across all shutdown paths, and the tests cover the new cleanup contracts (normal exit, exception exit, and close-error swallowing). No correctness or security issues found.

**`src/aleph/chains/chain_data_service.py`** (line 245): Minor: there is no test for the exception-swallowing path of PendingTxPublisher.__aexit__ (i.e. mq_conn.close() raising). The safe_async_cleanup behavior itself is tested in test_lifecycle.py, but a dedicated test would make the contract self-documenting, matching the closing_quietly test in test_lifecycle.py:62.

**`tests/jobs/test_process_pending_messages_cleanup.py`** (line 25): Calling __aexit__ directly without __aenter__ works because _watcher_task is None, but it would be more robust to test via `async with pending_message_processor:` to also exercise the super().__aexit__ path (watcher task cancellation + connection close). The current MqWatcher.__aexit__ does `await self._watcher_task` which can raise CancelledError — the test doesn't cover that interaction.

PR #1250, opened 2026-08-22
fix(db): bound lock/statement/idle timeouts on the worker DB engine

aliel:
## Why
Message processing and the cron jobs run **synchronous** psycopg2 on a single asyncio loop. A single query that blocks, waiting forever on a lock, or running unbounded — freezes the **entire** loop, and with it the RabbitMQ heartbeat and all message processing. This is the root pattern behind the two stalls fixed earlier (the credit/balance cron and the metrics-partition DDL lock): the node was frozen for ~30 minutes until the lock cleared.

By default Postgres never times these out (`lock_timeout = 0`, `statement_timeout = 0`, `idle_in_transaction_session_timeout = 0`), so there is no floor on how long the loop can be wedged.

## What
Set three per-session timeouts on the engine built by `make_engine`, via libpq `connect_args={"options": ...}`, configurable under `config.postgres` (each `0` disables):

| setting | default | caps |
|---|---|---|
| `lock_timeout_ms` | 30000 | time waiting to acquire a lock |
| `statement_timeout_ms` | 120000 | total runtime of a single statement |
| `idle_in_transaction_session_timeout_ms` | 300000 | a transaction left open but idle (reclaims leaked "idle in transaction" sessions — we observed one parked ~1h) |

A blocked/runaway query now fails fast with `OperationalError` (message retried) instead of freezing the loop. This is a safety net that would have **contained both prior incidents**.

## Scope / safety
- Applies to every engine from `make_engine` (API, main process, and the `fetch`/`txs`/`process` subprocesses) — all the right targets.
- **Migrations are unaffected**: `deployment/migrations/env.py` builds its own `create_engine`, so long backfills (e.g. 0056) won't be killed.
- The metrics-partition cron's `SET LOCAL lock_timeout = '5s'` still overrides the connection default within its own transaction.
- Defaults are generous (a real query never waits 30s on a lock or runs 120s); they only trip on genuinely stuck/runaway work, and ops can tune or disable per deployment.

Adds unit tests for the options builder (all set / some zero / all zero).

## Note
This is the "engine timeouts" safety-net item from the perf backlog; it complements — not replaces — the targeted fixes already merged (chunked crons, partition `lock_timeout`, cache/dedup, resource disposal). It does **not** address the `message_counts` trigger contention (separate, measure-first item).

## Update: API exempt from statement_timeout
Some aggregate reads legitimately run longer than the worker default — `address_aggregate` refreshes dirty aggregates on read, and `refresh_aggregate` does a full `jsonb_merge` over every element, which can exceed 120s for a large aggregate on a slow disk. The API path doesn't retry, so capping it would surface a 500.

`make_engine` now takes a `statement_timeout_ms` override; the API entrypoint passes `0` (uncapped statement runtime), while `lock_timeout` and `idle_in_transaction_session_timeout` still apply there. The workers (fetch/txs/process) and the main process keep the full set. The real fix for slow aggregate reads is the separate dirty-refresh optimization on the perf backlog.

### DIFF
--- a/src/aleph/api_entrypoint.py
+++ b/src/aleph/api_entrypoint.py
@@ -43,6 +43,11 @@ async def configure_aiohttp_app(
             config,
             echo=config.logging.level.value == logging.DEBUG,
             application_name="aleph-api",
+            # Some aggregate reads (dirty-refresh does a full jsonb_merge over
+            # all elements) can legitimately run longer than the worker default,
+            # so don't cap statement runtime on the API. lock_timeout and
+            # idle_in_transaction still apply.
+            statement_timeout_ms=0,
         )
         session_factory = make_session_factory(engine)
 
--- a/src/aleph/config.py
+++ b/src/aleph/config.py
@@ -228,6 +228,18 @@ def get_defaults():
             "pool_pre_ping": True,
             # Recycle DB connections after this many seconds of inactivity.
             "pool_recycle": 3600,
+            # Fail a statement that waits more than this long (ms) to acquire a
+            # lock. Bounds how long a blocked query can freeze the synchronous
+            # worker event loop (a lock wait took the node down for ~30 min).
+            # 0 disables.
+            "lock_timeout_ms": 30000,
+            # Abort any single statement running longer than this (ms), so a
+            # runaway query cannot wedge the loop. 0 disables. Migrations use a
+            # separate engine and are unaffected.
+            "statement_timeout_ms": 120000,
+            # Abort a transaction left open but idle longer than this (ms),
+            # reclaiming leaked "idle in transaction" sessions. 0 disables.
+            "idle_in_transaction_session_timeout_ms": 300000,
         },
         "ipfs": {
             # Whether to enable storage and communication on IPFS.
--- a/src/aleph/db/connection.py
+++ b/src/aleph/db/connection.py
@@ -88,6 +88,7 @@ def make_engine(
     config: Optional[Config] = None,
     echo: bool = False,
     application_name: Optional[str] = None,
+    statement_timeout_ms: Optional[int] = None,
 ) -> Engine:
     if config is None:
         config = get_config()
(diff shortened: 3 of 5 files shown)


foxpatch-aleph (dismissed):
This PR correctly adds per-session PostgreSQL timeouts (lock_timeout, statement_timeout, idle_in_transaction_session_timeout) to the worker DB engine via libpq connect_args, preventing a single blocked or runaway query from freezing the synchronous event loop. The _timeout_options helper is clean, correctly skips zero-valued timeouts (0 = disabled), and produces an empty connect_args when all are zero. Security is sound — values are integers from config, no injection risk. The migration engine is confirmed to use its own create_engine, so long-running backfills are unaffected. The metrics partition cron's SET LOCAL lock_timeout correctly overrides within its own transaction. Test coverage covers all three branches (all set, some zero, all zero). No bugs or issues found.

**`src/aleph/db/connection.py`** (line 85): Minor nit: there's only one blank line between `_timeout_options` and `make_engine`. PEP 8 recommends two blank lines between top-level functions. This won't affect functionality but would be flagged by pycodestyle.

**`src/aleph/db/connection.py`** (line 117): The async engine (`make_async_engine`) does not get the same timeout treatment. It's currently never called in the codebase so this is fine for now, but if it's ever used in production, it would need the same safety net. Not a blocker for this PR — just a note for the future.

foxpatch-aleph (dismissed):
The PR correctly sets per-session lock, statement, and idle-transaction timeouts on the worker DB engine via libpq connect_args, with a clean override mechanism for the API to disable statement_timeout. The _timeout_options helper is well-factored, the all-zero case correctly omits connect_args entirely, and migrations are confirmed to bypass make_engine. Tests cover the key code paths (all-set, partial-zero, all-zero, and the API override). The code is clean, well-commented, and addresses a real production incident pattern.

**`tests/db/test_connection.py`** (line 38): Minor: consider adding a test for the all-zero case at the make_engine level — verifying that `connect_args` is absent (or an empty dict) when all three config timeouts are 0, not just that the options string is empty. The current test only covers the override path for statement_timeout.

**`tests/conftest.py`** (line 99): Heads-up: the test fixture calls make_engine with the real config (get_defaults), so test sessions will now have lock_timeout=30s and statement_timeout=120s applied. Unlikely to cause issues for unit tests, but if you ever see intermittent lock-timeout failures during parallel schema setup, this is the source.

foxpatch-aleph (approved):
Well-scoped safety-net PR that sets per-session lock/statement/idle-transaction timeouts on the worker DB engine via libpq connect_args. The implementation is correct: zero values are properly skipped, the API exemption for statement_timeout is cleanly handled via an override parameter, migrations are unaffected (they use their own engine), and the existing SET LOCAL lock_timeout in the metrics partition cron still takes precedence. Tests cover all relevant combinations (all set, all zero, partial zero, API override). One minor style nit: only a single blank line separates _timeout_options from make_engine where PEP 8 expects two.

**`src/aleph/db/connection.py`** (line 86): Only one blank line between `_timeout_options` and `make_engine` — PEP 8 calls for two blank lines between top-level definitions. The project's `hatch run linting:all` may flag this.

PR #1251, opened 2026-08-27
feat(cost): price V-PROGRAM artifacts and the runtime bundle

odesenfans:
## Summary
V-PROGRAMs were priced as confidential instances on vCPUs/memory only; every artifact they reference (workload image, hash trees, verified volumes, and the runtime bundle named by the manifest) is a user-published STORE the CRN downloads in full, so the disk footprint was unpriced and unbounded. Since `/api/v0/price/{hash}` is what the credit billing service charges per second, what `check_balance` admits against, and what the CLI/console show, this is fixed in the cost model.

- **Dedicated `vprogram` product type** (`ProductPriceType.VPROGRAM`), borrowing `instance_confidential` numbers when the price aggregate lacks the key while keeping its own identity (`resolve_price_type_key`, `ProductPricing.with_type`); same 20480 MiB per-CU disk allowance as instances.
- **Every artifact is billed as an `EXECUTION_VPROGRAM_VOLUME` execution volume**, sized from the CCN's pinned file size only (never from user-declared numbers), netted against the CU allowance by the existing `EXECUTION_VOLUME_DISCOUNT`. Row names: `workload`, `workload:hash_tree`, `#i:<comment>`, `#i:<comment>:hash_tree`, `runtime`.
- **Runtime bundle resolution** (`services/vprogram_runtime.py`): the pinned manifest is read (capped at 1 MiB) only to find `bundle.ref`; its declared `size`/`sha256` are ignored. Invalid manifest -> `InvalidVProgramRuntime` (new error code 305, migration 0067); transiently unreadable manifest or unpinned bundle -> `VmVolumeNotFound` (retry). The resolved ref is persisted as `VProgramDb.runtime_bundle_ref` (migration 0068), FORGET-protected like the other artifacts, and resolved once per message via a per-handler cache.
- **Estimation**: `estimated_size_mib` / `estimated_hash_tree_size_mib` on workload and volumes plus `runtime_estimated_size_mib` so `/price/estimate` can quote before upload (estimate wins over the manifest; `ge=0`).
- **Recalculation** reads the persisted bundle ref (resolver fallback for legacy rows) and no longer deletes a message's cost rows before the replacement rows are computed.

## Effects
- Newly processed V-PROGRAMs get `EXECUTION` row name `vprogram`; artifact footprint above 20 GiB x CUs is billed at the storage rate per MiB-hour. Existing processed V-PROGRAMs keep their old price until `/price/{hash}/recalculate` is run.
- New DB migrations: 0067 (error code 305), 0068 (`vms.runtime_bundle_ref`).
- Message acceptance now requires the CCN to read the runtime manifest content.

## Test plan
- `hatch run testing:test tests -q`: 999 passed, 10 skipped, 0 failed (6 errors in `tests/chains/test_ethereum.py` need a local RPC node; pre-existing/environmental)
- `hatch run linting:all` clean (ruff, black, isort, mypy)
- New coverage: product type fallback, under/over-allowance arithmetic (service and end-to-end handler), manifest resolver failure modes, single-resolution cache, estimate precedence, FORGET protection of the bundle, recalculation keeping rows on failure

https://claude.ai/code/session_01PEusSpUkyY6EU1JayZJpHh


foxpatch-aleph (dismissed):
Well-structured PR that correctly prices V-PROGRAM artifacts as execution volumes, with a clean runtime manifest resolver (proper transient vs permanent error handling), a bounded bundle-ref cache, a price type fallback that preserves product identity, and a recalculation path that avoids deleting cost rows before replacements exist. Test coverage is thorough across all new code paths. The minor observations (0-estimate falsiness, display-only size mismatch, comment-ending-with-:hash_tree edge case) are non-blocking.

**`src/aleph/services/cost.py`** (line 87): `if estimated_size_mib:` is falsy for `0`, so a 0 estimate falls back to `RefVolume` (file-based sizing) instead of using the estimate. The schema allows `ge=0`, so 0 is valid. Harmless in practice but slightly inconsistent with intent — consider `if estimated_size_mib is not None:` if the distinction matters.

**`src/aleph/services/cost.py`** (line 1101): When `runtime_estimated_size_mib` is used and the manifest is pinned, `get_cost_component_size_mib` returns the manifest file size instead of the estimated bundle size. This only affects the `size_mib` display in `/price/estimate` — the actual cost is correct. The tests pass because the manifest is not pinned in those test scenarios.

**`src/aleph/services/cost.py`** (line 1012): If a verified volume's `comment` ends with `:hash_tree`, the `endswith(":hash_tree")` check would incorrectly return the hash tree estimate for the volume row. Edge case — same pattern exists in the existing immutable volume code.

foxpatch-aleph (dismissed):
Well-structured PR that correctly prices V-PROGRAM artifacts as execution volumes against the per-CU disk allowance. The runtime manifest resolver is robust (size cap, recursion guard, transient vs. permanent error distinction), the price-type fallback mechanism preserves product identity, the UNION ALL refactor of get_vms_dependent_volumes is a clear performance improvement, and the recalculation flow correctly defers row deletion until replacements are ready. Migration IDs are valid hex and chain correctly. Test coverage is comprehensive across all new code paths. No blocking issues found.

**`deployment/migrations/versions/0067_b45ceae89f65_vprogram_runtime_invalid_error_code.py`** (line 25): The INSERT INTO error_codes lacks `ON CONFLICT (code) DO NOTHING`. If code 305 already exists (e.g. from a partially-applied migration), this will fail with a primary key violation. Migration 0037 used the conflict guard for the same pattern. Not blocking since other migrations (0030, 0045) also omit it, but consider adding for robustness.

**`src/aleph/services/cost.py`** (line 87): `_vprogram_artifact` uses `if estimated_size_mib:` which treats `estimated_size_mib=0` as falsy, falling back to RefVolume (pinned file size lookup). This is arguably correct (a 0 MiB estimate is meaningless), but worth a brief comment noting the intent since the schema allows `ge=0`.

**`src/aleph/handlers/content/vprogram.py`** (line 137): The `_bundle_refs` OrderedDict cache is not protected against concurrent access from multiple async tasks. In practice the single-threaded event loop likely prevents true concurrency, but if the pipeline ever processes messages in parallel, two tasks could resolve the same manifest before either caches it. Benign race (same content-addressed result), but worth noting.

foxpatch-aleph (dismissed):
Well-implemented pricing for V-PROGRAM artifacts with thorough error handling, excellent test coverage, and sound architectural decisions. The runtime bundle resolver correctly distinguishes transient from permanent failures, the pricing model properly bills artifacts against the CU disk allowance, the recalculation path safely preserves old cost rows until replacements are computed, and the FORGET dependency check protects the bundle STORE. Minor observations (dropped max_length on estimation volumes, benign async cache race) are non-blocking.

**`src/aleph/schemas/cost_estimation_messages.py`** (line 109): The `volumes` override drops the `max_length=MAX_VERIFIED_VOLUMES` constraint from the base `VerifiableProgramContent`. While this only affects the cost estimation endpoint (not message validation), keeping the constraint would be more consistent and prevent surprise estimates with an unreasonable number of volumes.

**`src/aleph/handlers/content/vprogram.py`** (line 143): The `_bundle_refs` cache has a benign race: two concurrent asyncio tasks processing messages with the same `runtime_ref` can both miss the cache and both call `resolve_runtime_bundle_ref`. Since manifests are content-addressed, both resolve to the same `bundle_ref` and the cache ends up correct. If this becomes a concern, an `asyncio.Lock` per `runtime_ref` (or a single lock around the cache) would eliminate the duplicate resolution.

**`src/aleph/services/cost.py`** (line 1116): The priority change in `get_cost_component_size_mib` (estimate-first instead of file-size-first) is correct for estimation content — `_get_estimated_size_from_content` returns `None` for non-estimation content types, so the real file size is still used for production cost rows. Worth noting that for existing estimation types (STORAGE, EXECUTION_PROGRAM_VOLUME_CODE, etc.), the truthiness check `if content.estimated_size_mib:` means an estimate of 0 is treated as "no estimate" and falls through to the file size, while the new V-PROGRAM types use `_optional_float` which correctly treats 0 as a value. This pre-existing inconsistency is not introduced by this PR.

foxpatch-aleph (dismissed):
Well-structured PR with comprehensive test coverage. The runtime manifest resolver correctly distinguishes permanent failures (InvalidVProgramRuntime) from transient ones (VmVolumeNotFound), the cost model properly handles 0 MiB estimates with `is not None` checks, and the recalculation safety improvement (delete-rows-after-compute) prevents data loss on resolver failures. The UNION ALL refactor for FORGET dependency checks is correct and more efficient. The priority change in get_cost_component_size_mib (estimate-first) is safe for existing types since their truthiness checks on 0 estimates fall through to the real file size, matching prior behavior.

**`src/aleph/web/controllers/prices.py`** (line 500): When the resolver raises `VmVolumeNotFound`, `_message_processing_error_detail` returns just the volume ref hash (e.g. `"cafecafe..."`) rather than a human-readable message, so the warning reads `"runtime bundle not priced: cafecafe..."`. Consider including more context in the warning, e.g. `f"runtime bundle not priced: manifest {content.runtime.ref} references an unpinned volume"` for the `VmVolumeNotFound` case. Minor UX nit, not blocking.

**`src/aleph/services/vprogram_runtime.py`** (line 112): `content.value` is accessed without checking its type. If `get_hash_content` returns bytes rather than str, `json.loads` handles it fine, but if it returns None for some edge case, this would raise an unhandled `TypeError`. The `ValueError` catch covers `json.JSONDecodeError` (a subclass of `ValueError`) but not `TypeError` or `AttributeError`. Consider adding `TypeError` to the exception tuple for robustness.

**`src/aleph/handlers/content/vprogram.py`** (line 137): The `_bundle_refs` cache is a mutable instance attribute on the handler. Since the handler is a singleton (one per `MessageHandler`), this cache persists across messages. The comment says content-addressed STOREs can't go stale, which is correct. However, if a manifest is ever found to be invalid after caching (e.g. a bug in resolution), the stale entry would persist until evicted by LRU. This is a very low risk since manifests are content-addressed, but worth noting.

foxpatch-aleph (commented):
Well-structured PR that prices V-PROGRAM artifacts previously unpriced. The runtime manifest resolver has proper permanent-vs-transient error categorization, the per-handler cache avoids redundant manifest reads, and recalculation preserves old cost rows until replacements are computed. Migrations use valid hex revision IDs with partial indexes matching the polymorphic discriminator. Test coverage is comprehensive. One real bug: the `endswith(":hash_tree")` check in `_get_estimated_size_from_content` is ambiguous when a volume's comment field ends with `:hash_tree`, causing the ref row to incorrectly match and return the hash_tree estimate. A couple of minor documentation gaps in the `size_mib` field description and a test docstring that claims negative-case coverage it doesn't actually test.

**`src/aleph/services/cost.py`** (line 1041): The `cost.name.endswith(":hash_tree")` check is ambiguous when a volume's `comment` field ends with `:hash_tree`. For example, `comment = "data:hash_tree"` produces ref row name `#0:data:hash_tree` and hash_tree row name `#0:data:hash_tree:hash_tree`. The ref row incorrectly matches `endswith(":hash_tree")` and returns `estimated_hash_tree_size_mib` instead of `estimated_size_mib`. Consider comparing against the full expected name (e.g. strip the `:hash_tree` suffix, look up the volume, and check whether the original name equals `#{index}:{comment}:hash_tree`) or using a more precise suffix test.

**`src/aleph/schemas/api/costs.py`** (line 78): The `size_mib` field description lists all cost types it applies to but omits the new `EXECUTION_VPROGRAM_VOLUME`. Same for line 135.

**`tests/db/test_vprograms_db.py`** (line 103): The docstring says "Each case also exercises the negative side: an unknown hash must return None" but the test only checks the positive case. Add an assertion with an unknown hash returning None to match the documented behavior.

foxpatch-aleph (approved):
This PR introduces a dedicated V-PROGRAM product type with proper pricing of all artifacts (workload image, hash trees, verified volumes, and the runtime bundle) as execution volumes against the per-CU disk allowance. The implementation is thorough and well-structured: the runtime manifest resolver correctly distinguishes permanent failures from transient issues with appropriate retry/reject semantics, the handler-level LRU cache avoids redundant manifest reads, and the recalculation path safely defers cost-row deletion until replacement rows are computed. The UNION ALL refactor of `get_vms_dependent_volumes` with partial indexes is a clean performance improvement. Test coverage is comprehensive, covering under/over-allowance arithmetic, manifest failure modes, cache behavior, estimate precedence, FORGET protection of the bundle, and recalculation edge cases. I found no bugs or security issues.

**`src/aleph/services/cost.py`** (line 996): Pre-existing (not introduced by this PR): `_get_estimated_size_from_content` uses truthiness checks (`if content.estimated_size_mib:`) for the older cost types (STORAGE, EXECUTION_PROGRAM_VOLUME_CODE, etc.), which means an estimate of 0 MiB is treated as "no estimate" and falls through to the pinned file size. The new V-PROGRAM code correctly uses `_optional_float` with `is not None` semantics. Consider aligning the older branches in a follow-up.

**`src/aleph/types/cost.py`** (line 175): Pre-existing (not introduced by this PR): `ProductPricing.from_aggregate` sets `compute_unit` based on `if tiers` rather than `if compute_unit`. If `compute_unit` is present in the aggregate but `tiers` is absent, `compute_unit` will be None, which would cause `_get_nb_compute_units` to fall back to the default. Not an issue for any current product type (they all have both or neither), but worth noting.

**`src/aleph/services/vprogram_runtime.py`** (line 112): The `json.loads(content.value)` call receives bytes from `RawContent.value`. `json.loads` accepts both `str` and `bytes` in Python 3.6+, so this is correct. Just noting it in case someone later changes `RawContent.value` to a non-JSON-compatible type.

PR #1252, opened 2026-08-27
test(db): reset the test database per test instead of re-migrating it

odesenfans:
## Summary
`session_factory` dropped the schema and re-ran all 68 migrations for every DB-backed test (~0.8 s per test on CI, ~9 min of pytest per run; 30+ min on fsync-bound local disks). This PR migrates once per session and resets per test.

- **Session-scoped `migrated_db`**: one drop + `alembic upgrade head`, then a snapshot of the seeded tables (`alembic_version`, `cron_jobs`, `error_codes`, discovered by row count) into a `seed` schema.
- **Per-test reset** (`session_factory`, contract unchanged): probe non-empty tables with `EXISTS`, `DELETE` them under `SET LOCAL session_replication_role = replica` with a 5 s `lock_timeout`, restore seeds, `ALTER SEQUENCE ... RESTART` for read sequences, re-enable disabled user triggers, drop tables a test created. If an expected table is missing the schema is rebuilt. `TRUNCATE` was measured and rejected: it fsyncs a new relfilenode per relation (~112 ms/table locally, 4.5 s per reset).
- **`fresh_schema` marker** forces a full rebuild for tests that alter the schema (`test_migration_0061_column_additions_are_rerunnable` is marked). Drift detection compares table name sets only; DDL on existing tables is the marker's job.
- **Teardown guard**: a test that leaks a checked-out connection now fails at its source.
- **Migration 0017** no longer hard-codes `aleph.public.rejected_messages` (same table, same rows; nodes that already applied it are unaffected), and `run_db_migrations` passes `-x db_url=` (env.py ignored alembic's `tag=`). Covered by a test that migrates a scratch database under another name.
- Markers are now registered in pytest's list form (the inline-table form registered nothing; `ledger_hardware` was never registered either).

## Numbers
- Local full suite (fsync-bound disk): 30+ min before, **51 s** after (1026 passed, 6 pre-existing Ethereum-RPC errors).
- `session_factory` setup: ~0.8 s -> ~0.03 s per test; `tests/db/test_credit_balances.py` 288 s -> 7.5 s locally.
- CI `tests` job baseline on `main`: ~10 min (7.5 min in May). Fill in the after number from this PR's run.

## Requirements
The test role must be a superuser (`session_replication_role`, `CREATE DATABASE` in the migration test); CI's `postgres:15.1` with `POSTGRES_USER: aleph` already is. A non-superuser now gets a clear error.

https://claude.ai/code/session_01PEusSpUkyY6EU1JayZJpHh

### DIFF
--- a/deployment/migrations/versions/0017_f9fa39b6bdef_vm_instances.py
+++ b/deployment/migrations/versions/0017_f9fa39b6bdef_vm_instances.py
@@ -50,7 +50,7 @@ def reprocess_failed_instance_messages():
             """
         UPDATE message_status
           SET status = 'pending'
-          FROM aleph.public.rejected_messages rm
+          FROM rejected_messages rm
           WHERE message_status.item_hash = rm.item_hash
             AND rm.message ->> 'type' = 'INSTANCE'
         """
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -225,7 +225,7 @@ norecursedirs = [ "*.egg", "dist", "build", ".tox", ".venv", "*/site-packages/*"
 testpaths = [ "tests/unit" ]
 markers = [
   "ledger_hardware: marks tests as requiring ledger hardware",
-  "fresh_schema: rebuild the database schema from migrations for this test instead of the fast TRUNCATE reset",
+  "fresh_schema: rebuild the database schema from migrations for this test instead of the fast DELETE-based reset",
 ]
 
 [tool.coverage.run]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -223,7 +223,10 @@ filterwarnings = [
 ]
 norecursedirs = [ "*.egg", "dist", "build", ".tox", ".venv", "*/site-packages/*", ".claude", ".worktrees" ]
 testpaths = [ "tests/unit" ]
-markers = { ledger_hardware = "marks tests as requiring ledger hardware" }
+markers = [
+  "ledger_hardware: marks tests as requiring ledger hardware",
+  "fresh_schema: rebuild the database schema from migrations for this test instead of the fast TRUNCATE reset",
+]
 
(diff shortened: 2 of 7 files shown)


foxpatch-aleph (approved):
Well-structured performance optimization replacing per-test migrations with a session-scoped migration plus fast DELETE-based reset. The reset logic correctly handles table drift, sequence rewinding, trigger re-enabling, and seed restoration. The migration 0017 fix removes a hard-coded database name dependency. SQL interpolation is safe (catalog-sourced values only). Test coverage for the new fixture logic is thorough with ordered step tests and graceful skip mechanics. Minor brittleness in trigger count assertions is acceptable.

**`tests/db/test_session_factory.py`** (line 119): The assertion `assert enabled == 2` is brittle: if a future migration adds another user trigger, this test will fail with no obvious connection to the migration that caused it. Consider asserting `enabled >= 2` or computing the expected count from the freshly migrated schema.

**`tests/conftest.py`** (line 254): After `ALTER SEQUENCE ... RESTART`, `pg_sequences.last_value` is set to the restart value (typically 1), so the sequence remains "dirty" on the next reset. This is harmless (restarting an already-restarted sequence is idempotent) but means every subsequent reset does an unnecessary `ALTER SEQUENCE` for each previously-used sequence. Not a correctness issue, just a minor perf note.

**`tests/conftest.py`** (line 327): When the rebuild path is taken, `config = aleph.config.app_config` reads the config set by `mock_config`. This works because `mock_config` runs before `session_factory` (function-scoped fixture dependencies are resolved first). But if someone reorders the fixture parameters, `config` would be the config from the previous `session_factory` rebuild (or the session-scoped `migrated_db` config). Consider adding `mock_config` as an explicit dependency or documenting the ordering requirement.

PR #1257, opened 2026-08-28
fix(ipfs): blank deprecated Provider/Reprovider config on kubo >= v0.38

aliel:
## Problem
kubo v0.43 **refuses to start** when the deprecated `Provider`/`Reprovider` config carries values. An older version of this script set `Reprovider.Strategy 'pinned'`, so upgraded nodes crash on boot. The previous attempt (a repo `--to=17 --allow-downgrade` round-trip) **crash-looped fresh installs**: the downgrade restores `config.17-to-18.bak`, which doesn't exist on a repo born at v18.

## Fix
kubo v0.43 *tolerates* these keys when empty — a fresh v0.43 init writes empty `"Provider": {}` / `"Reprovider": {}` blocks itself. So just blank them:
```sh
ipfs config --json Reprovider '{}'
ipfs config --json Provider '{}'
```
- **Fresh/clean config:** no-op (already empty).
- **Upgraded node with a stale `Reprovider.Strategy`:** replaced with `{}` → daemon starts.
- No repo migration, no downgrade, no crash loop.

**Gated on kubo >= v0.38.0**, where these keys are deprecated and superseded by `Provide`. Below that they're still *active* config, so the legacy branch keeps setting `Reprovider.Strategy`. An unparseable/empty version string — or any future kubo 1.x — takes the modern path rather than re-adding keys that stop a modern daemon.

### Carry-over
`Provider.Enabled`, `Provider.WorkerCount` and `Reprovider.Interval` are read and re-applied as `Provide.Enabled`, `Provide.DHT.MaxWorkers` and `Provide.DHT.Interval` before blanking, so a node that tuned them doesn't silently fall back to defaults. Verified against kubo's own `fs-repo-17-to-18` migration:
- `provider["WorkerCount"] → dht["MaxWorkers"]`, `reprovider["Interval"] → dht["Interval"]`, `dht → provide["DHT"]`
- `Provider.Strategy` is deliberately not carried (that migration skips it as unused); `Reprovider.Strategy` is superseded by the `Provide.Strategy 'pinned'` set here.

Repo-format migration for nodes coming from older kubo is handled by the daemon's `--migrate` flag (all four compose files run `daemon --enable-gc --migrate` on the pinned `ipfs/kubo:v0.43.0` image), so no `ipfs repo migrate` is needed here.

## Hardening
- **`set_config` wrapper**: every key is applied through a helper that records a failure instead of aborting, so one key rejected by a future kubo can't stop the remaining settings — exactly the failure mode that caused this bug. Failures are **tallied and the script exits 1** at the end, so a node never quietly runs on defaults with only a warning line. Verified with a stubbed `ipfs`: all-OK → exit 0; two keys rejected → both logged, all remaining keys still applied, exit 1. (`set -eu` was already present and is kept for real errors.)
- Quoted `$CONFIG_FILE` in the backup test.
- Removed a duplicated bootstrap peer (`QmNnooDu7bf…` was listed twice).
- Stripped trailing whitespace.

## Note
Out of scope but worth a look separately: this script sets `Datastore.StorageMax '10GB'` while the committed `kubo.json` sets `1500GB`. The script runs on every container start, so it wins — combined with `--enable-gc` and `GCPeriod 12h` that could GC hard against a 10GB ceiling on a node with a large pinset.

### DIFF
--- a/deployment/scripts/001-update-ipfs-config.sh
+++ b/deployment/scripts/001-update-ipfs-config.sh
@@ -5,58 +5,95 @@ set -eu
 
 CONFIG_FILE="/data/ipfs/config"
 
+# Apply one config key, recording instead of aborting when kubo rejects it, so
+# a key removed by a future kubo release cannot stop the remaining settings from
+# being applied (that is how a single stale key broke startup before). Failures
+# are tallied and the script exits non-zero at the end, so a node never quietly
+# runs on defaults.
+config_failures=0
+set_config() {
+    if ! ipfs config "$@"; then
+        echo "ERROR: failed to apply ipfs config $*" >&2
+        config_failures=$((config_failures + 1))
+    fi
+}
+
 if [ -f "$CONFIG_FILE" ]; then
     cp "$CONFIG_FILE" "$CONFIG_FILE.backup"
 fi
 
 echo "Updating IPFS config file..."
 
-# The deprecated Provider/Reprovider config is removed in kubo v0.43 (the daemon
-# refuses to start with it present). Repo version 18 -- and the migration to it --
-# was introduced in kubo v0.38.0, so only migrate on a binary that supports it
-# (bump the 38/18 targets on future kubo upgrades). The daemon must be stopped
-# while `ipfs repo migrate` runs (it locks the repo).
-if [ -f /data/ipfs/version ]; then
-    ver=$(cat /data/ipfs/version)
-    kubo_minor=$(ipfs version --number | cut -d. -f2)
+# Only announce recursively pinned CIDs. The reprovide strategy key changed in
+# kubo v0.38.0: Reprovider/Provider were deprecated and superseded by the
+# unified Provide key. v0.43 refuses to start when the deprecated keys carry
+# values (an older version of this script set Reprovider.Strategy), but tolerates
+# them empty, so blank them there. On older kubo those keys are still ACTIVE and
+# must not be erased. Repo-format migration is handled by the daemon's --migrate.
+#
+# Assume the current (>= v0.38) layout unless the version string clearly says
+# otherwise: an unparseable version, or any kubo 1.x+, must not fall through to
+# the legacy branch and re-add keys that stop a modern daemon from starting.
+kubo_version=$(ipfs version --number 2>/dev/null || true)
+kubo_major=$(printf '%s' "$kubo_version" | cut -d. -f1)
+kubo_minor=$(printf '%s' "$kubo_version" | cut -d. -f2)
+case "$kubo_major" in
+    ''|*[!0-9]*) kubo_major=0 ;;
+esac
+case "$kubo_minor" in
+    ''|*[!0-9]*) kubo_minor=38 ;;
+esac
 
-    if [ "$kubo_minor" -ge 38 ]; then
-        if [ "$ver" -lt 18 ]; then
-            # migrate repo to the latest version
-            ipfs repo migrate --to=18
-        elif [ "$ver" -eq 18 ]; then
-            # A node that ran the old script has legacy Provider/Reprovider keys.
-            # Re-run the migration to strip them.
-            if ipfs config Provider >/dev/null 2>&1 || ipfs config Reprovider >/dev/null 2>&1; then
-                echo "legacy Provider/Reprovider present"
-                ipfs repo migrate --to=17 --allow-downgrade
-                ipfs repo migrate --to=18
-                echo "removed Provider/Reprovider"
-            fi
-        fi
-    fi
+if [ "$kubo_major" -gt 0 ] || [ "$kubo_minor" -ge 38 ]; then
+    # Carry the deprecated values over to their Provide equivalents before
+    # blanking, so a node that tuned them does not silently lose the setting.
+    # Mapping matches kubo's own fs-repo-17-to-18 migration:
+    # Provider.Enabled -> Provide.Enabled, Provider.WorkerCount ->
+    # Provide.DHT.MaxWorkers, Reprovider.Interval -> Provide.DHT.Interval.
+    # (Provider.Strategy is intentionally not carried: the migration skips it as
+    # unused. Reprovider.Strategy is superseded by the Provide.Strategy set
+    # below.)
+    provider_enabled=$(ipfs config Provider.Enabled 2>/dev/null || true)
+    provider_workers=$(ipfs config Provider.WorkerCount 2>/dev/null || true)
+    reprovider_interval=$(ipfs config Reprovider.Interval 2>/dev/null || true)
+
+    set_config --json Reprovider '{}'
+    set_config --json Provider '{}'
+
+    case "$provider_enabled" in
+        true|false) set_config --json Provide.Enabled "$provider_enabled" ;;
+    esac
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (dismissed):
This PR replaces a fragile repo downgrade/migrate round-trip (which crash-looped fresh installs) with a simple, correct approach: blank the deprecated Provider/Reprovider config keys on kubo >= v0.38 where they are deprecated. kubo v0.43 tolerates these keys when empty but refuses to start with values in them, so writing '{}' is a no-op on fresh installs and fixes upgraded nodes. The gating on kubo_minor >= 38 is correct — below that these keys are active config. The compose files confirm the daemon is started with --migrate, so repo-format migration is handled separately. The script runs as a container init script before the daemon starts, so ipfs config commands operate offline. The removal of the /data/ipfs/version file guard is safe because ipfs version --number reports the binary version, not the repo version.

foxpatch-aleph (dismissed):
This PR replaces a fragile repo-migration round-trip (downgrade to v17 then re-upgrade to v18) that crash-looped fresh installs with a simpler approach: blanking the deprecated Provider/Reprovider config keys via `ipfs config --json ... '{}'` on kubo >= v0.38, and setting `Provide.Strategy 'pinned'` in the same branch. The else-branch preserves the legacy `Reprovider.Strategy` for older kubo. The logic is correct, the version-gating is sound, and the compose files confirm the daemon's `--migrate` flag handles repo-format migration. The script hardening (`set -eu`, quoted `$CONFIG_FILE`) is welcome. No bugs, security issues, or correctness problems found.

**`deployment/scripts/001-update-ipfs-config.sh`** (line 20): The `kubo_minor` extraction via `cut -d. -f2` works for both `0.43.0` and `v0.43.0` formats, but it would break if `ipfs version --number` ever returned a multi-digit minor with a suffix (e.g., `0.43.1-rc1` yields `43` which is fine, but a hypothetical `0.43.1-dirty` also yields `43`). This is not a problem today and the container image is pinned to a specific kubo version, so no change needed — just noting for awareness.

**`deployment/scripts/001-update-ipfs-config.sh`** (line 46): Pre-existing (not introduced by this PR): the bootstrap peer `QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN` is listed twice on lines 45-46. kubo deduplicates automatically, but the duplicate entry could be removed for cleanliness in a future PR.

foxpatch-aleph (dismissed):
The fix correctly replaces the crash-looping repo-migration round-trip with simple JSON blanking of deprecated Provider/Reprovider keys to {}, which kubo v0.43 tolerates. The set_config wrapper is a solid defensive improvement that prevents one rejected key from aborting all remaining config. Version gating with fallback to 38 on unparseable versions is safe-by-default. The carry-over logic for Provider.Enabled/WorkerCount is properly guarded with case statements. The duplicate bootstrap peer removal and trailing whitespace cleanup are good housekeeping. No bugs or security issues found.

**`deployment/scripts/001-update-ipfs-config.sh`** (line 53): Worth verifying that `Provide.DHT.MaxWorkers` is the exact key name in kubo v0.43's config schema for the Provider.WorkerCount equivalent. The mapping looks reasonable but this is the only key that isn't a direct 1:1 rename.

**`deployment/scripts/001-update-ipfs-config.sh`** (line 11): The PR description mentions hardening with `set -eu`, but `set -eu` was already present in the previous version of the script. The actual hardening is the `set_config` wrapper — consider updating the PR description to reflect this accurately.

foxpatch-aleph (dismissed):
This fix is correct and well-verified: I confirmed against upstream kubo sources that v0.43 stops with an error when deprecated Provider/Reprovider settings carry values, that the Provide section and repo 17-to-18 migration arrived in v0.38 (making the >= 38 gate correct), and that the carry-over mappings (Provider.Enabled -> Provide.Enabled, Provider.WorkerCount -> Provide.DHT.MaxWorkers, Provider.Strategy dropped as unused) match kubo's own deprecation annotations. Dropping the script-side repo migration is safe because all four compose files run daemon --migrate on the pinned ipfs/kubo:v0.43.0 image. Blanking the keys to {} is idempotent and matches what kubo itself serializes for a fresh init. The set_config wrapper correctly prevents a single rejected key from aborting the whole init script. The version-parse guard (empty/non-numeric -> assume >= 0.38) is fail-safe in the right direction, and sh -n passes. Remaining points are non-blocking nicks listed in comments.

**`deployment/scripts/001-update-ipfs-config.sh`** (line 45): Nit: `Reprovider.Interval` (mapped by kubo to `Provide.DHT.Interval`) is not carried over, so an operator who tuned the reprovide interval silently falls back to the default. Since `Reprovider.Strategy` gets an explicit `Provide.Strategy` below this doesn't matter for the 'pinned' case the old script wrote, but adding `Reprovider.Interval` -> `Provide.DHT.Interval` alongside the other two carry-overs would make this complete.

**`deployment/scripts/001-update-ipfs-config.sh`** (line 11): The wrapper swallows failures for every key including critical ones (e.g. `Bootstrap`, `Routing.Type`), so a rejected write leaves the node running on defaults with only one log line to notice. Consider tallying failures in set_config and `exit 1` at the end of the script: all keys still get attempted (unlike the old set -e mid-script abort), but the container exits loudly on the first run where something went wrong.

**`deployment/scripts/001-update-ipfs-config.sh`** (line 33): Minor latent foot-gun: a hypothetical future kubo 1.x would parse minor=0, fail the >= 38 check, and take the legacy branch, re-adding Reprovider.Strategy. Purely theoretical while the image is pinned to v0.43.0 (and 0.100.0 etc. parse fine), but a comment noting the gate assumes kubo stays 0.x, or a major/minor compare, would future-proof it.

foxpatch-aleph (approved):
Correct and well-hardened fix. I verified the version gating, legacy-value carry-over, and failure tallying with a stubbed ipfs binary: 0.43 takes the modern path, 0.37 takes the legacy path, unparseable/garbage versions fail safe to the modern path, carry-over reads happen before blanking with sensible input guards, and rejected keys are tallied and logged while remaining settings still apply (exit 1). Removing the repo-migration round-trip is safe since all four compose files pin kubo v0.43.0 with daemon --migrate. The set_config wrapper directly addresses the failure mode that caused the original bug. Non-blocking: no automated test coverage exists for this script despite two boot-failure incidents, and the StorageMax 10GB vs 1500GB conflict the author flagged deserves follow-up.

**`deployment/scripts/001-update-ipfs-config.sh`** (line 14): Given this script has now caused two separate boot-failure incidents (the original Reprovider.Strategy breakage, then the migration round-trip crash-loop), it would be worth committing the stubbed-ipfs test harness described in the PR description alongside the script — even a minimal `tests/` fixture that asserts the version-gate branches and the exit-1-on-failure contract. I re-implemented the harness to verify and it caught all the behaviors described, so the approach works well as a regression test.

**`deployment/scripts/001-update-ipfs-config.sh`** (line 107): Echoing the author's own note: this 10GB overrides the 1500GB in the committed `deployment/docker-build/*/kubo.json` on every container start, and combined with `--enable-gc` and GCPeriod 12h a node with a large pinset will GC aggressively against the 10GB ceiling. Definitely out of scope here, but worth a tracking issue so it isn't lost.

PR #1258, opened 2026-08-31
fix(ethereum): always fetch logs in bounded batches

odesenfans:
## Problem

Fresh installs cannot sync Ethereum on-chain messages with most RPC providers.

`_get_logs` first tried to fetch everything in a single `eth_getLogs` request from the last synced height (11.4M on a fresh install) to `latest`, and only fell back to batched fetching when the provider rejected the request with error code `-32005` (Infura's convention). Providers that signal oversized ranges differently (`-32000`, `-32602`, "block range too large", HTTP 413, or plain timeouts) made the request fail without triggering the fallback. The exception bubbled up to the retry loop in `fetch_sync_events_task`, which restarted the sync from scratch: an infinite retry loop with zero progress. Established nodes were unaffected since their block delta is small, which is why only fresh installs hit this.

## Solution

There is no reliable way to detect every provider-specific failure mode, so stop trying the optimistic full-range request altogether:

- `_get_logs` now always fetches in batches of at most `max_block_range` blocks (default 100,000). The existing range-halving logic for `-32005` responses within a batch is kept.
- When the node is already synced to the chain tip, return without issuing a request. This case was previously absorbed by the `start..latest` request; without a guard, every poll on a caught-up node would send `fromBlock > toBlock`, which some providers reject.
- Removed the now-unneeded `Literal["latest"]` typing.

A fresh mainnet install now issues ~120 bounded requests instead of one unbounded one.

## Tests

- `test_sync_fetches_logs_in_bounded_block_ranges`: spies on the actual filters passed to `eth_getLogs` and asserts every request spans an explicit range of at most `max_block_range` blocks while all emitted events are still picked up.
- `test_sync_when_already_at_chain_tip_requests_no_logs`: asserts a caught-up sync issues no `eth_getLogs` request at all.

Full `tests/chains/` suite passes locally (43 passed, 3 skipped) against anvil.

### DIFF
--- a/src/aleph/chains/ethereum.py
+++ b/src/aleph/chains/ethereum.py
@@ -2,17 +2,7 @@ import asyncio
 import importlib.resources
 import json
 import logging
-from typing import (
-    Any,
-    AsyncIterator,
-    Dict,
-    List,
-    Literal,
-    Self,
-    Tuple,
-    TypedDict,
-    Union,
-)
+from typing import Any, AsyncIterator, Dict, List, Self, Tuple, TypedDict, Union
 
 from aleph_message.models import Chain
 from configmanager import Config
@@ -53,7 +43,7 @@ class GetLogsException(Exception): ...
 
 class TooManyLogsInRange(GetLogsException):
     start_block: BlockNumber
-    end_block: BlockNumber | Literal["latest"]
+    end_block: BlockNumber
 
 
 class FeeEstimate(TypedDict, total=False):
@@ -172,7 +162,7 @@ class EthereumConnector(ChainWriter):
     async def _get_logs_in_block_range(
         self,
         start_block: BlockNumber,
-        end_block: BlockNumber | Literal["latest"] = "latest",
+        end_block: BlockNumber,
     ) -> List[LogReceipt]:
         """
         Retrieves logs from the Aleph message sync contract and handles RPC-specific exceptions.
@@ -197,9 +187,6 @@ class EthereumConnector(ChainWriter):
             # Unexpected issue, pass the exception to the caller
             raise
 
-    async def _get_all_logs(self, start_block: BlockNumber) -> List[LogReceipt]:
-        return await self._get_logs_in_block_range(start_block, "latest")
-
     async def _get_all_logs_in_batches(
         self, start_block: BlockNumber, max_block_range: int
     ) -> AsyncIterator[LogReceipt]:
@@ -207,6 +194,10 @@ class EthereumConnector(ChainWriter):
 
         while True:
             last_eth_block = await self.web3_client.eth.block_number
+            # Already synced to the chain tip, nothing to fetch. Some RPC
+            # providers reject requests where fromBlock > toBlock.
+            if start_block > last_eth_block:
+                return
             # Note: the range in get_logs is [start, end].
             end_block = min(last_eth_block, BlockNumber(start_block + block_range - 1))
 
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (approved):
My review is complete. Here's my assessment:

**Correctness:** The core change is sound. The batch loop was already correct (inclusive `[start, end]` ranges, `start_block = end_block + 1` advance, halving on `-32005` with a floor that prevents infinite loops), and the new `start_block > last_eth_block` guard correctly handles the caught-up case that the removed `"latest"` request used to absorb — no events can be missed since no blocks exist past `last_eth_block`, and the range math has no off-by-one. Removal of `Literal["latest"]` is consistent with the code: `_get_logs_in_block_range` and `TooManyLogsInRange` are only ever used with concrete `BlockNumber`s now, and the deleted `_get_all_logs` had no other callers.

**Tests:** Both new tests directly exercise the behavioral contract (bounded explicit ranges, no request at chain tip) and follow the existing patterns in the file. I couldn't execute them here (the `web3_client` fixture needs the Docker-hosted anvil node), but static analysis is consistent with the author's reported pass.

**Non-blocking observations:** a pre-existing quirk in the halving flow (if `TooManyLogsInRange` fires on the final batch, the halved retry is skipped by the `break`), a harmless `self.max_block_range` vs. parameter inconsistency on line 210, and a suggestion to assert range contiguity in the spy test.

```json
{
  "verdict": "APPROVE",
  "summary": "Correct and well-scoped fix. Removing the optimistic full-range fetch in favor of always-bounded batches is the right call given that providers signal oversized requests in incompatible ways, and the new chain-tip guard correctly prevents inverted fromBlock > toBlock requests without risking missed events. Range math has no off-by-one (inclusive [start, end], start advances to end + 1, halving has a floor that raises instead of looping forever), the removed Literal typing is consistent with all remaining call sites, and the deleted _get_all_logs has no other callers. Both new tests directly verify the behavioral contract. Only non-blocking notes below. (I could not execute the chain tests locally as they require the Docker-hosted anvil fixture.)"
  "comments": [
    {
      "path": "src/aleph/chains/ethereum.py",
      "line": 210,
      "body": "Pre-existing nit: this resets to self.max_block_range while the method also takes a max_block_range parameter (initialized into block_range on line 193). Harmless in the sole caller since both values are identical, but the parameter is effectively ignored after any successful batch."
    },
    {
      "path": "src/aleph/chains/ethereum.py",
      "line": 221,
      "body": "Pre-existing quirk, not introduced by this PR: if TooManyLogsInRange fires on the batch that ends at last_eth_block, block_range is halved but this break exits the loop without retrying the range, ending the fetch early. It self-heals on the next poll (sync status only advances per processed event, so start_block is unchanged), so no data is lost — just a wasted poll. Worth a follow-up if you touch this loop again."
    },
    {
      "path": "tests/chains/test_ethereum.py",
      "line": 552,
      "body": "The boundedness assertion is good; consider also asserting contiguity — each filter's fromBlock equals the previous toBlock + 1 (after the first). That would directly catch gap/overlap regressions (e.g. start_block advancing by more than end_block + 1), which the current span check alone would miss. The call_count == 3 assertion gives partial coverage for the 3 emitted events, so this is only a suggestion."
    }
  ]
}
```

**`src/aleph/chains/ethereum.py`** (line 210): Pre-existing nit: this resets to self.max_block_range while the method also takes a max_block_range parameter (initialized into block_range on line 193). Harmless in the sole caller since both values are identical, but the parameter is effectively ignored after any successful batch.

**`src/aleph/chains/ethereum.py`** (line 221): Pre-existing quirk, not introduced by this PR: if TooManyLogsInRange fires on the batch that ends at last_eth_block, block_range is halved but this break exits the loop without retrying the range, ending the fetch early. It self-heals on the next poll (sync status only advances per processed event, so start_block is unchanged), so no data is lost — just a wasted poll. Worth a follow-up if you touch this loop again.

**`tests/chains/test_ethereum.py`** (line 552): The boundedness assertion is good; consider also asserting contiguity — each filter's fromBlock equals the previous toBlock + 1 (after the first). That would directly catch gap/overlap regressions (e.g. start_block advancing by more than end_block + 1), which the current span check alone would miss. The call_count == 3 assertion gives partial coverage for the 3 emitted events, so this is only a suggestion.

PR #1259, opened 2026-08-31
fix(metrics): do not report -1 as the ETH reference height on provider errors

odesenfans:
## Problem

`fetch_eth_height()` returns `-1` when the Ethereum RPC provider replies with an error (e.g. a 429 rate limit). The value is cached for 10 minutes and exported as `pyaleph_status_chain_eth_height_reference_total = -1`, which makes `pyaleph_status_chain_eth_height_remaining_total = max(-1 - last_committed_height, 0) = 0`. A failing provider therefore makes the node look fully synced.

## Fix

Return `None` instead (with a warning log), matching the behavior of `fetch_reference_total_messages()`. The Prometheus formatter and the remaining-height computation already skip `None`, so both metrics are simply omitted until the provider recovers.

## Tests

Added `test_fetch_eth_height_returns_none_on_client_response_error`, which mocks the web3 provider to raise `aiohttp.ClientResponseError` and asserts the function returns `None`.

### DIFF
--- a/src/aleph/web/controllers/metrics.py
+++ b/src/aleph/web/controllers/metrics.py
@@ -177,7 +177,8 @@ async def fetch_eth_height() -> Optional[int]:
         w3 = AsyncWeb3(provider)
         return await w3.eth.block_number
     except aiohttp.ClientResponseError:
-        return -1
+        LOGGER.warning("ETH height could not be obtained")
+        return None
     finally:
         await provider.disconnect()
 
--- a/tests/web/controllers/test_metrics.py
+++ b/tests/web/controllers/test_metrics.py
@@ -1,8 +1,13 @@
 from dataclasses import dataclass
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import aiohttp
+import pytest
 
 from aleph.web.controllers.metrics import (
     BuildInfo,
     Metrics,
+    fetch_eth_height,
     format_dataclass_for_prometheus,
     format_dict_for_prometheus,
 )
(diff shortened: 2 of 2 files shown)


foxpatch-aleph (approved):
Correct, minimal fix. Reporting -1 as the ETH reference height made eth_height_remaining_total collapse to 0 on any provider error, falsely signaling a fully synced node; returning None is the right sentinel, matches fetch_reference_total_messages(), and is handled correctly downstream (format_dataclass_for_prometheus skips None fields and get_metrics omits the remaining-height metric when the reference is None). I confirmed web3 raises aiohttp.ClientResponseError on 4xx via raise_for_status() in its session manager, so the premise of the fix is valid, and I ran the test file locally: all 4 tests pass, including the new one. One behavioral nuance worth knowing: aiocache treats a stored None as a cache miss (if value is not None: return value), so on failure the RPC is now retried on every get_metrics call (~every 10s, gated by the metrics cache) rather than backing off for the 10-minute TTL as the PR description implies — benign and arguably better recovery. Remaining comments are minor nits.

**`src/aleph/web/controllers/metrics.py`** (line 181): Note: with None returned here, aiocache's @cached treats the stored value as a miss (`if value is not None: return value`), so on provider failure the RPC is re-attempted on every get_metrics call (~every 10s via the metrics cache) rather than being held for the 10-minute TTL as the PR description implies. That's arguably better recovery behavior, so no change needed — just flagging that the failure is no longer TTL-cached. If negative-result caching were ever desired, aiocache supports `skip_cache_func`.

**`src/aleph/web/controllers/metrics.py`** (line 161): Pre-existing (not introduced by this PR), but since this PR adds the same message in fetch_eth_height: this log line in fetch_reference_total_messages says "ETH height could not be obtained" for what is a reference-node message-count failure — a copy-paste that makes failures of the two functions indistinguishable in logs. Worth fixing in a follow-up.

**`tests/web/controllers/test_metrics.py`** (line 66): Works, but assigning a manually-built coroutine to w3.eth.block_number is subtle — it only raises because the production code does `await w3.eth.block_number`, and would silently degrade to a non-awaited-coroutine warning if that ever changed. `w3.eth.block_number = AsyncMock(side_effect=error)` would be more idiomatic and fail more clearly. A companion success-path test (e.g. ethereum.enabled = False returning None, or a returned block number) would also round out coverage. Minor nits; the test as written does pass.

PR #1260, opened 2026-08-31
fix(config): lower the default ethereum.max_block_range from 100k to 10k

odesenfans:
Lowers the default `ethereum.max_block_range` from 100,000 to 10,000 blocks.

Many RPC providers reject or rate-limit `eth_getLogs` queries spanning very large block ranges, which breaks batch synchronization with the 100k default. 10k stays within common provider limits. Operators with permissive endpoints can still raise the value in their node configuration.

One-line change to the default in `src/aleph/config.py`; no code path depends on the previous value.

### DIFF
--- a/src/aleph/config.py
+++ b/src/aleph/config.py
@@ -191,7 +191,7 @@ def get_defaults():
             # Ethereum block height to start from when fetching sync events.
             "start_height": 11_400_000,
             # Largest range of blocks used when synchronizing messages in batches.
-            "max_block_range": 100_000,
+            "max_block_range": 10_000,
             # Delay in seconds between publication attempts.
             "commit_delay": 35,
             # Maximum gas price accepted when publishing to Ethereum.


foxpatch-aleph (approved):
Minimal, correct one-line default change with no side effects. max_block_range is consumed only by the Ethereum connector's batched log fetch (src/aleph/chains/ethereum.py:154, used at lines 206/219), where it is a starting batch size that self-adjusts via halving on TooManyLogsInRange, so lowering it from 100k to 10k is safe and aligns with common RPC provider eth_getLogs range limits. No tests, docs, or deployment samples hard-code the old value. Only note: max_block_range governs just the batched fallback path — the initial fetch in _get_logs still requests start..latest in one call, and only RPC error -32005 triggers batching — so the default change won't fully shield nodes from providers whose range limits differ, but that is outside this PR's scope.

**`src/aleph/config.py`** (line 194): Non-blocking observation: this default only affects the batched fallback in _get_all_logs_in_batches. The initial fetch in _get_logs (src/aleph/chains/ethereum.py:236) still requests start..latest in a single call, and only RPC error code -32005 is mapped to TooManyLogsInRange — so nodes using providers with smaller range limits (some cap at 2k–5k) may still need to set this explicitly. Could be worth a brief note in operator docs, but the change itself is sound.

foxpatch-aleph (approved):
Minimal, correct one-line default change. I verified that ethereum.max_block_range is only consumed by the Ethereum connector (src/aleph/chains/ethereum.py:154) as the initial/ceiling batch size in _get_all_logs_in_batches, which halves the range on TooManyLogsInRange and accepts any positive integer, so 10,000 is safe with no boundary or off-by-one concerns. No tests, sample configs, or docs reference the old 100k default, so nothing else needs updating. The 10x increase in eth_getLogs calls during a full historical resync is a reasonable trade for avoiding provider rejections and rate limits.

PR #1261, opened 2026-08-31
fix(vprogram): reject coerced scalars at ingestion (serde parity)

odesenfans:
Consumer side of aleph-im/aleph-message#160.

## Problem

The node validates V-PROGRAM content with the aleph-message models but stores and serves the **raw** `item_content` dict (`MessageDb.from_pending_message` keeps `content=content_dict`). Any coercion pydantic performs during validation (`"vcpus": "1"` -> `1`, `"policy": 196608.0` -> `196608`, `"internet": "true"` -> `True`) therefore produces a *processed* message whose served content strict decoders (the aleph-rs SDK) cannot parse. One such message poisons the scheduler's page-strict history fetch on every poll (mitigated read-side by aleph-im/aleph-rs#386); this closes the gap at the network acceptance boundary.

Found by differential-fuzzing aleph-message 1.3.0 against aleph-rs v0.17.0. All V-PROGRAM messages ever published already conform, so nothing grandfathered breaks. Scoped to V-PROGRAM (PROGRAM/INSTANCE history may contain coerced legacy messages that must keep re-validating during sync).

## Change

- Bump `aleph-message` to the git rev carrying serde-strict V-PROGRAM scalars.
- Ingestion-level regression tests: coerced scalars raise through `validate_message_content`, integral `time` stays accepted (serde parses integer JSON into f64).

**Merge gate:** swap the git pin for the released aleph-message version once aleph-im/aleph-message#160 is released (same flow as #1220).

https://claude.ai/code/session_01QkaSkidXtLL63prnqo7NS5

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,9 +30,7 @@ dependencies = [
   "aiohttp-swagger3>=0.10.0",
   "aioipfs~=0.7.1",
   "alembic==1.19.1",
-  # TODO: swap for the aleph-message release carrying serde-strict
-  # V-PROGRAM scalars (aleph-im/aleph-message#160) before merging.
-  "aleph-message @ git+https://github.com/aleph-im/aleph-message@62d2cfcf8b1bd535e80b76256602507c9c574f9c",
+  "aleph-message==1.3.1",
   "aleph-nuls2==0.1.1",
   "aleph-p2p-client @ git+https://github.com/aleph-im/p2p-service-client-python@1e992aaacdb0071a34ffb130cbbba9123509e08b",
   "asyncpg==0.31.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ dependencies = [
   "alembic==1.19.1",
   # TODO: swap for the aleph-message release carrying serde-strict
   # V-PROGRAM scalars (aleph-im/aleph-message#160) before merging.
-  "aleph-message @ git+https://github.com/aleph-im/aleph-message@e2feb5c1005a1344ea5c373f9485e0de6279b297",
+  "aleph-message @ git+https://github.com/aleph-im/aleph-message@62d2cfcf8b1bd535e80b76256602507c9c574f9c",
   "aleph-nuls2==0.1.1",
   "aleph-p2p-client @ git+https://github.com/aleph-im/p2p-service-client-python@1e992aaacdb0071a34ffb130cbbba9123509e08b",
   "asyncpg==0.31.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,9 @@ dependencies = [
   "aiohttp-swagger3>=0.10.0",
   "aioipfs~=0.7.1",
   "alembic==1.19.1",
-  "aleph-message==1.3.0",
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (dismissed):
Correct, well-scoped fix at the right boundary. validate_message_content is the single ingestion gate (used by both from_pending_message and the lazy parsed_content property), and ValidationErrors there cleanly become InvalidMessageFormat rejections rather than pipeline crashes. I verified the new tests' expectations against actual pydantic v2 strict-type semantics (StrictFloat accepts int, rejects str; StrictInt rejects 2.0/"2"; StrictBool rejects 1), so the parametrized rejection matrix and the integral-time acceptance test are internally consistent and would fail against the old aleph-message==1.3.0 — proper red/green regression coverage. Scoping to V-PROGRAM only, with PROGRAM/INSTANCE left lax for legacy re-validation, is deliberate and documented, and the existing repair path can reject any legacy non-conforming message. Only flags are the self-documented merge gate (git-SHA pin must be swapped for the release before merging) and a minor style nit on the deep-copy idiom.

**`pyproject.toml`** (line 35): Merge gate noted in the TODO — this must be swapped for the released aleph-message version (aleph-im/aleph-message#160) before merge. The immutable SHA pin is the right interim choice (reproducible, unlike a branch ref). Also worth confirming the version metadata at that commit doesn't conflict with any transitive aleph-message constraints (e.g. from aleph-client), since a git build installs whatever version string the repo declares at that rev — the existing aleph-p2p-client git pin suggests this pattern is fine in practice.

**`tests/messages/test_vprogram.py`** (line 134): Minor nit: `json.loads(json.dumps(...))` as a deep-copy is a bit roundabout; `copy.deepcopy(VPROGRAM_CONTENT)` says the intent directly. Not blocking.

foxpatch-aleph (changes requested):
The test design and logic are correct and well-executed: I verified against aleph-message's git main (self-identified as .1) that all nine parametrized coerced-scalar cases raise ValidationError through validate_message_content (the true ingestion entry point used by MessageDb.from_pending_message), and that integral time remains accepted. However, the PR is not mergeable as-is because aleph-message==1.3.1 is not published on PyPI — a live check of the PyPI JSON API shows the latest available version is 1.3.0 — so the pin fails dependency resolution and any fresh install or CI run will break with 'No matching distribution found'. Either hold this PR until the 1.3.1 release actually lands on PyPI, or temporarily revert to the git-rev pin as the PR description's merge-gate note originally intended.

**`pyproject.toml`** (line 33): Blocking: aleph-message==1.3.1 is not published on PyPI yet — the latest available version is 1.3.0 (confirmed via the PyPI JSON API; the strict-scalar change currently only exists on the aleph-message git main, which self-reports as 1.3.1). As written, dependency resolution fails for any fresh install or CI run. Since the PR description already marks the upstream release as a merge gate, either hold this PR until 1.3.1 actually lands on PyPI or temporarily keep the git-rev pin.

**`tests/messages/test_vprogram.py`** (line 124): Nit: the docstring's first line is a bit garbled — "A scalar serde would reject must not reach processed status." reads more clearly as "A scalar that serde would reject must not reach processed status." Non-blocking.

Issue #161, opened 2022-02-11
Error in logs of `dev` branch 


hoh:
This error does not appear on Sentry, but there are logs of POST to sentry.io 🤔

See logs attached to this file, with the following extract:
```
2022-02-11 13:42:21 [DEBUG] P2P.peers: New message received {'from': b'12D3KooWBfGpZ2zuAjtiSVEynYfCN6dHv5JkFe4m9qYov1uJcNeC', 'data': b'%7B%22address%22%3A%20%22/ip4/51.89.96.79/tcp/4025/p2p/QmYEro66g9xCr5hh57BtVcN32Ly7fxuG4mUuVn489PfgRz%22%2C%20%22interests%22%3A%20null%2C%20%22peer_type%22%3A%20%22P2P%22%2C%20%22version%22%3A%20%22v0.1.11%22%7D', 'seqno': b'\x16\xd0\xf2\xa2\xb8\xa6\x19Q', 'topicIDs': ['ALEPH_ALIVE']}
2022-02-11 13:42:21 [ERROR] P2P.peers: Exception in pubsub peers monitoring, resubscribing
Traceback (most recent call last):
  File "/opt/venv/lib/python3.8/site-packages/aleph/services/peers/monitor.py", line 73, in monitor_hosts_p2p
    await read_pbmsg_safe(stream, pubsub_msg)
  File "/opt/venv/lib/python3.8/site-packages/p2pclient/utils.py", line 25, in read_pbmsg_safe
    len_msg_bytes = await read_unsigned_varint(stream)
  File "/opt/venv/lib/python3.8/site-packages/p2pclient/serialization.py", line 33, in read_unsigned_varint
    data = await stream.receive_exactly(1)
  File "/opt/venv/lib/python3.8/site-packages/anyio/_networking.py", line 250, in receive_exactly
    raise IncompleteRead
anyio.exceptions.IncompleteRead: The stream was closed before the read operation could be completed
2022-02-11 13:42:22 [ERROR] P2P.protocol: Exception in pubsub, reconnecting.
Traceback (most recent call last):
  File "/opt/venv/lib/python3.8/site-packages/aleph/services/p2p/protocol.py", line 195, in incoming_channel
    async for mvalue in sub(topic):
  File "/opt/venv/lib/python3.8/site-packages/aleph/services/p2p/pubsub.py", line 17, in sub
    await read_pbmsg_safe(stream, pubsub_msg)
[...]

PR #1266, opened 2026-09-02
feat: accept Intel TDX confidential instances

odesenfans:
Bumps aleph-message to the TDX schema tip (aleph-im/aleph-message#161, git-sha pin per the usual V-PROGRAM bump pattern): `TeePlatform`/`mode` grow `tdx`, `LaunchMeasurement.registers` becomes a union discriminated on `platform`, and measured instances are credit-only regardless of platform.

No CCN code changes turn out to be needed: TEE validation is fully delegated to aleph-message, and both persistence (`handlers/content/vm.py`) and pricing (`services/cost.py` keys on the *presence* of `trusted_execution`, not its mode) handle a tdx instance as-is, priced as `INSTANCE_CONFIDENTIAL`. V-PROGRAM is untouched: the schema keeps its backend `sev_snp`-only.

The new test processes a `mode: tdx` instance end to end (credit payment, runtime + declared TDX registers) and pins the two DB-visible facts: `environment_trusted_execution_firmware` is NULL and the policy column stays at the schema default, since TDX has no host-chosen launch policy.

**Update**: aleph-message 1.4.0 (carrying #161) is released; the pin now reads `aleph-message==1.4.0` and the tests pass against the PyPI package. No merge gate remains.

Noted while here (pre-existing, not touched): `environment_trusted_execution_policy` is an int4 column, which a real SEV-SNP 64-bit policy can overflow; tdx is unaffected because its policy is forced to the default.

https://claude.ai/code/session_01DUWEzYLgSidwttzQGtzB43

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ dependencies = [
   "aiohttp-swagger3>=0.10.0",
   "aioipfs~=0.7.1",
   "alembic==1.19.1",
-  "aleph-message==1.3.1",
+  "aleph-message @ git+https://github.com/aleph-im/aleph-message@e07d0fb5c01deab3526c0477aba64c3643275a83",
   "aleph-nuls2==0.1.1",
   "aleph-p2p-client @ git+https://github.com/aleph-im/p2p-service-client-python@1e992aaacdb0071a34ffb130cbbba9123509e08b",
   "asyncpg==0.31.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ dependencies = [
   "aiohttp-swagger3>=0.10.0",
   "aioipfs~=0.7.1",
   "alembic==1.19.1",
-  "aleph-message @ git+https://github.com/aleph-im/aleph-message@e07d0fb5c01deab3526c0477aba64c3643275a83",
+  "aleph-message==1.4.0",
   "aleph-nuls2==0.1.1",
   "aleph-p2p-client @ git+https://github.com/aleph-im/p2p-service-client-python@1e992aaacdb0071a34ffb130cbbba9123509e08b",
   "asyncpg==0.31.0",
--- a/tests/message_processing/test_process_confidential.py
+++ b/tests/message_processing/test_process_confidential.py
@@ -10,6 +10,7 @@ from aleph_message.models import (
     Chain,
     ExecutableContent,
     InstanceContent,
+    ItemHash,
     ItemType,
     MessageType,
 )
@@ -18,9 +19,11 @@ from aleph_message.models.execution.volume import ImmutableVolume
 from more_itertools import one
 
 from aleph.db.accessors.files import insert_message_file_pin, upsert_file_tag
+from aleph.db.accessors.messages import get_message_status
 from aleph.db.accessors.vms import get_instance, get_vm_version
 from aleph.db.models import (
     AlephBalanceDb,
+    AlephCreditBalanceDb,
     EphemeralVolumeDb,
     ImmutableVolumeDb,
     MessageStatusDb,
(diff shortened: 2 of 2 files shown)


foxpatch-aleph (dismissed):
Minimal, correct change. The dependency bump is verified to contain only the TDX schema additions relative to 1.3.1, and I confirmed by executing the test fixture against the pinned schema that a mode:"tdx" instance parses, defaults policy to 1 with firmware None, and validates as credit payment — exactly what the test asserts and what vm.py persists. No TEE-mode branching exists in pyaleph, so the claim that no CCN code changes are needed checks out, and pricing correctly keys on trusted_execution presence. The test follows the established end-to-end pattern (same credit-balance pattern as other passing tests) and passes lint. Two non-blocking notes below.

**`pyproject.toml`** (line 33): Confirmed the git-sha pin style matches the existing aleph-p2p-client pin. As noted in the description, remember to re-pin to a tagged aleph-message release once #161 ships in one (or merge on the sha pin, consistent with prior V-PROGRAM bumps).

**`tests/message_processing/test_process_confidential.py`** (line 454): Optional nit: alongside the two TEE facts pinned here, asserting `instance.payment_type == "credit"` would also lock in the credit-only requirement for measured instances — though that rule is enforced upstream in aleph-message, so this is purely a nice-to-have.

foxpatch-aleph (approved):
Minimal, correct change: the aleph-message 1.4.0 pin is the only dependency change and brings exactly the described TDX schema delta (TeePlatform.tdx, TdxRegisters, platform-discriminated registers union, measured-TEE credit-only generalization). I verified the 'no CCN changes needed' claim by running the vm handler mapping with the test's TDX content against the real 1.4.0 package: it produces payment_type=credit, policy=1 (AMDSEVPolicy.NO_DBG) and firmware=None, exactly what the new test asserts, and pricing correctly keys on TEE presence via INSTANCE_CONFIDENTIAL. The test fixture content validates against the 1.4.0 schema, the credit-lot fixture matches established patterns in other passing tests, get_message_status is sync as used, and black/isort/ruff/mypy all pass on the changed files. I could not run the full pipeline test locally (no PostgreSQL/Docker in the sandbox), so the PROCESSED-status path was verified by direct execution of the validation and mapping layers rather than the DB round-trip. Two non-blocking nits: the new fixture duplicates the PendingMessageDb/MessageStatusDb boilerplate from the existing confidential fixture, and the pre-existing int4 policy column (which a real 64-bit SEV-SNP policy can overflow, already documented at vprogram.py:82-84) deserves a follow-up issue.

**`tests/message_processing/test_process_confidential.py`** (line 454): Verified: aleph-message 1.4.0 defaults `policy` to `AMDSEVPolicy.NO_DBG` (1) and `firmware` to None for tdx mode, so both this assertion and the `is None` firmware check below are correct — I ran `_map_content_to_db_model` with this exact content and got these values. Optional nit: importing and comparing against `AMDSEVPolicy.NO_DBG` instead of the literal `1` would self-document the 'schema default' claim in the comment, though the existing comment already explains it well.

**`tests/message_processing/test_process_confidential.py`** (line 375): The PendingMessageDb + MessageStatusDb setup (~35 lines) duplicates `fixture_confidential_vm_message` above. A small parameterized helper for the common skeleton would reduce drift risk if the schema of these fixtures changes again. Non-blocking — consistent with the file's existing style.

**`src/aleph/handlers/content/vm.py`** (line 125): Pre-existing and correctly called out in the PR description: `content.environment.trusted_execution.policy` is stored into the int4 `environment_trusted_execution_policy` column (src/aleph/db/models/vms.py:158), but a real SEV-SNP 64-bit guest policy with high bits set would overflow int4. The 1.4.0 schema now validates sev_snp policies as 64-bit, so a follow-up widening the column (or following the vprogram.py approach of leaving it NULL) would be worthwhile. Out of scope for this PR since tdx policy is pinned to the default.

PR #1267, opened 2026-09-14
Grandfather STORE/PROGRAM messages using on-chain confirmation height

odesenfans:
## Explain what problem this PR is resolving

A node syncing from scratch replays historical messages (STORE/PROGRAM) long after they were published. During first-time processing, `message.confirmations` is always empty — confirmations are only inserted *after* the balance checks — so `are_store_and_program_free()` always fell back to `observed_time` (i.e. "now") and current balance rules were applied retroactively to pre-cutoff messages.

Consequence, observed in production: a freshly synced node rejected the entire historical message set of senders whose balance is depleted today (~164k messages diverging from long-running nodes on one node), including a STORE message pinning a file still actively served by the network — the node could never fetch the file because the balance exception fires before `fetch_related_content`.

Related Clickup or Jira tickets : ALEPH-XXX

## Self proofreading checklist

- [x] Is my code clear enough and well documented
- [x] Are my files well typed
- [ ] New translations have been added or updated if new strings have been introduced in the frontend
- [ ] Database migrations file are included
- [x] Are there enough tests
- [ ] Documentation has been included (for new feature)

## Changes

- `toolkit/costs.py`: `are_store_and_program_free()` accepts an optional `confirmed_height` that takes precedence over `message.confirmations[0].height`, itself falling back to `observed_time` (unchanged — backdated unconfirmed messages still cannot claim the free legacy tier).
- `db/accessors/chains.py`: new `get_tx_height()` accessor.
- `handlers/message_handler.py`: resolve the confirmation height from the pending message's `tx_hash` (set on the on-chain replay path, origin `onchain`) and thread it to `pre_check_balance` / `check_balance`.
- Content handler signature update (`content_handler.py`, `store.py`, `vm.py`, `vprogram.py`): new optional `confirmed_height` keyword argument.

Only messages proven on-chain to predate the cost cutoff are grandfathered; nothing else changes.

## How to test

- `test_pre_check_balance_confirmed_height_grandfathering`: a post-cutoff-observed message from a depleted sender is rejected without chain evidence, accepted with a confirmation below the cutoff height.
- `test_pre_check_balance_confirmed_height_after_cutoff`: a confirmation at/after the cutoff still applies the balance check.
- `test_process_store_confirmed_before_cutoff_from_depleted_account`: end-to-end through `MessageHandler.process()` — reproduces the fresh-node sync incident (pre-cutoff on-chain STORE, sender has 20 ALPH accumulated costs and zero balance) and verifies the message is now processed.
[...]

PR #1268, opened 2026-09-14
Don't bill accumulated costs in the STORE pre-check for native storage

odesenfans:
## Explain what problem this PR is resolving

`pre_check_balance()` validated native-storage STORE messages with a **zero message cost**. Since `validate_balance_for_payment()` adds the sender's *entire accumulated cost* to the required balance, a brand-new message was rejected purely because of the sender's past storage — even though the new message contributed nothing to the requirement.

Observed in production: a node replaying an old STORE message from a sender with 20.196 ALPH accumulated costs and zero balance rejected it with `required_balance: 20.196...`, `account_balance: 0`. Because the exception fires **before** `fetch_related_content()`, the pinned file was never fetched at all.

Follows up on #1267 (which fixed grandfathering; this PR fixes the same validation path for post-cutoff messages).

- [x] Are my files well typed
- [x] Are there enough tests

## Changes

- `handlers/content/store.py` (`pre_check_balance`): when the message cost cannot be estimated before the content is fetched (native storage engine, IPFS disabled, or unknown IPFS size), defer entirely to `check_balance()` instead of validating with `message_cost = 0`. The full check runs after `fetch_related_content()` where the real file size is known.
- The existing small-file allowance in `check_balance()` is untouched and is now reachable for native-storage files. No new exemption is introduced.

## How to test

- `test_process_store_small_file_from_depleted_account`: a small native-storage STORE from a depleted sender (accumulated costs, zero balance) is processed — it previously failed the pre-check with `InsufficientBalanceException` before the file was even fetched.
- `test_process_store_without_tx_hash_rejected_from_depleted_account`: control — a large native-storage file from the same depleted sender is still rejected by `check_balance()`.
- Two tests from #1267 were moved to the IPFS path (large IPFS file), since the storage-engine pre-check they exercised no longer validates.

Run: `hatch run testing:test tests/message_processing/test_process_stores.py`

## Notes

- IPFS messages are unaffected: when the size is known the pre-check still validates early (including accumulated costs), as covered by `test_pre_check_balance_with_existing_costs`.


foxpatch-aleph (approved):
This PR correctly fixes a real production bug: the STORE pre-check validated native-storage messages with message_cost=0, which still required balance >= accumulated costs and rejected depleted senders before their content was even fetched. Deferring to check_balance() (which runs after fetch_related_content() and can apply the small-file allowance) is the right fix, and the IPFS known-size path keeps its early validation. The confirmed_height plumbing is secure: only the trusted chain-sync path sets tx_hash, the chain_txs lookup fails closed (charges) when the tx is missing, and height is non-nullable in the schema. Tests cover both grandfathered and post-cutoff heights plus a control case proving large files from depleted senders are still rejected. Two non-blocking notes: the deferred check means unfunded senders can now force a content fetch before rejection (an accepted, bounded trade-off), and the tx-height lookup runs twice per message.

**`src/aleph/handlers/content/store.py`** (line 324): Behavioral trade-off worth noting explicitly in the PR description: by deferring entirely to check_balance() here (and for unknown-size IPFS at line 347), a sender with zero balance can now force nodes to fetch and store file content (and pin on IPFS) before the message is rejected — the old zero-cost pre-check rejected depleted senders before the fetch. This seems like an accepted trade-off to make the small-file allowance reachable, and it's bounded by the fetch timeout/tries, but it slightly widens the DoS surface for unfunded senders. If this is a concern, a post-fetch size cap for unauthenticated senders could restore most of the protection.

**`src/aleph/handlers/message_handler.py`** (line 513): Minor: get_tx_height_from_pending_message() issues a chain_txs lookup both in verify_and_fetch_message() (pre-check) and here in process(), i.e. two extra queries per message. Consider resolving the height once in process() and passing it into verify_and_fetch_message(). Non-blocking.

**`src/aleph/handlers/content/vprogram.py`** (line 189): confirmed_height is accepted but unused here. It's fine for interface consistency (check_balance's signature is shared across handlers), but a one-line comment noting that v-programs are credit-only post-cutoff and thus never take the grandfathering path would help future readers.

PR #1269, opened 2026-09-14
Count only post-cutoff costs in the hold balance requirement

odesenfans:
## Explain what problem this PR is resolving

The balance cron job and message ingestion disagreed on what an address owes:

- The **reaper** (`balance_job.py`) only removes messages whose confirmation height is at or after `STORE_AND_PROGRAM_COST_CUTOFF_HEIGHT` — pre-cutoff (grandfathered) and unconfirmed resources are never removed.
- **Ingestion** (`validate_balance_for_payment`) required the sender's balance to cover their *entire* accumulated cost, grandfathered resources included.

Consequence: a sender with pre-cutoff storage who depleted their balance fails the hold balance check for every new message, while the reaper will never free those resources — a dead end. Freshly synced nodes, which replay history against today's balances, diverge from long-running nodes on the entire message set of such senders (~164k messages on one production node; see also #1267 and #1268).

- [x] Are my files well typed
- [x] Are there enough tests

## Changes

- `db/accessors/cost.py`: new `get_post_cutoff_total_cost_for_address()` — sums cost rows whose message has a chain confirmation at or after the cutoff height (unconfirmed messages excluded, same semantics as the reaper).
- `services/cost_validation.py`: the hold branch of `validate_balance_for_payment()` uses the cutoff-aware sum. Credit and superfluid validation are unchanged.

## How to test

- `tests/services/test_cost_validation.py::test_hold_balance_check_ignores_pre_cutoff_costs`: a sender with only pre-cutoff costs passes the hold validation with zero balance (previously rejected).
- `tests/services/test_cost_validation.py::test_hold_balance_check_counts_post_cutoff_costs`: post-cutoff costs still require balance.
- `tests/db/test_cost.py::test_get_post_cutoff_total_cost_for_address`: the accessor sums only post-cutoff confirmed costs among pre-cutoff, post-cutoff and unconfirmed rows.

Run: `hatch run testing:test tests/services/test_cost_validation.py tests/db/test_cost.py`

## Notes

- No migration: the query joins existing tables (`account_costs` → `message_confirmations` → `chain_txs`).
- Related PRs: #1267 (grandfathering via confirmation height), #1268 (pre-check accumulated-cost billing).


foxpatch-aleph (approved):
The core fix is correct and well-aligned with the reaper's semantics: get_post_cutoff_total_cost_for_address counts only costs confirmed at or after the cutoff height, excludes unconfirmed rows (matching the reaper's inner-join behavior), and uses exists() to avoid double-counting multi-confirmed messages. The confirmed_height plumbing is sound since chain tx rows are committed before pending messages are published, and the height comes from verified chain data so grandfathering cannot be forged. The store.py pre-check restructure correctly defers to check_balance() (which runs unconditionally with the real file size), and test coverage spans accessor, service, and end-to-end levels with good control cases. Only minor nits remain.

**`src/aleph/handlers/message_handler.py`** (line 513): Minor: get_tx_height_from_pending_message() is evaluated twice per on-chain message — once here for check_balance and once at line 446 for pre_check_balance (which process() calls via verify_and_fetch_message). It's a cheap indexed PK lookup, but you could resolve it once in process() and thread it through to avoid the redundant query.

**`src/aleph/db/accessors/cost.py`** (line 57): Nit: this function is only ever called with PaymentType.hold (from validate_balance_for_payment), yet it accepts any payment_type. Unlike get_total_cost_for_address — where get_costs_summary skips the filter when payment_type is None — passing None here would silently filter on payment_type IS NULL and always return 0. Consider tightening the signature (drop the parameter or assert hold) to avoid the footgun.

**`src/aleph/db/accessors/cost.py`** (line 64): Worth being conscious of: excluding unconfirmed costs mirrors the reaper, but it also means a sender's required hold balance stops growing between publish and on-chain aggregation — a sender with balance >= one message's cost can accumulate several unconfirmed hold resources (only reaped once confirmed). Deliberate per the docstring and tested, just flagging the tradeoff.

**`src/aleph/handlers/content/vprogram.py`** (line 185): check_balance accepts confirmed_height but never uses it (only interface conformance with the base class). Fine as-is; a one-line comment noting the parameter is intentionally unused here would help future readers wondering whether vprogram should grandfather on chain height.

PR #1270, opened 2026-09-15
chore(deps): aleph-message 1.5.0, the V-PROGRAM confidential GPU block

odesenfans:
## Why

aleph-message 1.5.0 (aleph-message#162) adds the `gpu` block to `VerifiableProgramContent`: `{vendor: nvidia, arch: hopper|blackwell, count, models?, mode: cc}`. aleph-vm 2.1 places and launches V-PROGRAMs that carry it, and the aleph-rs SDK builds them, but a CCN on 1.4.0 rejects such a message as schema-invalid, so no GPU V-PROGRAM can be published until the network runs this.

The 1.4.0 to 1.5.0 diff is that block plus a pydantic type annotation on `ItemHash`'s JSON schema hook; nothing else moves.

## What changes for the node

Only the pin. The block is not a passthrough GPU: `gpu_requirements` is empty on a V-PROGRAM (the inherited property reads `requirements.gpu`, which 1.5.0 refuses on V-PROGRAMs), so the cost service keeps pricing a GPU V-PROGRAM as a plain V-PROGRAM and the instance GPU tiers are untouched. Charging for confidential cards needs a pricing rule keyed on architecture rather than on a device model, which is a separate decision.

## Tests

`tests/messages/test_vprogram.py`:
- a V-PROGRAM with a `gpu` block is accepted at the ingestion entry point, `requires_gpu` is true and `gpu_requirements` stays empty
- a coerced `count`, an architecture without a confidential mode, a `models` entry that is not a PCI id, and an unknown key are all refused
- `requirements.gpu` on a V-PROGRAM is refused

https://claude.ai/code/session_01BZyMqLgxxVfeDwiNy34sfM


foxpatch-aleph (approved):
Minimal, well-scoped PR: the aleph-message 1.4.0→1.5.0 pin plus ingestion-level tests for the new confidential GPU block. I verified the upstream library diff is exactly what the description claims (gpu block on VerifiableProgramContent, requires_gpu override, refusal of requirements.gpu, ItemHash JSON-schema hook annotation), confirmed the pyaleph consumers are unaffected as stated (cost.py prices V-PROGRAMs via the isinstance at cost.py:306 regardless of gpu; gpu_requirements at cost.py:186 is instance-only; no code calls model_json_schema), ran the test file — all 20 pass on 1.5.0 and 2 of the new tests fail on 1.4.0, so they genuinely pin the bump — and the wider messages/schemas/V-PROGRAM suites pass with only environmental postgres errors. The new tests are well-documented and target the right entry point.

**`tests/messages/test_vprogram.py`** (line 190): This refusal is also a wire-format tightening on pyaleph's side: a V-PROGRAM with requirements.gpu parsed fine under 1.4.0 and now fails validation, so any such historical message would stop re-parsing (repair/forget/API re-validation). V-PROGRAM is new enough that this is almost certainly moot, but it's the one behavioral risk of the bump — worth a conscious confirmation that none exist on-chain.

**`tests/messages/test_vprogram.py`** (line 159): Nice touch asserting list(parsed.gpu_requirements) == [] here: it pins the cost-service-visible behavior (V-PROGRAM still priced as a plain V-PROGRAM) without reaching into the cost service. Non-blocking idea for a follow-up: a test through format_message_dict / the cost-estimation endpoint would also pin the API round-trip of the gpu block, but the ingestion entry point is the right scope for this PR.

